PR #55, opened 2019-05-29
feat(cli): add css files to HMR as well

nguyenlam1:
## Motivation

- make the web app refresh when making changes to css files as well. Short demo below!

cssHMR

### DIFF
--- a/packages/serve/lib/webpack.prod.js
+++ b/packages/serve/lib/webpack.prod.js
@@ -22,11 +22,17 @@ const cfg = ({
       },
     },
     module: {
-      rules: [{
-        enforce: 'pre',
-        test: /\.js$/,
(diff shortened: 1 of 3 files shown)


packages/serve/lib/webpack.prod.js
```diff
@@ -22,11 +22,17 @@ const cfg = ({
...
-        loader: sourceMapLoaderPath,
-      }],
+      rules: [
+        {
+          enforce: 'pre',
+          test: /\.js$/,
+          loader: sourceMapLoaderPath,
+        },
+        {
+          test: /\.css$/i,
```
nguyenlam1: @miralemd Not sure if I am missing something else here? 🤔

nguyenlam1:
Seems to be working in master.

PR #95, opened 2019-08-27
refactor: remove `run` for all `yarn`

stoffeastrom:
The `run` argument is not necessary so removing it.


miralemd:
I would prefer if we kept `run` as it makes it obvious that we are running one of our own scripts, otherwise it could become ambiguous as to whether a script or a yarn built in command is being executed.

PR #225, opened 2019-12-06
test: spice up `SelectionToolbarItem`

stoffeastrom:
## Motivation

Add `SelectionToolbarItem` tests

Did some refactoring also

### DIFF
--- a/apis/nucleus/src/components/SelectionToolbarItem.jsx
+++ b/apis/nucleus/src/components/SelectionToolbarItem.jsx
@@ -12,35 +12,30 @@ const useStyles = makeStyles(theme => ({
   },
 }));
 
-const Item = props => {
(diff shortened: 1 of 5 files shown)


apis/nucleus/src/components/SelectionToolbarItem.jsx
```diff
@@ -12,35 +12,32 @@ const useStyles = makeStyles(theme => ({
...
-      onChange = () => {
-        setDisabled(props.item.enabled && !props.item.enabled());
-      };
-
-      props.item.on('changed', onChange);
+    if (item && item.action && item.on) {
+      item.on('changed', onChanged);
     }
 
     return () => {
```
miralemd: might as well return from the `if` above

PR #244, opened 2019-12-16
feat: add `createKeyStore`

stoffeastrom:
## Motivation

Introduce `createKeyStore` which is used for model/layout caching.


stoffeastrom:
This add a new generic hook for rpc communication that handles the rpc state and longrunning features. New hooks can easily be added e.g

```
import useRpc from './useRpc';

export default function useLayout(model) {
  return useRpc(model, 'getLayout');
}

export function useAppLayout(model) {
  return useRpc(model, 'getAppLayout');
}
```

PR #296, opened 2020-02-04
fix(app-selections): maintain modal state

stoffeastrom:
## Motivation

Only maintain modal state for listbox popover

Remove som unused modules


stoffeastrom:
Kapture 2020-02-04 at 15 40 37

PR #349, opened 2020-03-12
feat: make dims and measures optional in qae

Caele:
## Motivation
Building a chart that doesn't use either dimensions or measures will still get these injected into the qae definition. For example it defaults to allowing any number of dimensions.

Discussion: Should we instead check for "max: 0" and exclude them then?

### DIFF
--- a/apis/supernova/src/qae.js
+++ b/apis/supernova/src/qae.js
@@ -7,7 +7,8 @@ function fallback(x, value) {
   return () => x;
 }
 
-function defFn(def = {}) {
+function defFn(input) {
+  const def = input || {};
   return {
     min: typeof def.min === 'function' ? def.min : fallback(def.min, 0),
     max: typeof def.max === 'function' ? def.max : fallback(def.max, 1000),
@@ -16,6 +17,7 @@ function defFn(def = {}) {
     moved: def.moved || def.move || noop,
     removed: def.removed || def.remove || noop,
     replaced: def.replaced || def.replace || noop,
+    isDefined: () => !!input,
   };
 }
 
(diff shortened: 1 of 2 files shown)


miralemd:
It's not ideal to have optional methods from an API perspective, this would mean that anywhere we want to check for dimensions we have to do it conditionally:

```js
const min = qae.dimensions ? qae.dimensons.min() : 0;
```

Any particular reason why this should be optional for the internal API?

Caele:
My whole reason for this is that the client displays dimension options if the dimensions are included.
For example the KPI in the client does:
```
data: {
    measures: {
      min: 1,
      max: 2,
      description(properties, index) { ... },
    },
  },
```
But doing that in a nebula version of the KPI gives you dimension options in the property panel. I could set this instead:
```
data: {
    measures: { ... },
    dimensions: { min:0, max:0 }
  },
```
But that causes the client to still display the dimension header and allow for alternative dimensions.

It is likely an oversight in the client as well, so if this is not an appropriate fix, it can be done there instead.

miralemd:
Then I would suggest to expand the API to support that case, something like:

```js
qae.dimensions.isDefined()
```

Caele:
Soo something like:
```
function defFn(input) {
const def = input || {};
  return {
    min: typeof def.min === 'function' ? def.min : fallback(def.min, 0),
    max: typeof def.max === 'function' ? def.max : fallback(def.max, 1000),
    added: def.added || def.add || noop, // TODO - deprecate add in favour of added
    description: def.description || noop,
    moved: def.moved || def.move || noop,
    removed: def.removed || def.remove || noop,
    replaced: def.replaced || def.replace || noop,
   isDefined() {return !!def;}
  };
}
```
Then check that in the client?

apis/supernova/src/qae.js
```diff
@@ -16,6 +17,7 @@ function defFn(def = {}) {
     moved: def.moved || def.move || noop,
     removed: def.removed || def.remove || noop,
     replaced: def.replaced || def.replace || noop,
+    isDefined: () => !!input,
```
miralemd: test?

PR #357, opened 2020-03-13
refactor(nucleus): rename `configured` to `createConfiguration`

stoffeastrom:
## Motivation

API polishing 💅 

BREAKING CHANGE: Use `nucleus.createConfiguration` instead of `nucleus.configured`

### DIFF
--- a/test/mashup/snaps/configured.js
+++ b/test/mashup/snaps/configured.js
@@ -23,7 +23,7 @@ const bar = function(env) {
 };
 
 // eslint-disable-next-line
-const configured = nucleus.configured({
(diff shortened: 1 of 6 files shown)


test/mashup/snaps/configured.js
```diff
@@ -23,7 +23,7 @@ const bar = function(env) {
 };
 
 // eslint-disable-next-line
-const configured = nucleus.configured({
+const createConfiguration = nucleus.createConfiguration({
```
miralemd:
```suggestion
const configured = nucleus.createConfiguration({
```

PR #376, opened 2020-03-24
feat(cli-build): add sourcemaps option to build

Caele:
## Motivation

For integrated uses of the built packages we sometimes need to disable sourcemaps, this adds an option to do that via the build command.

### DIFF
--- a/commands/build/lib/init-config.js
+++ b/commands/build/lib/init-config.js
@@ -17,8 +17,8 @@ const options = {
     alias: 'w',
     default: false,
   },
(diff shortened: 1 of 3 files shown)


commands/build/lib/init-config.js
```diff
@@ -17,6 +17,12 @@ const options = {
     alias: 'w',
     default: false,
   },
+  sourcemaps: {
```
miralemd: Would prefer the option as `sourcemap` to keep it the same as rollup

PR #418, opened 2020-04-29
Update render-charts.md

nilzona:
Typo


miralemd:
`supernovae` is plural for `supernova`, so not a typo :P

PR #426, opened 2020-05-15
chore: generate all locales

stoffeastrom:
## Motivation

This ensures using the new locale json files.

Use new script `generate-all.sh` under `@nebula.js/locale`. Right now it's handled manually but should be automated


stoffeastrom:
Revamped the script with node to make @miralemd happy :)

PR #453, opened 2020-06-18
feat: better incomplete view

### LINKED ISSUE #422: Better "Incomplete Visualization" message
## 🚀 Feature

Based on the hypercube requirements coming from the supernova we're rendering the messages should be showing what is missing until requirements are met.

### Motivation

Much better user experience

### Additional context

If for some reason there are technical limitations blocking this feature it would be nice to be able to control the rendering of the "Incomplete Visualisation" from the api.

stoffeastrom:
## Motivation

Add better visual feedback for incomplete visualizations.

image

For visualizations with multiple hypercubes we use one list with combined dimensions and measures:

image

Closes #422

### DIFF
--- a/apis/locale/locales/en-US.json
+++ b/apis/locale/locales/en-US.json
@@ -83,8 +83,36 @@
     "value": "Select possible",
     "comment": "Action text to select possible selections"
   },
-  "Supernova.Incomplete": {
+  "Visualization.LayoutError": {
+    "value": "Error",
+    "comment": "Status text shown when a visualization has layout errors"
+  },
+  "Visualization.Incomplete": {
     "value": "Incomplete visualization",
     "comment": "Status text shown when a visualization is incomplete"
+  },
+  "Visualization.Incomplete.Dimensions": {
+    "value": "{0} of {1} dimensions",
+    "comment": "Status text shown when a visualization is incomplete"
+  },
+  "Visualization.Incomplete.Measures": {
+    "value": "{0} of {1} measures",
+    "comment": "Status text shown when a visualization is incomplete"
+  },
+  "Visualization.Invalid.Dimension": {
+    "value": "Invalid dimension",
+    "comment": "Status text shown when a dimension is not valid"
+  },
+  "Visualization.Invalid.Dimension.CalculationCondition": {
+    "value": "Invalid dimension",
+    "comment": "Status text shown when a dimension has a unfulfilled calculation condition"
+  },
+  "Visualization.Invalid.Measure": {
+    "value": "Invalid measure",
+    "comment": "Status text shown when a measure is not valid"
+  },
+  "Visualization.Invalid.Measure.CalculationCondition": {
+    "value": "Invalid dimension",
+    "comment": "Status text shown when a measure has a unfulfilled calculation condition"
   }
 }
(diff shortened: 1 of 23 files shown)


stoffeastrom:
@Linguisht Can we rename the `key` like it's done [here](https://github.com/qlik-oss/nebula.js/pull/453/files#diff-5d186e6b3c4e47122543e7cb92d7d57fR86) or will it break the automation? Also not sure how you generate the hashes.

Linguisht:
> @Linguisht Can we rename the `key` like it's done [here](https://github.com/qlik-oss/nebula.js/pull/453/files#diff-5d186e6b3c4e47122543e7cb92d7d57fR86) or will it break the automation? Also not sure how you generate the hashes.
Yes, this is fine. Hashes are computed from English key+value+comment (md5 base64 encoded) and then stored in the locale files as `version`. Since this resource doesn't exist under this new key in the translation database, it will be picked up automatically and sent for translation, but the previous translation values will be re-used (unless the human translators intervene/correct something), so the only change in the translation PR that will created will be the `version` property.

apis/locale/locales/en-US.json
```diff
@@ -83,8 +83,36 @@
...
+    "value": "{0} of {1} measures",
+    "comment": "Status text shown when a visualization is incomplete"
+  },
+  "Visualization.Invalid.Dimension": {
+    "value": "Invalid dimension",
+    "comment": "Status text shown when a dimension is not valid"
+  },
+  "Visualization.Invalid.Dimension.CalculationCondition": {
+    "value": "Invalid dimension",
+    "comment": "Status text shown when a dimension has a unfulfilled calculation condition"
```
miralemd, line 108:
```suggestion
  "Visualization.UnfulfilledCalculationCondition": {
    "value": "The calculation condition is not fulfilled",
    "comment": "Message displayed when a calculation condition is not fulfilled"
```

apis/locale/locales/en-US.json
```diff
@@ -83,8 +83,36 @@
...
+    "value": "Error",
+    "comment": "Status text shown when a visualization has layout errors"
+  },
+  "Visualization.Incomplete": {
     "value": "Incomplete visualization",
     "comment": "Status text shown when a visualization is incomplete"
+  },
+  "Visualization.Incomplete.Dimensions": {
+    "value": "{0} of {1} dimensions",
+    "comment": "Status text shown when a visualization is incomplete"
```
miralemd, line 96:
```suggestion
    "comment": "Text showing the number of current {0} dimensions over the total {1} number of dimensions"
```

apis/locale/locales/en-US.json
```diff
@@ -83,8 +83,36 @@
...
+  },
+  "Visualization.Incomplete.Measures": {
+    "value": "{0} of {1} measures",
+    "comment": "Status text shown when a visualization is incomplete"
```
miralemd, line 100:
```suggestion
    "comment": "Text showing the number of current {0} measures over the total {1} number of measures"
```

apis/locale/locales/en-US.json
```diff
@@ -83,8 +83,36 @@
...
+  },
+  "Visualization.UnfulfilledCalculationCondition": {
+    "value": "The calculation condition is not fulfilled",
+    "comment": "Message displayed when a calculation condition is not fulfilled"
+  },
+  "Visualization.Invalid.Measure": {
+    "value": "Invalid measure",
+    "comment": "Status text shown when a measure is not valid"
+  },
+  "Visualization.Invalid.Measure.CalculationCondition": {
```
miralemd, line 114: Remove

PR #508, opened 2020-09-14
fix(stardust): support class-pyramid in theme

T-Wizard:
## Motivation

implements the // TODO - generate class-pyramid
matching what is done in the sense client

### DIFF
--- a/apis/theme/src/__tests__/theme-scale-generator.spec.js
+++ b/apis/theme/src/__tests__/theme-scale-generator.spec.js
@@ -1,0 +1,59 @@
+import scaleGenerator from '../theme-scale-generator';
+
+describe('Theme scale generator', () => {
+  const input = ['#ffffff', '#000000'];
+  const base8 = ['#ffffff', '#d4d4d4', '#aaaaaa', '#7f7f7f', '#545454', '#2a2a2a', '#000000'];
+
+  beforeEach(() => {});
+
+  it('Should generate a pyramid', () => {
+    const scales = [{ type: 'class', scale: input }];
+    scaleGenerator(scales);
+    expect(scales).to.have.length(1);
+    const { scale, type } = scales[0];
+
+    expect(type).to.equals('class-pyramid');
+    expect(scale.length).to.equal(8);
+    expect(scale[scale.length - 1].length).to.equal(7);
+  });
+
+  it('Should generate a correct base of colors', () => {
+    const scales = [{ type: 'class', scale: input }];
+    scaleGenerator(scales);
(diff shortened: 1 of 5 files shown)


miralemd (changes requested):
We're already using d3-color and should do so for this feature as well to avoid using a custom implementation.

apis/theme/src/colors/color-scale.js
```diff
@@ -0,0 +1,84 @@
+import Color from './color';
+
+const NaNColor = new Color(0xd2d2d2);
```
miralemd, line 3: Shouldn't the NaN color come from the applied theme?

T-Wizard: This match how it is done in the sense client.
I don't think it is important as it should not be used if the scale is correctly defined.
(If it should be from the theme what color in the theme should it use?)

T-Wizard:
now update to use d3-color

apis/theme/src/colors/color-scale.js
miralemd: Would be nice if we didn't just copy the mistakes from previous code ;) - NaN/Null color is specified in the [theme](https://github.com/qlik-oss/nebula.js/blob/master/apis/theme/src/themes/base.json#L9)

T-Wizard: done.
Is it relevant to have a warning when the nan color is used to indicate that there is a problem with the theme?

miralemd: No need for a warning specific for the NaN color right now, but if we see a lot of issues of poorly formatted themes then having a theme validator would be nice to have in the future.

T-Wizard:
ping @miralemd: I have fixed all you comments (did I miss something?)

miralemd:
I'm looking at how this affects the theme API input and output - If I understand this correctly, [these lines](https://github.com/qlik-oss/nebula.js/pull/508/files#diff-64d54da90306e9b4fae10be2a7cd06ddR88-R94) mean that there will never be a `'class'` output in `dataScales`, which then also means you can never provide a `'class'` scale that isn't changed internally? So the result in the API output should be to remove `'class'` [here](https://github.com/qlik-oss/nebula.js/blob/master/apis/theme/src/palette-resolver.js#L4) and replace it with `'class-pyramid'`?

T-Wizard:
Yes I think you are right and 'class' should be replaced with 'class-pyramid' and the type of the color property need to changed from {string[]} colors to {string[]|Array<Array<string>>}

Caele (approved):

PR #518, opened 2020-10-05
fix(theme): add additional palette fallback

Caele:
# Motivation

When a sense theme incorrectly includes empty arrays for palettes they end up as empty in the finished theme, even with inheritance enabled.

### DIFF
--- a/apis/theme/src/set-theme.js
+++ b/apis/theme/src/set-theme.js
@@ -15,7 +15,7 @@ export default function setTheme(t, resolve) {
   if (!rawThemeJSON.palettes.ui || !rawThemeJSON.palettes.ui.length) {
     rawThemeJSON.palettes.ui = root.palettes.ui;
   }
-  if (!rawThemeJSON.scales) {
+  if (!rawThemeJSON.scales || !rawThemeJSON.scales.length) {
     rawThemeJSON.scales = root.scales;
   }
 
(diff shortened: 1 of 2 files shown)


apis/theme/src/set-theme.js
```diff
@@ -9,10 +9,10 @@ export default function setTheme(t, resolve) {
...
   const rawThemeJSON = extend(true, {}, root, { scales: null, palettes: { data: null, ui: null } }, t);
-  if (!rawThemeJSON.palettes.data) {
+  if (!rawThemeJSON.palettes.data || !rawThemeJSON.palettes.data.length) {
     rawThemeJSON.palettes.data = root.palettes.data;
   }
-  if (!rawThemeJSON.palettes.ui) {
+  if (!rawThemeJSON.palettes.ui || !rawThemeJSON.palettes.ui.length) {
     rawThemeJSON.palettes.ui = root.palettes.ui;
   }
   if (!rawThemeJSON.scales) {
```
henrikgildeus, line 18: Maybe add the same for scales? Just in case

henrikgildeus (approved):

apis/theme/src/set-theme.js
Caele: yes indeed

T-Wizard (approved):

PR #535, opened 2020-11-11
feat: add basic support for object conversion

quanho:
This PR is the first step to support object conversion.


T-Wizard:
any support for a way object can indicate that they do not support object conversion?

T-Wizard:
what about documentation?

quanho:
> what about documentation?

Currently I put the method "convertTo" under DO_NOT_USE part. When we finish it we will have document for it.

quanho:
> any support for a way object can indicate that they do not support object conversion?

Yes, we can support it like in sense-client.

Caele (approved):

T-Wizard (approved):

PR #536, opened 2020-11-13
chore: add dev mashup example

Caele:
## Motivation

To simplify nebula development we add an mashup example targeting development of nebula itself.

### DIFF
--- a/examples/dev-mashup/README.md
+++ b/examples/dev-mashup/README.md
@@ -1,0 +1,27 @@
+# Nebula Development Mashup example
+
+Example mashup for nebula development
+
+## Install
+
+Update dependencies
+
+```
+yarn
+```
+
+To use a local nebula version, run this in the stardust folder to link you local version.
+
+```
+yarn link @nebula.js/stardust
+```
+
+## Run
+
+Run parcel server
+
+```
+yarn start
+```
+
+Open you browser to http://localhost:1234
--- a/examples/dev-mashup/README.md
+++ b/examples/dev-mashup/README.md
@@ -4,16 +4,17 @@ Example mashup for nebula development
 
 ## Install
 
-Update dependencies
+To use a local nebula version, run this in the stardust folder to allow linking your local version.
+
+```
+yarn link
+```
+
+Update dependencies and link nebula
 
 ```
 yarn
-```
-
-To use a local nebula version, run this in the stardust folder to link you local version.
-
-```
-yarn link @nebula.js/stardust
+yarn link "@nebula.js/stardust"
 ```
 
 ## Run
--- a/examples/dev-mashup/connect.js
+++ b/examples/dev-mashup/connect.js
@@ -1,0 +1,53 @@
+import enigma from 'enigma.js';
+
+export default function connect() {
+  const loadSchema = () => fetch('https://unpkg.com/enigma.js/schemas/3.2.json').then((response) => response.json());
+
+  const createConnection = () =>
+    loadSchema().then((schema) =>
+      enigma
+        .create({
+          schema,
+          url: `ws://${window.location.hostname || 'localhost'}:9076/app/${Date.now()}`,
+        })
+        .open()
+        .then((qix) => qix.createSessionApp())
+    );
+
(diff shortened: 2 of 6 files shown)


goldbergjeffrey:
Will this mashup example get posted on qlik.dev? Is it part of the documentation that flows from this repo to there?

examples/dev-mashup/connect.js
```diff
@@ -0,0 +1,53 @@
...
+
+export default function connect() {
+  const loadSchema = () => fetch('https://unpkg.com/enigma.js/schemas/3.2.json').then((response) => response.json());
+
+  const createConnection = () =>
+    loadSchema().then((schema) =>
+      enigma
+        .create({
+          schema,
+          url: `ws://${window.location.hostname || 'localhost'}:9076/app/${Date.now()}`,
```
goldbergjeffrey, line 11: I'd like for us to be using QSE SaaS in these examples. I know in the past we've targeted Windows and Qlik Core, but we want to subtly infer the use of Qlik Sense Enterprise SaaS.

Caele:
> Will this mashup example get posted on qlik.dev? Is it part of the documentation that flows from this repo to there?

No, currently this is just something for us to use internally to simplify thing.

examples/dev-mashup/connect.js
Caele: sure, but this example is for specific development in Nebula. So will worry later about that.

examples/dev-mashup/README.md
```diff
@@ -0,0 +1,27 @@
...
+Update dependencies
+
+```
+yarn
+```
+
+To use a local nebula version, run this in the stardust folder to link you local version.
+
+```
+yarn link @nebula.js/stardust
```
Caele, line 16: unsure about this line, so needs feedback.

quanho: yarn link "@nebula.js/stardust" should be run in the folder where we want to have stardust. So in this case we should run it in local-mashup folder.

quanho (approved):

quanho:
This is very helpful for debugging and testing object conversion!

examples/dev-mashup/connect.js
```diff
@@ -0,0 +1,53 @@
+import enigma from 'enigma.js';
+
+export default function connect() {
+  const loadSchema = () => fetch('https://unpkg.com/enigma.js/schemas/3.2.json').then((response) => response.json());
```
T-Wizard, line 4: consider using a newer schema version 3.2 is almost 4 years old (12.612.0.json looks to be the latest)

examples/dev-mashup/index.js
```diff
@@ -0,0 +1,36 @@
+import { embed } from '@nebula.js/stardust';
+import bar from '@nebula.js/sn-bar-chart';
+import line from '@nebula.js/sn-line-chart';
+import connect from './connect';
+
+function init() {
+  connect().then(async (app) => {
```
T-Wizard, line 7: as you are not using async await syntax (maybe you should) you should not need any async here

examples/dev-mashup/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "name": "local-mashup",
+  "version": "1.0.0",
+  "description": "Simple local mashup",
+  "main": "index.html",
+  "scripts": {
+    "start": "parcel index.html --open",
+    "build": "parcel build index.html"
+  },
+  "peerDependencies": {
+    "@nebula.js/stardust": "1.0.2-alpha.0"
```
T-Wizard, line 11: why is this a peer dependencies?
I think it should be a normal (or maybe dev) dependency

Caele: yea true, probably doesn't matter as it will be added anyway. I'll change version as well to latest.

quanho: It should be peerDependencies since we use sn-bar-chart and sn-line-chart in this example and sn-bar-chart and sn-line-chart also use stardust.
I tested moving stardust to dependencies but I got problems!

T-Wizard (approved):

PR #537, opened 2020-11-17
feat: add exportProperties for hypercube conversion

quanho:
This PR is to 
1. add conversion as a separated module,
2. add exportProperties for hypercube conversion.

The following example shows how it can be used in la-vie charts:
```
  const exportProperties = ({ propertyTree }) => {
    return conversion.hypercube.exportProperties({ propertyTree, hypercubePath });
  };
  const importProperties = ({ exportFormat }) => {
    return conversion.hypercube.importProperties({ exportFormat, hypercubePath });
  };
```

TODO:
1. add unit tests
2. prepare for document

Added unit tests and document.


apis/conversion/src/index.js
```diff
@@ -0,0 +1,33 @@
...
+  return SN;
+};
+
+const convertTo = async ({ halo, model, cellRef, newType }) => {
+  const propertyTree = await model.getFullPropertyTree();
+  const sourceQae = cellRef.current.getQae();
+  if (!sourceQae.exportProperties) {
+    throw new Error(
+      `Converting from ${propertyTree.qProperty.visualization} is not supported since there is no exportProperties method.`
+    );
```
Caele, line 20: Need to figure out how to handle this.
Nebula should fallback to default conversion if they are missing from the qae, so the question is how we go about disabling them. Setting them to a falsy value is one way, but it isn't very explicit. Currently I'm not sure if there are any other extension level settings.
The "support" thing we use in the client is tied to the ext() file, so that is no good. Maybe there is some kind of check for export/snapshot somewhere in nebula?

apis/conversion/src/utils.js
```diff
@@ -0,0 +1,121 @@
...
+  } else {
+    delete dataContainer[dataName];
+  }
+};
+
+const isEmpty = (object) => {
+  return Object.keys(object).length === 0 && object.constructor === Object;
+};
+
+const checkLibraryObjects = (exportFormat, dimensionList, measureList) => {
```
Caele, line 69: Is this used anywhere? It requires the dim/mea lists which are client things.

quanho: Removed since it is not used yet, will add it back in another PR.

apis/conversion/src/index.js
quanho: This will be handled in a separate PR, since this PR is too big already.

apis/conversion/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+    })
+    .supernova();
+  return SN;
+};
+
+const getPath = (qae) => {
+  return qae && qae.data && qae.data.targets && qae.data.targets[0] && qae.data.targets[0].propertyPath;
+};
+
+const getDefaultExportPropertiesFnc = (path) => {
```
veinfors, line 18: I would prefer to call it `getDefaultExportPropertiesFn` instead, and that goes for all of the functions here ending with `Fnc`, rename to `Fn` instead...

apis/conversion/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "@nebula.js/conversion",
+  "version": "1.0.0",
```
veinfors, line 3: I think it's better to set to to version 0.1.0 for now, to indicate it's not complete yet. At least until the basic functionality is in place (like listbox conversion and a nice way to disable object conversion through the Api etc...)

apis/conversion/src/export-format.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Used for exporting and importing properties between backend models. An object that exports to
+ * ExportFormat should put dimensions and measures inside one data group. If an object has two hypercubes
+ * each of the cubes should export dimensions and measures in two separate data groups.
+ * An object that imports from this structure is responsible for putting the existing properties where they should be
+ * in the new model.
+ *
+ * @param nDataGroups The number of hypercubes.
```
veinfors, line 8: Should we add the @ingore tag here maybe? I'm thinking it's an "internal" function...

apis/conversion/src/index.js
quanho: Fixed!

apis/conversion/package.json
quanho: Fixed!

apis/conversion/src/export-format.js
quanho: This is used as a parameter for importPropeties and as a return for exportProperties. I will update the doc for ExportFormat

Caele: it should be ignored until we release it, but also need to make sure the specs work properly. So maybe add in a label of sorts for now DO_NOT_USE so it shows up in the specs, just in case we do an early release.

veinfors (approved):

Caele (approved):

PR #538, opened 2020-11-26
feat: add importProperties for hypercube conversion

quanho:
This PR is to add importProperties for hypercube conversion.

TODO:
1. add more unit tests
2. prepare for document

### DIFF
--- a/apis/conversion/src/__tests__/index.spec.js
+++ b/apis/conversion/src/__tests__/index.spec.js
@@ -23,6 +23,9 @@ describe('objectConversion', () => {
         data: {
           targets: [{ propertyPath: '/qHyperCubeDef' }],
         },
+        properties: {
+          initial: {},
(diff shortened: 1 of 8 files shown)


apis/conversion/src/__tests__/index.spec.js
```diff
@@ -54,24 +57,32 @@ describe('objectConversion', () => {
     });
 
     it('should use hypercube.exportProperties and hypercube.importProperties when there is no source exportProperties and no target importProperties', async () => {
```
Caele, line 59: This statement:
It should check the hyperCubePath as well. (Maybe it does already).

quanho: Yes, I did check the hypercubePath.

Caele (approved):

PR #542, opened 2020-11-30
refactor: split big files into smaller files

quanho:
This PR is to refactor export and import properties to make it easier to add unit tests.


Caele (approved):
If it is just refactoring, then go ahead.

quanho:
> If it is just refactoring, then go ahead.

Yes, it is just refactoring.

PR #547, opened 2020-12-03
feat: add checking master item for object conversion

quanho:
This PR is to:
1. Move useLibraryList from commands/serve/web to nucleus so it can be reused
2. use useLibraryList in Cell.jsx to get library items
3. expose library items through cellRef to be used in conversion
4. use library items to update master dimensions/measures during conversion

### DIFF
--- a/apis/nucleus/src/components/Cell.jsx
+++ b/apis/nucleus/src/components/Cell.jsx
@@ -15,6 +15,7 @@ import useRect from '../hooks/useRect';
 import useLayout, { useAppLayout } from '../hooks/useLayout';
 import InstanceContext from '../contexts/InstanceContext';
 import useObjectSelections from '../hooks/useObjectSelections';
+import useLibraryList from '../hooks/useLibraryList';
 
 /**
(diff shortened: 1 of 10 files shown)


T-Wizard:
why is there a title that needs to be updated?
I don't think the should be needed or included in the default hypercube conversion

apis/nucleus/src/components/Cell.jsx
```diff
@@ -284,6 +285,8 @@ const Cell = forwardRef(({ halo, model, initialSnOptions, initialError, onMount
   const [selections] = useObjectSelections(app, model);
   const [hovering, setHover] = useState(false);
   const hoveringDebouncer = useRef({ enter: null, leave: null });
+  const [dimensionList] = useLibraryList(app, 'dimension');
```
veinfors, line 288: Including these hooks in here - does that mean the dimension and measure list will get fetched every time a chart is initialized?

quanho:
The title is only needed in filterpane.

Issue #548, opened 2020-12-04
Basic embeddable Listbox


Caele:
## 🚀 Feature

Allow mounting a basic listbox in a nebula mashup, similar to the AppSelections

### Motivation

Gives direct access to field selections without the need for extensions

PR #549, opened 2020-12-04
Tsm/field api

Caele:
## Motivation
#548 
Enables users to mount a basic listbox into their Nebula mashup.

ToDos
- [x] Update specs
- [x] Check intake for autosizer

Set as experimental, so delaying:
- [ ] Verify API design
- [ ] Update docs

### DIFF
--- a/apis/nucleus/src/index.js
+++ b/apis/nucleus/src/index.js
@@ -4,6 +4,7 @@ import appThemeFn from './app-theme';
 
 import bootNebulaApp from './components/NebulaApp';
 import AppSelectionsPortal from './components/selections/AppSelections';
+import ListBoxPortal from './components/listbox/ListBoxInline';
 
 import create from './object/create-session-object';
 import get from './object/get-object';
@@ -311,6 +312,25 @@ function nuked(configuration = {}) {
       },
       __DO_NOT_USE__: {
         types,
+        field: (fieldName) => {
+          return {
+            mount(element) {
+              this._instance = ListBoxPortal({
+                element,
+                app,
+                fieldName,
+              });
+              root.add(this._instance);
+            },
+            unmount() {
+              // se
+              if (this._instance) {
+                root.remove(this._instance);
+                this._instance = null;
+              }
+            },
+          };
+        },
       },
     };
 
--- a/apis/nucleus/src/index.js
+++ b/apis/nucleus/src/index.js
@@ -320,7 +320,12 @@ function nuked(configuration = {}) {
          * field.mount(element, { title: "Hello Field"});
          */
         field: (fieldName) => {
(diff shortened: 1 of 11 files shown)


veinfors (approved):

Caele:
@T-Wizard  @quanho Any comments?

apis/nucleus/src/index.js
```diff
@@ -266,9 +267,9 @@ function nuked(configuration = {}) {
...
       /**
        * Gets the app selections of this instance.
-       * @returns {Promise<AppSelections>}
+       * @returns {AppSelections}
        * @example
-       * const selections = await n.selections();
+       * const selections = n.selections();
        * selections.mount(element);
        */
       selections: async () => {
```
T-Wizard, line 275: this still return a promise.

apis/nucleus/src/index.js
```diff
@@ -309,6 +310,62 @@ function nuked(configuration = {}) {
         }
         return selectionsApi;
       },
+      /**
+       * Gets the instance of the specified field
+       * @returns {FieldSelections}
+       * @experimental
+       * @example
+       * const field = await n.selections();
```
T-Wizard, line 318:
```suggestion
       * const field = n.field(fieldName);
```

T-Wizard:
Also consider adding range select to the listbox (maybe out of scope for this PR)

apis/nucleus/src/index.js
veinfors: Yes it does, but looking at what it actually returns, I see no need for the function to be async actually...
It might break something though, to change the function into being synchronous.

Caele: yea, that is what confused me. Does an async function always return a promise wrapping whatever you return?

Caele: I can see now why it was a Promise, due to the:
// const appSelections = await root.getAppSelections(); // Don't expose this for now
 Should I make the Field one a promise as well, just be to consistent?

T-Wizard: > Does an async function always return a promise wrapping whatever you return?
Yes.

T-Wizard: as for the fields api, maybe to be future safe. (I am unsure what I would recommend)

Caele: > > Does an async function always return a promise wrapping whatever you return?
> 
> Yes.

yea, makes sense, just haven't considered it before.
I'll do the field one async for now as well, then we can discuss that before the official release.

Caele:
> Also consider adding range select to the listbox (maybe out of scope for this PR)

yea, I've though about it but trying to limit this a bit.

T-Wizard (approved):

PR #556, opened 2021-01-11
fix: support master dims

Caele:
## Motivation
Continueation of #548 

- Adds support for master dimensions
- Adds test case for inline listbox

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -146,7 +146,7 @@ export function ListBoxInline({ app, fieldIdentifier, stateName = '$', options =
         <Grid item>
           {showTitle && (
             <Typography variant="h6" noWrap>
-              {!isMaster ? layout.title || fieldName : layout.title || layout.qListObject.qDimensionInfo.qFallbackTitle}
+              {layout.title || layout.qListObject.qDimensionInfo.qFallbackTitle}
             </Typography>
           )}
         </Grid>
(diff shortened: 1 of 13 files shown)


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -137,7 +146,7 @@ export function ListBoxInline({ app, fieldName, stateName = '$', options = {} })
         <Grid item>
           {showTitle && (
             <Typography variant="h6" noWrap>
-              {layout.title || fieldName}
+              {!isMaster ? layout.title || fieldName : layout.qListObject.qDimensionInfo.qFallbackTitle}
```
T-Wizard, line 149: have you considered if setting title should override the title from the master dimension?
(if yes you could use `layout.title || layout.qListObject.qDimensionInfo.qFallbackTitle`

T-Wizard (approved):

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -137,7 +146,7 @@ export function ListBoxInline({ app, fieldName, stateName = '$', options = {} })
         <Grid item>
           {showTitle && (
             <Typography variant="h6" noWrap>
-              {layout.title || fieldName}
+              {!isMaster ? layout.title || fieldName : layout.title || layout.qListObject.qDimensionInfo.qFallbackTitle}
```
T-Wizard, line 149: when not a master item `layout.qListObject.qDimensionInfo.qFallbackTitle` should be the same as `fieldName`
```suggestion
              {layout.title || layout.qListObject.qDimensionInfo.qFallbackTitle}
```

Caele:
@quanho I changed the params for the convertTo to fit the jsdoc. I think using straight function params as opposed to objects is nicer in the exposed API

PR #558, opened 2021-01-25
chore: hide function in spec

Caele:


T-Wizard (approved):

Caele:
Api gov fails due to changes/additions to scriptappy, so will disable that for now.

PR #563, opened 2021-02-02
fix: update dimensions and measures on object conversion

### LINKED ISSUE #564: Failed on conversion from a line chart or bar chart to a combo chart
Cannot convert from a line chart or a bar chart to a combo chart.
The problem is that the combo chart requires series.type for each measure. However, conversion from a line chart or a bar chart to a combo chart fails to add this property for a measure.

quanho:
This PR is to fix #564 by adding necessary properties for dimensions/measures on conversion.

### DIFF
--- a/apis/conversion/src/hypercube/__tests__/import-properties.spec.js
+++ b/apis/conversion/src/hypercube/__tests__/import-properties.spec.js
@@ -26,6 +26,8 @@ describe('importProperties', () => {
       copyPropertyIfExist: sandbox.stub(),
       copyPropertyOrSetDefault: sandbox.stub(),
       importCommonProperties: sandbox.stub(),
+      updateDimensionsOnAdded: sandbox.stub(),
+      updateMeasuresOnAdded: sandbox.stub(),
     };
     [{ default: importProperties }] = doMock({ helpers });
     exportFormat = {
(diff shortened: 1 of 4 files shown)


apis/conversion/src/hypercube/__tests__/import-properties.spec.js
```diff
@@ -99,5 +105,7 @@ describe('importProperties', () => {
     expect(helpers.copyPropertyIfExist.callCount).to.equal(2);
     expect(helpers.copyPropertyOrSetDefault.callCount).to.equal(6);
     expect(helpers.importCommonProperties.callCount).to.equal(1);
+    expect(helpers.updateDimensionsOnAdded.callCount).to.equal(1);
+    expect(helpers.updateMeasuresOnAdded.callCount).to.equal(1);
```
Caele, line 109: The tests here only check if they've been called, which doesn't really tell us anything.
Should write a separate test to run teh functionality of the updateXYZOnAdded functions.

quanho: Will update it!

quanho: Updated!

Issue #573, opened 2021-02-05
Set/consume preferred change device type


Caele:
## 🚀 Feature

To optimize interaction with a chart it is often important to know what the target device is. This can be automatically detected in some cases, but in others is is ambiguous. The proposal is as follows
- Being able to consume the setting from within a supernova
- Being able to set it from a mashup

Set as: auto, touch, desktop
Consume as: touch or desktop

### Motivation

To improve integration with sense client and enable touch specific features in charts and mashups.

PR #587, opened 2021-03-03
feat: add device type

T-Wizard:
implement #573 (Set/consume preferred change device type)


apis/nucleus/src/device-type.js
```diff
@@ -0,0 +1,25 @@
+// from https://patrickhlauke.github.io/touch/touchscreen-detection/ (MIT License)
```
Caele, line 1: Should make an intake for this, but with MIT license it won't be a problem.

Caele:
Looks good!

Caele (approved):

PR #591, opened 2021-03-30
chore: add error for incorrect load function

Caele:
## Motivation

Adds error to type load function.


T-Wizard (approved):
Consider adding a unit test for this change

PR #592, opened 2021-03-31
chore: import editorial review

Caele:
## Motivation

Fixes spelling and grammar

TODO

- [x] Fix spec compatibility

### DIFF
--- a/apis/nucleus/src/hooks/useAppSelections.js
+++ b/apis/nucleus/src/hooks/useAppSelections.js
@@ -59,7 +59,6 @@ function createAppSelections({ app, currentSelectionsLayout, navState }) {
 
   /**
    * @class
-   * @hideconstructor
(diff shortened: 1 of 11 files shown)


apis/nucleus/src/hooks/useAppSelections.js
```diff
@@ -59,7 +59,6 @@ function createAppSelections({ app, currentSelectionsLayout, navState }) {
 
   /**
    * @class
-   * @hideconstructor
```
T-Wizard, line 62: why are you removing @hideconstructor? (here and in other places)

Caele: The spec is invalid with it, not sure if the schema is correct or not, but removing it won't affect the docs.

T-Wizard (approved):

PR #599, opened 2021-04-26
feat: add support for chart plugins

veinfors:
### Render with plugins
```js
embed(app).render({
  element,
  type: 'my-chart',
  plugins: [plugin]
});
```

### usePlugins hook
```js
import { usePlugins } from '@nebula.js/stardust';
// ...
const plugins = usePlugins();
plugins.forEach((plugin) => {
  // Invoke plugin
  plugin.fn();
});
```

### Plugin format
```js
const plugin = {
  info: {
    name: "example-plugin",
    type: "meta-type",
  },
  fn: () => {
    // Plugin implementation goes here
  }
};
```


apis/nucleus/src/plugins/plugins.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * @interface Plugin
+ * @property {object} info Object that can hold various meta info about the plugin
+ * @property {string} info.name The name of the plugin
+ * @property {function} fn The implementation of the plugin. Input and return value of this function can be whatever is necessery for This can be called from the chart implementation with necessary input.
```
Caele, line 5: - Add description (just a line of text at the top)
- Fix fn description to be readable.

Caele:
Looks good I think. Can we add a .md file to the docs here as well with a light example? That way we have something to bring over to the official docs as we release this.

veinfors:
> Looks good I think. Can we add a .md file to the docs here as well with a light example? That way we have something to bring over to the official docs as we release this.

Sure, good idea, I will come up with some example and att that.

apis/nucleus/src/plugins/plugins.js
veinfors: Done and done!

veinfors:
> > Looks good I think. Can we add a .md file to the docs here as well with a light example? That way we have something to bring over to the official docs as we release this.
> 
> Sure, good idea, I will come up with some example and add that.

@Caele I've added a `plugins.md` file... Take a look if it's OK or if something is missing...

Caele (approved):

PR #606, opened 2021-05-26
feat(serve): support url flags

T-Wizard:

### DIFF
--- a/commands/serve/web/components/App.jsx
+++ b/commands/serve/web/components/App.jsx
@@ -86,6 +86,7 @@ export default function App({ app, info }) {
         language: currentLanguage,
       },
       load: (type) => Promise.resolve(window[type.name]),
+      flags: info.flags,
       themes: info.themes
         ? info.themes.map((t) => ({
             key: t,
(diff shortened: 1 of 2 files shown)


quanho (approved):

T-Wizard:
@Caele, @veinfors: Any opinion on how we want to work with feature flags in nebula.js and nebula.js-charts
This is usefull if you want to temporary enable a flag in a chart you are working in

veinfors:
> @Caele, @veinfors: Any opinion on how we want to work with feature flags in nebula.js and nebula.js-charts
> This is usefull if you want to temporary enable a flag in a chart you are working in

Yeah, I think this is great. It's very useful.. One thing that's a little annoying though is that the syntax will differ from what we have in Qlik Sense (`/feature/blablaflag` vs `&flags=blablaflag`). It might cause some confusion, but it's not the whole world. We can always re-visit and adjust that later if we feel the need for it.

veinfors (approved):

veinfors:
We COULD of course extend the functionality even further by supporting to pass in a reference to a json file or a json object to `nebula serve` to provide flags. But let's start with this - enabling a flag through the url 👍

PR #622, opened 2021-06-03
feat(cli-build): include source maps of dependencies

T-Wizard:
## Motivation

Improved debugging experience


veinfors (approved):

Caele (approved):
Does it still allow you to turn them off?

T-Wizard:
> Does it still allow you to turn them off?

yes you can still turn of source maps, but it is all or nothing there are no options to select what dependencies you want to include source maps from

PR #625, opened 2021-06-17
feat(build): additions to build and serve cli

veinfors:
Adds the following to `build` and `serve` commands:
* Possibility to pass in replacement strings through a function `nebula.config.js`. (Had to make it a function due to problems with yargs and yargs-parser...)
* Include `jsx plugin` to allow building supernovas with react code.
* Suppress circular dep. warning messages to reduce noise (a lot of charts use d3, which has circular deps. - this is no problem for rollup though)

### DIFF
--- a/commands/build/lib/build.js
+++ b/commands/build/lib/build.js
@@ -20,6 +20,13 @@ const { terser } = require('rollup-plugin-terser');
 
 const initConfig = require('./init-config');
 
+const resolveReplacementStrings = (replacementStrings) => {
+  if (typeof replacementStrings !== 'function' || typeof replacementStrings() !== 'object') {
+    throw new Error('replacementStrings should be a function returning an object with key value pairs of strings!');
+  }
+  return replacementStrings();
+};
+
 const config = ({
   mode = 'production',
   format = 'umd',
@@ -32,7 +39,7 @@ const config = ({
   let pkg = require(path.resolve(CWD, 'package.json')); // eslint-disable-line
   const corePkg = core ? require(path.resolve(core, 'package.json')) : null; // eslint-disable-line
   const { name, version, license, author } = pkg;
-  const { sourcemap } = argv;
+  const { sourcemap, replacementStrings = () => ({}) } = argv;
 
   if (corePkg) {
     pkg = corePkg;
@@ -67,18 +74,13 @@ const config = ({
 
   return {
     input: {
-      onwarn(warning, warn) {
-        // Suppress circular dep. warnings to avoid noise (d3 circular dep. warnings...)
-        if (warning.code === 'CIRCULAR_DEPENDENCY') return;
-        warn(warning);
-      },
       input: path.resolve(CWD, 'src/index'),
       external,
       plugins: [
         replace({
           'process.env.NODE_ENV': JSON.stringify(mode === 'development' ? 'development' : 'production'),
           preventAssignment: true,
-          ...(typeof argv.replacementStrings === 'function' ? argv.replacementStrings() : {}),
+          ...resolveReplacementStrings(replacementStrings),
         }),
         nodeResolve({
           extensions: ['.mjs', '.js', '.jsx', '.json', '.node'],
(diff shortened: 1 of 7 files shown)


commands/build/lib/build.js
```diff
@@ -66,14 +67,22 @@ const config = ({
 
   return {
     input: {
+      onwarn(warning, warn) {
+        // Suppress circular dep. warnings to avoid noise (d3 circular dep. warnings...)
+        if (warning.code === 'CIRCULAR_DEPENDENCY') return;
+        warn(warning);
```
Caele, line 73: There's a risk with supressing things. We might want to for specific reasons, but along comes someone else who needs to see it.
Could we do it as a --silent or --verbose mode? Best would be if we use the same as rollup and pass that flag along.

Caele (commented):
Looks good, only the warning thing I must warn about.

commands/serve/lib/webpack.build.js
```diff
@@ -86,6 +86,7 @@ const cfg = ({ srcDir, distDir, dev = false, serveConfig = {} }) => {
         __NEBULA_DEV__: true,
         'process.env.NEBULA_VERSION': JSON.stringify(version),
         'process.env.NEBULA_VERSION_HASH': JSON.stringify(versionHash),
+        ...(typeof serveConfig.replacementStrings === 'function' ? serveConfig.replacementStrings() : {}),
```
T-Wizard, line 89: why is this needed? this looks to only compile the nebula serve page not the supernova used

commands/serve/lib/serve.js
```diff
@@ -138,6 +138,7 @@ module.exports = async (argv) => {
         watch: true,
         config: serveConfig.config,
         cwd: context,
+        replacementStrings: serveConfig.replacementStrings,
```
T-Wizard, line 141: when investagating why this was added I think I found that the config path is not passed correcly to build.
(I have not found why yet)

veinfors: I know, and I couldn't figure it out, that's why I solved it like this...

commands/build/lib/build.js
veinfors: Yeah, you are right. I read somewhere that rollup supports circular dependencies and therefore though it was OK to suppress these warning. But after some more reading and thinking, this only involves certain types of circular dependencies (of course) - hence, it's better to not suppress these warnings, as you said. - Done!

commands/serve/lib/webpack.build.js
veinfors: You are right, this is absolutely not needed. I misunderstood the code somehow... Removed!

commands/serve/lib/serve.js
T-Wizard: it can be fixed by changing
```
defaultBuildConfig = initConfig(yargs([])).argv;
```
to
```
const yargsArgs = argv.config ? ['--config', argv.config] : [];
defaultBuildConfig = initConfig(yargs(yargsArgs)).argv;
```
in [commands/build/lib/build.js](https://github.com/qlik-oss/nebula.js/blob/57a80910072d79d8e501f9b0b685495b24a87178/commands/build/lib/build.js#L219)

an alternative would be to not use yargs to load the config file (that would also make it possible for replacementStrings to be a normal object in the config file)

Caele (approved):
Approving this before my vaction, leaving it for Tobias to check.

commands/serve/lib/serve.js
veinfors: Thanks for the feedback! I will look into loading the config file in another way, not using yargs...

veinfors: I did what you suggested to solve the problem with the config path.
I also changed the format of `replacementStrings` into an object. I had to set the rule `dot-notations` to `false` through `yargs.parserConfiguration` to get it working nicely though.

T-Wizard (approved):

PR #627, opened 2021-06-30
feat: update build and build legacy for nebula sense

quanho:

### DIFF
--- a/commands/sense/lib/build.js
+++ b/commands/sense/lib/build.js
@@ -70,17 +70,9 @@ async function build(argv) {
   };
 
   const copySource = () => {
-    try {
-      fs.copySync(path.resolve(main), path.resolve(targetDir, main));
-    } catch (e) {
-      console.error(e);
-    }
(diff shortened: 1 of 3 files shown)


commands/sense/lib/build.js
```diff
@@ -69,6 +69,21 @@ async function build(argv) {
     }
   };
 
+  const copySource = () => {
+    try {
```
veinfors, line 73: I suggest you skip the try/catch blocks here... making a `console.error(e)` will print the error message but the execution will continue. But if the copy commands are not going through, there are no point in continuing the way I see it.

veinfors:
Looks good to me. Only left a comment about a minor thing

commands/sense/lib/build.js
quanho: Updated!

T-Wizard (approved):

veinfors (approved):

PR #632, opened 2021-07-01
feat: switch --partial to default and add --legacy option

quanho:


Caele (approved):
Looking good, we'll need to make sure the changelog mentions this correctly when merging.

PR #655, opened 2021-08-16
Tsm/keyhandler

Caele:


apis/nucleus/src/components/glue.jsx
```diff
@@ -5,11 +5,13 @@ import Cell from './Cell';
 export default function glue({ halo, element, model, initialSnOptions, initialSnPlugins, onMount, initialError }) {
   const { root } = halo;
   const cellRef = React.createRef();
+  const currentId = String(Date.now()); // TODO: different ID generator?
```
T-Wizard, line 8: yes use a different ID generator, Date.now() will not create unique values if called to fast

apis/nucleus/src/components/Supernova.jsx
```diff
@@ -16,8 +16,9 @@ const VizElement = {
 const Supernova = ({ sn, snOptions: options, snPlugins: plugins, layout, appLayout, halo }) => {
   const { component } = sn;
 
-  const { theme: themeName, language, constraints } = useContext(InstanceContext);
+  const { theme: themeName, language, constraints, keyboardNavigation } = useContext(InstanceContext);
   const [renderDebouncer] = useState(() => new RenderDebouncer());
+
```
haxxmaxx, line 21: remove

apis/stardust/api-spec/spec.json
```diff
@@ -1500,6 +1516,26 @@
         }
       }
     },
+    "Keyboard": {
```
haxxmaxx, line 1519: focus should be here also right?

apis/supernova/src/hooks.js
```diff
@@ -979,3 +985,111 @@ export function onTakeSnapshot(cb) {
...
+    const focusFunction = function () {
+      const acc = this.__hooks.accessibility;
+      if (acc.enabled && !acc.active) {
+        blurCallback && blurCallback(false);
+        focus(this);
+      }
+    }.bind(currentComponent);
+
+    currentComponent.__hooks.accessibility.focusFunction = focusFunction;
+  }
```
haxxmaxx, line 1051: I guess you need to check if focusFunction is set as well?

Caele: not really, they are both created inside that IF, so if one is missing then both are and vice-versa

haxxmaxx: ok good :)

apis/nucleus/src/components/Cell.jsx
```diff
@@ -270,212 +271,240 @@ const loadType = async ({ dispatch, types, visualization, version, model, app, s
...
-        type: 'ERROR',
-        error: {
-          title: `Could not find a version of '${layout.visualization}' that supports current object version. Did you forget to register ${layout.visualization}?`,
-        },
-      });
-      return undefined;
-    }
-    load(layout.visualization, withVersion);
+    const handleKeyDown = (e) => {
+      if ([13, 32].includes(e.keyCode)) {
```
T-Wizard, line 314: consider using named constants to make this more readable.

Caele: sure, should use `.key` actually

T-Wizard (approved):

haxxmaxx (approved):

PR #662, opened 2021-08-30
docs(nebula-create): add example docs for nebula create

LiKang6688:
## Motivation

 example docs for nebula create

### DIFF
--- a/commands/create/README.md
+++ b/commands/create/README.md
@@ -58,7 +58,7 @@ Options:
 nebula create hello-sunshine
 ```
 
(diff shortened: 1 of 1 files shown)


commands/create/README.md
```diff
@@ -58,8 +58,48 @@ Options:
 nebula create hello-sunshine
 ```
 
+Running nebula create without --picasso promotes a selection of the available
```
Caele, line 61:
```suggestion
Running nebula create without --picasso prompts a selection of the available
```

PR #664, opened 2021-08-31
fix: do not skip all updates if the rendering take long time

T-Wizard:
## Motivation

If the rendering of an object took long time it was possible to skip updates and end up with a last render call that was not up to date with the latest layout.

Only seen in objects that modify its on properties on render

Simple example supernova with this problem:
```
function randomValue() {
  return Math.floor(Math.random() * 100);
}

export default function supernova() {
  return {
    component() {
      const layout = useStaleLayout();
      const element = useElement();
      const model = useModel();
      const [value] = useState(() => randomValue());
      const str = `The value is: ${value}`;
      const text = layout.text || 'Hello!';

      usePromise(async () => {
        if (str === text) {
          return;
        }
        const props = await model.getProperties();
        props.text = str;
        await model.setProperties(props);
        await model.getLayout();
      }, [layout]);

      if (str === text) {
        element.innerHTML = `<div>${text}</div>`; // eslint-disable-line
      } else {
        element.innerHTML = `<div>Updating</div>`; // eslint-disable-line
      }
    },
  };
}
```

### DIFF
--- a/apis/nucleus/src/components/Supernova.jsx
+++ b/apis/nucleus/src/components/Supernova.jsx
@@ -1,6 +1,7 @@
-import React, { useState, useEffect, useCallback, useContext, useRef } from 'react';
+import React, { useState, useEffect, useCallback, useContext } from 'react';
 import InstanceContext from '../contexts/InstanceContext';
 import useRect from '../hooks/useRect';
+import RenderDebouncer from '../utils/render-debouncer';
 
 /**
  * @interface
@@ -16,7 +17,7 @@ const Supernova = ({ sn, snOptions: options, snPlugins: plugins, layout, appLayo
   const { component } = sn;
 
   const { theme: themeName, language, constraints } = useContext(InstanceContext);
-  const renderDebouncer = useRef(null);
+  const [renderDebouncer] = useState(() => new RenderDebouncer());
   const [isMounted, setIsMounted] = useState(false);
   const [renderCnt, setRenderCnt] = useState(0);
   const [containerRef, containerRect, containerNode] = useRect();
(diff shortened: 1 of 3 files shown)


quanho (approved):

T-Wizard:
> Looks good I think.
> But before we merge this in, have you tested it with some of our existing charts to make sure we don't break any scenarios?

I have now done some test with combo-chart and boxplot. Quan will probably also do some tests

Caele (approved):

T-Wizard:
This keep the limit to only have one rendering onging for an object at the same time.
But when running inside sense-client it do not have that limit and do run multiple render in parallell.
Removing the limit and allow multiple render in parallell here it an alternate solution to this problem.

veinfors:
> This keep the limit to only have one rendering onging for an object at the same time.
> But when running inside sense-client it do not have that limit and do run multiple render in parallell.
> Removing the limit and allow multiple render in parallell here it an alternate solution to this problem.

I can't think of a scenario where multiple renderings are necessary/better than handling it like this with a debouncer, but there might be... What do you think?

T-Wizard:
> > This keep the limit to only have one rendering onging for an object at the same time.
> > But when running inside sense-client it do not have that limit and do run multiple render in parallell.
> > Removing the limit and allow multiple render in parallell here it an alternate solution to this problem.
> 
> I can't think of a scenario where multiple renderings are necessary/better than handling it like this with a debouncer, but there might be... What do you think?

I think I prefer only one rendering at a time.
The main argument I see for multiple renderings is to do the same as in sense-client. But there we may want to change how it work inside sense-client instead.

PR #671, opened 2021-09-06
feat: type guesser

Caele:
## Motivation

While looking on how to expand theming capabilities I ran into type matching issues. Basically the sense-client uses two type names, ex: "barchart" and "barChart". First one is type and other one is theme key. This is fine as long as you only need to access theme attributes from within the chart, but as soon as you need them outside it becomes an issue.

Say that you aim to render charts in an existing app, and want to manipulate the theme:
```js
const nebbie = stardust.embed(app, {
      types: [ { name: "barchart" , load:()=>{..}} ],
      themes: [...]
    });

nebbie.render({ id: "BHF75"}); // of type "barchart"
```
This setup would render the chart correctly, but the theme would not find it properly as the type there is "barChart".
Other examples we have involves loading types as "bar-chart" for example, and that would also cause issues.

TODO

- [x] Test
- [x] More tests
- [x] Handle chart titles

Edit: Will limit this to reading chart background color from the sense theme. When you get into using it for chart titles it starts messing with the selections toolbar and overall Nebula look and feel, so that should be handled using a proper solution tied to Material.

### DIFF
--- a/apis/conversion/src/__tests__/type-alias.spec.js
+++ b/apis/conversion/src/__tests__/type-alias.spec.js
@@ -1,0 +1,32 @@
+import getMatch from '../type-alias';
+
+describe('type matcher', () => {
+  it('should match client known charts', () => {
+    expect(getMatch('bar chart', ['barchart'])).to.equal('barchart');
+    expect(getMatch('combochart', ['combochart'])).to.equal('combochart');
+    expect(getMatch('lineChart', ['linechart'])).to.equal('linechart');
+    expect(getMatch('mekko-chart', ['mekkochart'])).to.equal('mekkochart');
+    expect(getMatch('sn-pie-chart', ['piechart'])).to.equal('piechart');
+  });
+  it('should match themes known charts', () => {
+    expect(getMatch('bar chart', ['barChart'])).to.equal('barChart');
+    expect(getMatch('combochart', ['comboChart'])).to.equal('comboChart');
+    expect(getMatch('lineChart', ['lineChart'])).to.equal('lineChart');
+    expect(getMatch('mekko-chart', ['mekkoChart'])).to.equal('mekkoChart');
+    expect(getMatch('sn-pie-chart', ['pieChart'])).to.equal('pieChart');
+  });
+  it('should match dash known charts', () => {
+    expect(getMatch('bar chart', ['bar-chart'])).to.equal('bar-chart');
+    expect(getMatch('combochart', ['combo-chart'])).to.equal('combo-chart');
+    expect(getMatch('lineChart', ['line-chart'])).to.equal('line-chart');
+    expect(getMatch('mekko-chart', ['mekko-chart'])).to.equal('mekko-chart');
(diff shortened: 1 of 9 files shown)


T-Wizard:
Why do is this way instead of letting the supernova object specify its own backround color? (by theme key or direct color)

Caele:
> Why do is this way instead of letting the supernova object specify its own backround color? (by theme key or direct color)

I guess due to legacy reasons. Client does it in `object`, along with styles for title/subtitle/footer. Done that way to allow us to skip implementing it in all charts. That won't stop a chart from implementing it themselves as well of course.
In Nebula it is a bit trickier as not only the title etc needs coloring, but also the integrated selections toolbar. I don't really have a good solution for it yet, so did this as a first try.

The more I look at the MUI theming in Nebula the more I think it needs a bigger overhaul before it can be exposed, so what I'm looking for now is more of a middle ground solution. Maybe just setting background color doesn't really help.

Caele:
Gonna remove the background color from this for now, so it can be reviewed without that philosophical burden.

T-Wizard:
without the use of it for background color why do we want this fuzzy type matching?

Caele:
> without the use of it for background color why do we want this fuzzy type matching?

@T-Wizard So that matching charts created in the client is easier, same goes for future theme things.
Do you see any potential negatives with this?

T-Wizard:
> > without the use of it for background color why do we want this fuzzy type matching?
> 
> @T-Wizard So that matching charts created in the client is easier, same goes for future theme things.
> Do you see any potential negatives with this?

Mostly that I don't see the value in the matching. I think it is more confusing then useful.
other negatives is that it can be hard to remove if we change out minds later.

Another thing is that is asymetric if A mathces B that do not mean that B matches A but I don't know how intentional this is or if it only is a limitation of the current code.
(casued by asymetric handling of spaces, case and splitting names without separators)

Caele:
> > > without the use of it for background color why do we want this fuzzy type matching?
> > 
> > 
> > @T-Wizard So that matching charts created in the client is easier, same goes for future theme things.
> > Do you see any potential negatives with this?
> 
> Mostly that I don't see the value in the matching. I think it is more confusing then useful.
> other negatives is that it can be hard to remove if we change out minds later.
> 
> Another thing is that is asymetric if A mathces B that do not mean that B matches A but I don't know how intentional this is or if it only is a limitation of the current code.
> (casued by asymetric handling of spaces, case and splitting names without separators)

I hear you, and mostly agree. I really isn't that useful if not needed for "fixing" the themes matching.
The asymetri comes from not being able to split strings like `barchart`, so it is mostly intentional (and partly limited by current code).

I might be looking for a problem that isn't there. In my mind the issue is the clients naming convention, which we can't change unless we start using some kind of matching. The reason I see for changing it would be a more granular and specific way to name charts, ie `sn-bar-chart` compared to just `barchart`. It would make chart names "clearer", as they could match the npm name of the extension. But what we actually gain is likely minimal, with a major effort/risk.
So yea, i'll put this on hold for now. Might want to come back to something like this later on, especially if we start work on multi-version support.

PR #673, opened 2021-09-10
feat: expose classnames

Caele:
## Motivation
Adds static classnames to charts:
- title
- subtitle
- footer
- action buttons

To allow for some basic re-styling of embedded charts through css:
Example:
```css
div.njs-cell {
    background: maroon;
}

p.njs-cell-footer, h6.njs-cell-title, p.njs-cell-sub-title {
    color: white;
}

button.njs-cell-action {
    color: white;
}
```

This would be a workaround until a proper full UI. theme can be exposed


T-Wizard (approved):

apis/nucleus/src/components/ActionsToolbarItem.jsx
```diff
@@ -19,6 +19,7 @@ const Item = React.forwardRef(({ item, addAnchor = false }, ref) => {
       onClick={item.action}
       disabled={disabled}
       style={style}
+      className="njs-cell-action"
```
veinfors, line 22: Should this be exposed in the spec as well?

Caele: yes it should

veinfors (approved):

PR #677, opened 2021-09-17
feat: useRenderState hook

T-Wizard:
## Motivation

Add useRenderState hook

Used to update properties and get a new layout without triggering onInitialRender.

### DIFF
--- a/apis/supernova/src/hooks.js
+++ b/apis/supernova/src/hooks.js
@@ -994,14 +994,14 @@ export function onTakeSnapshot(cb) {
  * Used to update properties and get a new layout without triggering onInitialRender.
  * @entry
  * @experimental
- * @returns {{ setPending, restore }} The render state.
+ * @returns {{ pending, restore }} The render state.
  * @example
  * import { useRenderState } from '@nebula.js/stardust';
  *
  * const renderState = useRenderState();
  * useState(() => {
  *   if(needProperteisUpdate(...)) {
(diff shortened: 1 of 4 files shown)


apis/supernova/src/hooks.js
```diff
@@ -979,3 +980,37 @@ export function onTakeSnapshot(cb) {
...
+ * @example
+ * import { useRenderState } from '@nebula.js/stardust';
+ *
+ * const renderState = useRenderState();
+ * useState(() => {
+ *   if(needProperteisUpdate(...)) {
+ *      useRenderState.setPending();
+ *      updateProperties(...);
+ *   } else {
+ *      useRenderState.restore();
```
veinfors, line 1000: It might be a little bit unclear what is actually meant by `restore`... I'm thinking if it would be easier with the following api:
```js
useRenderState.setPending(true);
...
useRenderState.setPending(false);
```

Caele: yea I kinda agree. But nothing feels right here.

T-Wizard: should I change it?

if we go with this the renderState object could be removed and the hook return the setPending method
const setPending = useSetPending();
...
setPending(true):
...
setPending(false);
but I unsure if that is more usable.

veinfors (approved):

Caele (approved):

PR #681, opened 2021-09-24
chore: add chart conversion example

LiKang6688:
## Motivation
add chart conversion example


examples/chart-conversion/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "@nebula.js/sn-bar-chart": "latest",
+    "@nebula.js/sn-combo-chart": "^1.4.1",
+    "@nebula.js/sn-line-chart": "latest",
+    "@nebula.js/sn-pie-chart": "^1.9.1",
+    "@nebula.js/stardust": "latest",
+    "enigma.js": "2.7.2"
+  },
+  "devDependencies": {
+    "@babel/core": "7.12.3",
+    "parcel-bundler": "^1.12.4"
```
Caele, line 20: parcel bundler is deprecated, we should avoid using it.
There are issues with the new parcel, but hopefully they can resolve later, so I think we should stick with it.

LiKang6688: Sure. gonna use the new parcel.

LiKang6688: Done

PR #683, opened 2021-09-30
feat(build): add typescript support in build

niekvanstaveren:
## Motivation

Adding the support for typescript in nebula build

### DIFF
--- a/commands/build/lib/build.js
+++ b/commands/build/lib/build.js
@@ -10,7 +10,7 @@ const postcss = require('rollup-plugin-postcss');
 const replace = require('@rollup/plugin-replace');
 const sourcemaps = require('rollup-plugin-sourcemaps');
 const jsxPlugin = require('@babel/plugin-transform-react-jsx');
-const typescript = require('@rollup/plugin-typescript');
+const typescriptPlugin = require('@rollup/plugin-typescript');
 const json = require('@rollup/plugin-json');
 const { nodeResolve } = require('@rollup/plugin-node-resolve');
 const commonjs = require('@rollup/plugin-commonjs');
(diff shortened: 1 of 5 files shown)


commands/build/lib/build.js
```diff
@@ -54,6 +55,7 @@ const config = ({
 
   const auth = typeof author === 'object' ? `${author.name} <${author.email}>` : author || '';
   const moduleName = name.split('/').reverse()[0];
+  const extensions = ['.mjs', '.js', '.jsx', '.json', '.node', '.tsx', '.ts'];
```
Caele, line 58: Only add .ts and .tsx when typescript is enabled?

commands/build/lib/init-config.js
```diff
@@ -33,6 +33,11 @@ const options = {
     type: 'string',
     default: undefined,
   },
+  typescriptEnabled: {
```
Caele, line 36: just name it `typescript`

commands/build/lib/build.js
niekvanstaveren: Yes that is a good idea

T-Wizard (approved):

veinfors (approved):

commands/build/lib/build.js
```diff
@@ -83,13 +88,14 @@ const config = ({
...
         nodeResolve({
-          extensions: ['.mjs', '.js', '.jsx', '.json', '.node'],
+          extensions,
         }),
         commonjs(),
         json(),
         babel({
           babelrc: false,
           inputSourceMap: false, // without this you get wrong source maps, but I don't know why
+          extensions,
```
Caele, line 98: this is a bit odd, but as long as it works.

Caele (approved):

LiKang6688 (approved):
Look forward to using it in the real case, could start with sn-pivot-table

PR #694, opened 2021-10-26
feat: handle tabbing from cell header

haxxmaxx:
## Motivation
part of https://github.com/qlik-oss/sn-table/issues/356

When `keyboardNavigation` is true, the selection toolbar should refocus the chart when you tab out of it (tab on last element or shift tab on first element). Also adding a method to the `keyboard` hook so that you can set focus on the selection toolbar from the chart


haxxmaxx:
Closing this since i opend this: https://github.com/qlik-oss/nebula.js/pull/701

PR #700, opened 2021-11-02
chore: update api compliance

joakim-ronning:
Hello!
We are migrating to a new image for doing API compliance.  
Check that the paths and names are correct to your repo.
If you have any questions don't hesitate to reach out.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,6 +173,7 @@ workflows:
               only:
                 - /v.*/
       - api-governance:
+          context: api-governance
           filters:
             branches:
               # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -176,6 +176,9 @@ workflows:
       - api-governance:
           context: api-compliance
           filters:
+            branches:
+              # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
+              ignore: /pull\/[0-9]+/
             tags:
               only:
                 - /v.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -130,10 +130,6 @@ jobs:
       - attach_workspace:
           at: ~/project
       - run:
-          name: Log into Github Docker Registry
-          command: |
-              echo $QLIK_DOCKER_DEV_PASSWORD | docker login -u $QLIK_DOCKER_DEV_USERNAME --password-stdin $QLIK_DOCKER_DEV_REGISTRY
-      - run:
           name: Create version.txt
           command: |
               set -x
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -133,12 +133,36 @@ jobs:
...
+          command: >
+            VER=$(cat ./version.txt)
+
+            docker run --volumes-from specs
+            -e SPEC_PATHS="abdc47c2-921a-45d3-ba5c-80d448574c5b@/specs/spec.json"
+            -e COMMIT_SHA="$CIRCLE_SHA1"
+            -e RELEASE_TAG="$VER"
+            -e CREDENTIALS_S3_SECRETKEY="$APICULTURIST_S3"
+            -e CREDENTIALS_GITHUB="$APICULTURIST_GITHUB"
+            -e CREDENTIALS_COLONY="$APICULTURIST_TOKEN"
```
T-Wizard, line 164: The APICULTURIST_* secrets needs to be added to qlik-oss organisation or nebula.js in circelci

joakim-ronning: We added an api-compliance context to OSS and I added it to the api governance job. Let me know if this fixes the issue.

T-Wizard: (I will not know if it actually works until after the PR is merged, as it is not run on PRs from forks)
but the context is named `api-compliance` i circelci but the this config use the name `api-governance` (I expect that to be a problem)

joakim-ronning: 100% my fault! Thank you for pointing it out. I'm going to fix it. If you want, you can give me access to the repository and I can recreate the branch there.

.circleci/config.yml
```diff
@@ -130,15 +130,40 @@ jobs:
       - attach_workspace:
           at: ~/project
       - run:
-          name: Login for Docker
-          command: echo $DOCKER_TOKEN | docker login -u "$DOCKER_USER" --password-stdin
+          name: Log into Github Docker Registry
+          command: |
+              echo $QLIK_DOCKER_DEV_PASSWORD | docker login -u $QLIK_DOCKER_DEV_USERNAME --password-stdin $QLIK_DOCKER_DEV_REGISTRY
```
T-Wizard, line 135: change back to using DOCKER_TOKEN and DOCKER_USER
or add QLIK_DOCKER_DEV_PASSWORD and QLIK_DOCKER_DEV_USERNAME to circle ci
(looks like QLIK_DOCKER_DEV_REGISTRY already exists)

.circleci/config.yml
```diff
@@ -149,10 +174,8 @@ workflows:
               only:
                 - /v.*/
       - api-governance:
+          context: api-compliance
           filters:
-            branches:
-              # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
-              ignore: /pull\/[0-9]+/
```
T-Wizard, line 155: Restore this condition
it is not secure to expose the secrets needed to PRs from forks and without them it will always fail

joakim-ronning: done

.circleci/config.yml
joakim-ronning: Because of the public image, we have removed this step.

joakim-ronning:
Test is failing because engine is expired. However, we won't know if the api compliance in CI is passing unless we can create the PR directly on the repo.

PR #701, opened 2021-11-02
feat: focus in and out of selection toolbar

haxxmaxx:
## Motivation

part of qlik-oss/sn-table#356

When keyboardNavigation is true, the selection toolbar should refocus the chart when you tab out of it (tab on last element or shift tab on first element). Adding a method to the keyboard hook so that you can set focus on the selection toolbar from the chart. Also enabling optional chaining.

What is missing from it behaving like a sheet is clicking outside to confirm selections, not sure we want this.

### DIFF
--- a/apis/nucleus/src/components/ActionsToolbar.jsx
+++ b/apis/nucleus/src/components/ActionsToolbar.jsx
@@ -119,8 +119,14 @@ const ActionsToolbar = ({
   useEffect(() => {
     if (!focusHandler) return;
 
-    const focusFirst = () => getEnabledButton(false)?.focus?.();
-    const focusLast = () => getEnabledButton(true)?.focus?.();
+    const focusFirst = () => {
+      const enabledButon = getEnabledButton(false);
+      enabledButon && enabledButon.focus();
+    };
+    const focusLast = () => {
+      const enabledButon = getEnabledButton(true);
+      enabledButon && enabledButon.focus();
+    };
     focusHandler.on('focus_toolbar_first', focusFirst);
     focusHandler.on('focus_toolbar_last', focusLast);
   }, []);
@@ -151,7 +157,7 @@ const ActionsToolbar = ({
 
   const tabCallback =
     // if keyboardNavigation is true, create a callback to handle tabbing from the first/last button in the toolbar that resets focus on the content
-    keyboardNavigation && focusHandler?.refocusContent
+    keyboardNavigation && focusHandler && focusHandler.refocusContent
       ? (evt) => {
           if (evt.key !== 'Tab') return;
           const isTabbingOut =
--- a/apis/nucleus/src/components/ActionsToolbar.jsx
+++ b/apis/nucleus/src/components/ActionsToolbar.jsx
@@ -91,6 +91,7 @@ const ActionsToolbar = ({
     anchorEl: null,
   },
   focusHandler = null,
+  actionsRefMock = null, // for testing
 }) => {
   const defaultSelectionActions = useDefaultSelectionActions(selections);
   const { itemSpacing } = useStyles();
@@ -105,8 +106,10 @@ const ActionsToolbar = ({
   const dividerStyle = useMemo(() => ({ margin: theme.spacing(0.5, 0) }));
 
   const getEnabledButton = (last) => {
-    if (!actionsRef.current) return null;
-    const buttons = actionsRef.current.querySelectorAll('button:not(.Mui-disabled)');
+    const actionsElement = actionsRef.current || actionsRefMock;
+    if (!actionsElement) return null;
+    const buttons = actionsElement.querySelectorAll('button:not(.Mui-disabled)');
+
     return buttons[last ? buttons.length - 1 : 0];
   };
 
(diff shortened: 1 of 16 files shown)


apis/nucleus/src/components/Cell.jsx
```diff
@@ -92,7 +93,7 @@ const LoadingSn = ({ delay = 750 }) => {
     return () => clearTimeout(handle);
   }, []);
 
-  return showLoading && <Loading />;
+  return showLoading ? <Loading /> : null;
```
haxxmaxx, line 96: lint fix

apis/nucleus/src/components/ActionsToolbar.jsx
```diff
@@ -90,24 +90,41 @@ const ActionsToolbar = ({
...
     setMoreEnabled(more.enabled);
   }, [more.enabled]);
 
+  useEffect(() => {
+    if (!focusHandler) return;
+
+    const focusFirst = () => getFirstEnabledButton(false)?.focus?.();
+    const focusLast = () => getFirstEnabledButton(true)?.focus?.();
+    focusHandler.on('focus_toolbar_first', focusFirst);
+    focusHandler.on('focus_toolbar_last', focusLast);
```
haxxmaxx, line 125: Don't think I need to remove these, since they are attached to `focusHandler`, and not to the DOM somewhere

apis/nucleus/src/components/Cell.jsx
```diff
@@ -282,12 +283,32 @@ const Cell = forwardRef(
...
+
+    focusHandler.current.blurCallback = (resetFocus) => {
+      halo.root.toggleFocusOfCells();
+      if (resetFocus && contentNode) {
+        contentNode.focus();
+      }
+    };
+    focusHandler.current.refocusContent = () => {
+      typeof state.sn.component?.focus === 'function' && state.sn.component.focus();
+    };
```
haxxmaxx, line 311: So the `focusHandler` instance need to stay the same, and `eventmixin` should only run the first time. The other methods I need to add afterwards. Not sure why but it worked this way

apis/nucleus/src/components/Cell.jsx
```diff
@@ -486,11 +498,19 @@ const Cell = forwardRef(
...
+              anchorEl={cellNode}
+              hovering={hovering}
+              focusHandler={focusHandler.current}
+            >
               &nbsp;
             </Header>
           )}
           <Grid
+            tabIndex={keyboardNavigation ? 0 : -1}
+            onKeyDown={keyboardNavigation ? handleKeyDown : null}
```
haxxmaxx, line 513: Moving these so it is not listening to events in the action toolbar

apis/nucleus/src/components/Header.jsx
```diff
@@ -71,9 +71,15 @@ const Header = ({ layout, sn, anchorEl, hovering }) => {
   const Toolbar = (
     <ActionsToolbar
       show={showToolbar}
-      selections={{ show: showInSelectionActions, api: sn.component.selections }}
+      selections={{
+        show: showInSelectionActions,
+        api: sn.component.selections,
+        onConfirm: focusHandler.refocusContent,
+        onCancel: focusHandler.refocusContent,
```
niekvanstaveren, line 78: Would this show when selecting with mouse as well?

apis/nucleus/src/components/ActionsToolbar.jsx
haxxmaxx: In cell.js, yes possibly... I was getting errors when I added returned a function with `focusHandler.remove`

apis/nucleus/src/components/Header.jsx
haxxmaxx: yes... It is not that pretty. I could try to make different key and click handlers. I think I ran into some problems when I tried before, but should be possible

niekvanstaveren: Would it be possible to check keyboard.active? Docs say that it is only active when space/enter is used to activate the chart

haxxmaxx: yes but it is set to false when you blur the chart which you do when you navigate to the toolbar...

niekvanstaveren: Maybe we can check the event instead. Not super pretty, but don't refocus on click events or something

apis/nucleus/src/components/ActionsToolbar.jsx
haxxmaxx: I just checked, it does not. It is wrapped in a useEffect that only runs the first render, which is when the chart is mounted, not every time the toolbar shows up

niekvanstaveren (approved):
Code looks good I think. Quite a lot of things going on. Could we add some specific unit tests for this behavior?

haxxmaxx:
> Code looks good I think. Quite a lot of things going on. Could we add some specific unit tests for this behavior?
Agree. Not straight forward but I'll try to write some where it is possible before merging

PR #706, opened 2021-11-15
chore: bump node version used in cicleci

T-Wizard:
should we remove the ignore in renovate.json?


veinfors:
> should we remove the ignore in renovate.json?

Sure, we can do that

veinfors (approved):

PR #709, opened 2021-11-22
fix: keyDown null when no keyboardAction

haxxmaxx:
## Motivation

Found a bug on the `ActionsToolbar`. `item.keyboardAction` would still run even if it is not defined. Added a sanity check.


T-Wizard (approved):

apis/nucleus/src/components/ActionsToolbarItem.jsx
```diff
@@ -22,12 +22,16 @@ const Item = React.forwardRef(({ item, addAnchor = false }, ref) => {
   const { hidden, disabled, style, hasSvgIconShape } = useActionState(item);
   if (hidden) return null;
 
+  const handleKeyDown = item.keyboardAction
+    ? (e) => ['Enter', ' ', 'Spacebar'].includes(e.key) && item.keyboardAction
```
niekvanstaveren, line 26: Does it still call the function?

haxxmaxx: Nope, fixed it

niekvanstaveren (approved):

PR #713, opened 2021-11-30
test: migrate integration tests to use fixture rendering

streamside:
## Motivation

Migrate integration tests to render visualization from a fixture instead of using Qlik engine.

* Table integration tests moved into subfolder to make it easier to add additional integration tests in a later stage. New path: `test/integration/table`.
* Table visualization test - which tests is performed on - moved to location of tests.
* Extracted common puppeteer related logic to `test/utils/puppeteerUtil.js`.


test/integration/table/table.int.js
```diff
@@ -1,10 +1,28 @@
...
+    s = await serve({
+      entry: path.resolve(__dirname, 'sn-table'),
+      open: false,
+      fixturePath: 'test/integration/table',
+    });
+    puppeteerUtil.addListeners(page);
+  });
+
+  after(() => {
+    s.close();
```
veinfors, line 19: Should you invoke `puppeteerUtil.removeListeners` from here maybe?

streamside: Good point. I have adjusted the code to remove the listeners after the test cases are completed.

veinfors (approved):

PR #714, opened 2021-11-30
refactor: remove feature to launch docker image if ACCEPT_EULA serve config parameter is true

LiKang6688:


streamside:
I noticed that the `commands/serve/package.json` has a reference to `docker-compose.yml` in the files property. Should it be removed?

LiKang6688:
> I noticed that the `commands/serve/package.json` has a reference to `docker-compose.yml` in the files property. Should it be removed?

Good find. Will do that.

PR #723, opened 2021-12-01
refactor(cli-create): migrate templates to use fixture rendering in tests

streamside:
## Motivation

Template projects are generated by running `npx nebula create --picasso <name>`. The templates "none" and "barchart" includes tests which requires a running Qlik engine. These tests have been updated to instead render the visualization from a fixture.

The build pipeline is updated to create, build and test created projects based on the source code being built.

The "barchart" tests included some nice interaction tests which unfortunately had to be removed as it's currently not supported. When adding support for it in EnigmaMocker this would be a good example to build a test case around.


LiKang6688:
I do not think it is integration testing anymore. what do you think?

commands/create/templates/sn/none/test/integration/fixtures/hello.fix.js
```diff
@@ -0,0 +1,17 @@
+export default function fixture() {
+  return {
+    type: 'hello',
+    genericObjects: [
```
streamside, line 4: It would be a nice improvement if it wasn't mandatory to supply a generic object when rendering a viz from a fixture.

streamside:
> I do not think it is integration testing anymore. what do you think?

Me and Li had a discussion about his I the outcome was that we believe that the support for rendering a visualization from a fixture can be used for various types of tests. It can be used for example for both integration and rendering tests.

veinfors (approved):

PR #724, opened 2021-12-01
refactor: remove ACCEPT_EULA option from nebula serve due to qlikcore engine deprecation

LiKang6688:
## Motivation

Remove `ACCEPT_EULA` option from `@nebula.js/cli-serve` since `qlikcore/engine` is not being published any longer (deprecated).
Remove `engine` (used to spin up `qlikcore/engine` docker image) option from start:mashup script
Remove sse (used to spin up a server-side extension for internal testing) option from start:mashup mashup script
Remove YAML files that contain `qlikcore/engine`
Remove `qlikcore/engine` related code and files


veinfors:
@LiKang6688 You can go ahead and remove `test/fixtures/data` as well.
And also `test/fixtures/sse` (server-side extension) and the usages of it since there is not an engine in place for this scenario any longer.

LiKang6688:
> @LiKang6688 You can go ahead and remove `test/fixtures/data` as well. And also `test/fixtures/sse` (server-side extension) and the usages of it since there is not an engine in place for this scenario any longer.

I think it is more suitable to delete those files in another PR https://github.com/qlik-oss/nebula.js/pull/723. As they are more related.

veinfors:
> > @LiKang6688 You can go ahead and remove `test/fixtures/data` as well. And also `test/fixtures/sse` (server-side extension) and the usages of it since there is not an engine in place for this scenario any longer.
> 
> @veinfors I think it is more suitable to delete those files in another PR #723. As they are more related, /app/ctrl00.qvf was not used anymore there.

Sure, that's fine to do it in that PR.

streamside:
Just remember something we talked about during the last weekly status meeting related to that we are breaking backward compatibility. Should we make a commit on this branch informing build systems etc that we are doing a breaking change?

LiKang6688:
> Just remember something we talked about during the last weekly status meeting related to that we are breaking backward compatibility. Should we make a commit on this branch informing build systems etc that we are doing a breaking change?

Yes, good reminder. We should

PR #728, opened 2021-12-06
refactor: update build targets

veinfors:
Update build target with which nebula.js is built:
* Remove ie 11
* Chrome 47 -> 62
* Safari 10.0 -> 11.0
* iOS 11.2 -> 12.2

Update build targets used by CLI build tools:
* Remove ie 11
* Chrome 47 -> 62


tasseKATT (approved):

T-Wizard:
should dropping IE11 support count as a breaking change?

PR #738, opened 2022-01-13
feat(Listbox): Enable sense-client integration through extended customisation possibilities

johanlahti:
**Note: I intend to split this PR into smaller ones.**

## Motivation

**To make it possible to use the Nebula Listbox from Sense, we require some changes to the API and capabilities:**
- Allow for a custom sessionModel and API
- Control over when data is refreshed
- Allow for a custom selections toolbar by hiding Nebula's own (new option: `toolbar=true/false`)

**Other:**
- Allow for drag (range) selections
- Quicker UI feedback by modifying the selections client-side first


apis/nucleus/src/components/listbox/ListBox/useSelectionInteractions.js
```diff
@@ -0,0 +1,109 @@
...
+      setSelectingValues(false);
+    });
+  });
+
+  document.addEventListener(
+    'mouseup',
+    useCallback(() => {
+      setMouseDown(false);
+    }, [model])
+  );
```
T-Wizard, line 65: I think this should be in a useEffect.

johanlahti: Yes, that makes sense. And also unregister the listener there.

veinfors:
Closing this draft for now, feel free to re-open if needed

PR #742, opened 2022-01-18
refactor(listbox): Enable integration of Listbox by providing additional options

johanlahti:
## Motivation

When using the Nebula Listbox from sense-client (or other containing application) we want to keep the **same session model** and **API**:
- Without a **custom session model** Nebula will create its own model which then lives its own life (not responding to selection events in sense-client and vice versa)
- A **custom API** lets the containing application (e.g. sense-client) design the API to behave in line with the containing application's selection flow.

To be able to integrate the Listbox in a containing application (e.g. sense-client) we need also to be able to **trigger a data refresh** from outside with an optional:
- `update function`

which will trigger a new refresh when called.

### New options:
```
- @param {object} [options.sessionModel] Use a custom sessionModel.
- @param {object} [options.selectionsApi] Use a custom selectionsApi to customize how values are selected.
- @param {ReceiverFunction} [options.update] A function which receives an update function which upon call will trigger a data fetch.
```


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -71,7 +82,13 @@ export function ListBoxInline({ app, fieldIdentifier, stateName = '$', options =
...
 
   const theme = useTheme();
-  const [model] = useSessionModel(listdef, app, fieldName, stateName);
+
+  let model;
+  if (sessionModel) {
+    model = sessionModel;
+  } else {
+    [model] = useSessionModel(listdef, app, fieldName, stateName);
+  }
```
T-Wizard, line 91: this breaks [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html)
hooks should not be called conditionally
and will probably need a more complicated workaround

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -83,7 +100,7 @@ export function ListBoxInline({ app, fieldIdentifier, stateName = '$', options =
 
   const { translator } = useContext(InstanceContext);
   const moreAlignTo = useRef();
-  const [selections] = useObjectSelections(app, model);
+  const selections = selectionsApi || useObjectSelections(app, model)[0];
```
T-Wizard, line 103: this breaks [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html)
hooks should not be called conditionally
and will probably need a more complicated workaround

apis/nucleus/src/components/listbox/ListBoxInline.jsx
johanlahti: I read up a bit and I found two typical workarounds:
- The first option is to *wrap* the conditional parts inside a component which is then executed on render (since this is a valid call)
- The second option is to always call the hook but then overwrite it conditionally

The first option is not viable from what I can see (we need this to happen *before* render). *Maybe* I can wrap the Listbox like this to acquire that behaviour:
```
<SessionModel> <-- Conditional logic wrapped in component
  <SelectionsAPI> <-- Conditional logic wrapped in component
    <ListBoxInline app={app} fieldIdentifier={fieldIdentifier} stateName={stateName} options={options} />,
  </SelectionsAPI>
</SessionModel>
```

The second option means unnecessary code will run on each hook call. However, assuming the hook does its job in cleaning up when unmounted (as hooks should do) I don't see any issue with this.

Thoughts?

apis/nucleus/src/components/listbox/ListBoxInline.jsx
johanlahti: Same as above (see my comment).

apis/nucleus/src/components/listbox/ListBoxInline.jsx
streamside: I think the solution you have in place now looks ok and not too over-engineered.

An alternative I've used in similar situations is to return different components based on whats available. Example to illustrate it in this context:

```js
const ListboxInLine = (props) => {
  // Listbox related logic
  return model ? <WithModel model={model} {...props} /> : <WithHook {...props} />;
};

const WithModel = ({model, ...props}) => {
  return <UI model={model} {...props} />;
}

const WithHook = ({...props}) => {
  const [model] = useSessionModel(...);
  return <UI model={model} {...props} />;
}

const UI = ({ model, ...props}) => {
  // Use `model` somehow 
  return (<Grid>...</Grid>);
}
```

Another approach is to wrap the `useSessionModel()` in your own hook which accepts the sessionModel along with the parameters to the existing hook. Might be a bit overkill in this case but it would make the code in the React component more concise:

```js
import useSessionModel from '....';

const useSessionModelWrapped = (sessionModel, listdef, app, fieldName, stateName) => {
  const hookSessionModel = useSessionModel(...);
  return sessionModel || hookSessionModel;
} 

const ListboxInline = () => {
  const model = useSessionModelWrapped(sessionModel, listdef, app, fieldName, stateName);
  // ...
}
```

johanlahti: Yep, I just pushed that change now. For the `useObjectSelections` I need to send in an empty object to prevent an error but it has same effect. Good suggestions.

examples/dev-mashup/package.json
```diff
@@ -8,13 +8,14 @@
...
     "@nebula.js/sn-bar-chart": "latest",
     "@nebula.js/sn-line-chart": "latest",
+    "@nebula.js/stardust": "latest",
     "enigma.js": "2.7.2"
   },
   "devDependencies": {
     "@babel/core": "7.12.3",
-    "parcel": "^2.0.0-rc.0"
+    "parcel": "^2.2.1",
+    "parcel-bundler": "^1.12.5"
```
johanlahti, line 19: I had to add the parcel-bundler to get the `dev-mashup` to run.

T-Wizard: it feel very wrong to use both 1.x and 2.x version of parcel.
why was it needed?
what whent wrong?

johanlahti: Try this:
```
cd examples/dev-mashup
yarn && yarn start
```

I get this error:
image

It works when including the `parcel-bundler`.

johanlahti: Unless you already have a better workaround, I suggest we keep this until finding a better solution. Right now it's broken.

T-Wizard: following the upgrade guide for Parcel 2
https://parceljs.org/getting-started/migration
(and remove parcel-bundler)

or

remove  `parcel": "^2.2.1`

I get cryptic `error Command failed with exit code 3221226356.` using parcel 2 on windows most of the time (but not allways) so using `parcel-bundler` may be safer (but deprecated) but then you should only have that dependency

johanlahti: I removed `parcel`, only using `parcel-bundler` now and works on my machine (Mac).

johanlahti:
✅

T-Wizard (approved):

PR #745, opened 2022-01-26
chore: update api compliance

joakim-ronning:
continues #700

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -154,7 +154,8 @@ jobs:
       - run:
           name: Run API Compliance
           command: >
-            VER=$(cat package.json | jq .version)
+            cd ~/enigma.js && VER=v$(cat  package.json | jq -r .version) && cd -
+            
             docker run --volumes-from specs
             -e SPEC_PATHS="abdc47c2-921a-45d3-ba5c-80d448574c5b@/specs/spec.json"
             -e COMMIT_SHA="$CIRCLE_SHA1"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,6 +145,18 @@ jobs:
       - attach_workspace:
           at: ~/project
       - run:
+        name: Create version.txt
+        command: |
+          set -x
+          if [ -n "${CIRCLE_TAG}" ]; then
+            version=${CIRCLE_TAG#v}
+          else
+            version=$(git describe --tags --abbrev=7 --match "v*")
+            version=${version#v}
+          fi
+          echo "$version" > ./version.txt
+          echo "Building $version"
+      - run:
           name: Prepare API Compliance
           command: |
             docker pull ghcr.io/qlik-download/api-compliance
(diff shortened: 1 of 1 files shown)


joakim-ronning:
Looks like the changes were successful, but the api spec itself is not passing the api compliance. @Caele could you take a look at it?

veinfors:
> Looks like the changes were successful, but the api spec itself is not passing the api compliance. @Caele could you take a look at it?

Hello @joakim-ronning! I looked at the failures, and they seem fine. It's mostly complaining because of the version has not been bumped yet. We should change when/how we run the api gov. check, so that we either only run it after the version number has been updated/when making a new release, or we would have to add some mechanism that "temporarily" bumps the version numbers for running the api gov checks, otherwise we will always get failures when we make changes (even if they don't actually break anything in the api).

When thing that confused me a little bit when looking at the failure was this summary that looked like this:
```
Summary
info:    	Changes
info:    	- Added: 1
info:    	- Updated: 3
info:    	- Deprecated: 0
info:    	- Removed: 0
info:    	Violations
info:    	- Breaking changes: 0
info:    	- Deprecation: 0
info:    	- Stability: 0
info:    	- Versioning: 1
info:    	- Visibility: 0
```

There it says there are 3 updates and 1 addition, but when I look closer at the failures there are actually 2 additions and 1 update. That is weird, something wrong with that summary?

Anyway, there are 2 additions which are fine, and the only update seem to be on the following example:
```
"op": "replace",
  "path": "#/definitions/Theme/entries/getStyle/examples/0",
  "value": "theme.getStyle('object', 'title.main', 'fontSize');\ntheme.getStyle('', '', 'fontSize');"
```

it surprises me a little that examples causes the api check to fail though...


I think it's fine to go ahead and merge this PR though.

joakim-ronning:
> > Looks like the changes were successful, but the api spec itself is not passing the api compliance. @Caele could you take a look at it?
> 
> Hello @joakim-ronning! I looked at the failures, and they seem fine. It's mostly complaining because of the version has not been bumped yet. We should change when/how we run the api gov. check, so that we either only run it after the version number has been updated/when making a new release, or we would have to add some mechanism that "temporarily" bumps the version numbers for running the api gov checks, otherwise we will always get failures when we make changes (even if they don't actually break anything in the api).
> 
> When thing that confused me a little bit when looking at the failure was this summary that looked like this:
> 
> ```
> Summary
> info:    	Changes
> info:    	- Added: 1
> info:    	- Updated: 3
> info:    	- Deprecated: 0
> info:    	- Removed: 0
> info:    	Violations
> info:    	- Breaking changes: 0
> info:    	- Deprecation: 0
> info:    	- Stability: 0
> info:    	- Versioning: 1
> info:    	- Visibility: 0
> ```
> 
> There it says there are 3 updates and 1 addition, but when I look closer at the failures there are actually 2 additions and 1 update. That is weird, something wrong with that summary?
> 
> Anyway, there are 2 additions which are fine, and the only update seem to be on the following example:
> 
> ```
> "op": "replace",
>   "path": "#/definitions/Theme/entries/getStyle/examples/0",
>   "value": "theme.getStyle('object', 'title.main', 'fontSize');\ntheme.getStyle('', '', 'fontSize');"
> ```
> 
> it surprises me a little that examples causes the api check to fail though...

Yeah. I think for now the solution is to know what type of commit the update is and change the version manually (in this case, a minor should suffice).  Api Compliance only reads the specs, so any changes must be made beforehand.

veinfors (approved):

PR #747, opened 2022-01-27
refactor(listbox): instant selection feedback on click

johanlahti:
## Motivation

- Get instant selection feedback on click (replaced by `onMouseDown`) by modifying the pages directly for better user experience
- This is also needed for range selections (next PR)

### DIFF
--- a/apis/nucleus/src/components/listbox/listbox-selections.js
+++ b/apis/nucleus/src/components/listbox/listbox-selections.js
@@ -33,10 +33,9 @@ export function applySelectionsOnPages(pages, elmNumbers, toggle = false) {
       const qState = selectionMatchesElement ? getNewSelectionState(p0.qState) : p0.qState;
       return [{ ...p0, qState }, p.slice(1)];
     });
-    return qMatrix;
+    return { ...page, qMatrix };
   });
-  const newPages = pages.map((page, i) => ({ ...page, qMatrix: matrices[i] }));
-  return newPages;
+  return matrices;
 }
 
 export async function selectValues({ selections, elemNumbers, isSingleSelect = false }) {
--- a/apis/nucleus/src/components/listbox/listbox-selections.js
+++ b/apis/nucleus/src/components/listbox/listbox-selections.js
@@ -7,7 +7,7 @@ function isStateSelected(qState) {
 }
 
 export function getUniques(arr) {
-  return Array.from(new Set(arr));
+  return Array.isArray(arr) ? Array.from(new Set(arr)) : undefined;
 }
 
 export function getSelectedValues(pages) {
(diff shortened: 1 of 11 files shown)


apis/nucleus/src/components/listbox/useSelectionsInteractions.js
```diff
@@ -0,0 +1,70 @@
...
+
+  const select = ({ elemNumbers }) => {
+    setSelectingValues(true);
+    const isSingleSelect = layout.qListObject.qDimensionInfo.qIsOneAndOnlyOne;
+    return selectValues({ selections, elemNumbers, isSingleSelect }).then(() => {
+      setSelectingValues(false);
+    });
+  };
+
+  const onMouseDown = useCallback(
```
T-Wizard, line 18: Consider moving some logic inside onMouseDown. (direct or by function call)
instead of triggering adittional useEffects (I hope it would make the code easier to read)

johanlahti: Yes, it looks a bit unmotivated for simple selections. I think this will make more sense when adding the code for range selections since it will then divide the handling of performing selections vs preparing for selections. But I will keep that in mind for the next PR.

apis/nucleus/src/components/listbox/listbox-selections.js
```diff
@@ -1,7 +1,46 @@
...
+    const qMatrix = page.qMatrix.map((p) => {
+      const [p0] = p;
+      const selectionMatchesElement = elmNumbers.includes(p0.qElemNumber);
+      const qState = selectionMatchesElement ? getNewSelectionState(p0.qState) : p0.qState;
+      return [{ ...p0, qState }, p.slice(1)];
+    });
+    return qMatrix;
+  });
+  const newPages = pages.map((page, i) => ({ ...page, qMatrix: matrices[i] }));
+  return newPages;
```
T-Wizard, line 39: maybe someting like
```suggestion
    return { ...page, qMatrix };
  });
  return matrices;
```
(to avoid a extra .map call)

apis/nucleus/src/components/listbox/listbox-selections.js
```diff
@@ -1,7 +1,46 @@
...
+      const [p0] = p;
+      return isStateSelected(p0.qState) ? p0.qElemNumber : false;
+    });
+    return elementNumbers.filter((n) => n !== false);
+  });
+  return flattenArr(elementNbrs);
+}
+
+export function applySelectionsOnPages(pages, elmNumbers, toggle = false) {
+  const getNewSelectionState = (qState) => (toggle && elmNumbers.length === 1 && isStateSelected(qState) ? 'A' : 'S');
```
T-Wizard, line 28: why is elmNumbers.length === 1 relevant?

I think there is a lot of improvment possible on guessing what the new state will be.
but that is also true for the current listbox and probably out of scope for this PR.

T-Wizard (approved):

apis/nucleus/src/components/listbox/listbox-selections.js
johanlahti: We only want to toggle selected values on "click". For range selection (>1 values) we don't want to toggle selected values.

apis/nucleus/src/components/listbox/listbox-selections.js
johanlahti: 👍  Nice one

PR #749, opened 2022-02-02
feat: add --carbon flag to build react-native supernova packages

vcellu:
## Motivation

For up coming react-native supernova's, there needs a way to build packages for react-native.  This is still a work in progress.   This is just one phase. It is undocumented on purpose.  More documentation on this flag will come once the full workflow is completed.

### DIFF
--- a/commands/build/lib/build.js
+++ b/commands/build/lib/build.js
@@ -163,7 +163,6 @@ const config = ({
       file: path.resolve(dir, fileTarget), // fileTargetformat === 'esm' && pkg.module ? pkg.module : pkg.main,
       name: moduleName,
       sourcemap,
-      plugins: carbonCopy,
       globals: {
         '@nebula.js/stardust': 'stardust',
       },
--- a/commands/build/lib/build.js
+++ b/commands/build/lib/build.js
@@ -3,7 +3,6 @@ const chalk = require('chalk');
 const readline = require('readline');
 
 const fs = require('fs');
-const copy = require('rollup-plugin-copy');
 const extend = require('extend');
 const yargs = require('yargs');
 const rollup = require('rollup');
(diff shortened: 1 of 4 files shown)


commands/build/lib/build.js
```diff
@@ -28,6 +31,20 @@ const resolveReplacementStrings = (replacementStrings) => {
   return replacementStrings;
 };
 
+const doINeedACarbonCopy = (argv) => {
+  const { carbon } = argv;
```
veinfors, line 35: Could we call it `reactNative` instead of `carbon`? That will make it a lot easier/useful if someone else want's to use it...

commands/build/lib/build.js
```diff
@@ -36,11 +53,23 @@ const config = ({
   core,
 } = {}) => {
   const CWD = argv.cwd || cwd;
+  const carbon = doINeedACarbonCopy(argv);
   let dir = CWD;
   let pkg = require(path.resolve(CWD, 'package.json')); // eslint-disable-line
   const corePkg = core ? require(path.resolve(core, 'package.json')) : null; // eslint-disable-line
+  pkg = carbon ? require(path.resolve('./react-native', 'package.json')) : pkg; // eslint-disable-line
```
veinfors, line 60: instead of having this path hardcoded to `./react-native/package.json`, could we let the user specify the path in a similar way as we do for the `core` package?

commands/build/lib/build.js
```diff
@@ -123,6 +154,7 @@ const config = ({
               })
             : false,
         ],
+        ...carbonCopy,
```
veinfors, line 157: Is this necessary to have in this build? Or could it be done by the one building the chart outside of nebula.js?
We do similar "automatic" copy operations (for convenience) in some repos when building, but that's handled outside of nebula.js... see https://github.com/qlik-oss/sn-scatter-plot/blob/main/scripts/copy-ext.js for instance.

It might be a little different when developing for react native though, so I might miss some aspect of it...

vcellu: The idea is that we want to have a workflow that doesn't require the user to think too much.  For example, I do a nebula --build --reactNative and I get a package in my react-native folder, without any more interference.   But, the copy is not necessary and we can have the build command output to the target folder.

vcellu: Done!

commands/build/lib/build.js
vcellu: Done!

commands/build/lib/build.js
vcellu: Done!

veinfors:
I looks good to me @vcellu! You just need to run yarn again after you removed the copy-plugin to update `yarn.lock`, then it's good to go!

PR #750, opened 2022-02-02
docs: update jsdocs enabling possibility to generate typescript defs

streamside:
## Motivation

Update JSDoc with the target to make it possible to generate typescript definitions.

See [nebula.js typescript definitions](https://gist.github.com/streamside/c625101902aa1ff3e105e929fa46c8a3) created based on changes in this PR.

####  References to Engine API

References to enigma / qae in the documentation is updated to use the EngineAPI published in the package [@types/qlik-engineapi](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/qlik-engineapi/index.d.ts).

For example, referencing the generic object layout uses `EngineAPI.IGenericObjectLayout` (previously `qae.GenericObjectLayout`).

Mapping table:
```
qae.GenericObjectLayout     => EngineAPI.IGenericObjectLayout
qae.GenericObjectProperties => EngineAPI.IGenericObjectProperties
qae.NxAppLayout             => EngineAPI.INxAppLayout
qae.NxDimension             => EngineAPI.INxDimension
qae.NxMeasure               => EngineAPI.INxMeasure
enigma.Doc                  => EngineAPI.IApp
enigma.GenericObject        => EngineAPI.IGenericObject
enigma.Global               => EngineAPI.IGlobal
```

The majority of the changes in the PR is related to this which also causes the api specification to deviate significantly. An alternate solution is to keep the old references and do the mapping when generating TS defs from the api specification.

#### Attributes on `VizElement`

Move attributes in `VizElement` to interface extending `NamedNodeMap`.

#### Add type `ThemeJSON` for ThemeInfo

Add type definition for `ThemeJSON`. It's returned as the generic type when loading a theme.

#### Return value from `useRenderState()`

Add interface `RenderState` which is returned from `useRenderState()`.

#### Callbacks on `FieldTarget`

Callbacks extracted to separate callbacks in order to be able to define parameter names.


veinfors:
I like this idea of pointing the "engine types" towards `EngineAPI.IGenericObjectLayout` etc. It feels like the right thing to do.

I didn't know about those typescript defs being available for the QIX Api though and I'm curious about how "official"/reliable those definitions are. Are they being published by Qlik or is it some third party/private person who have just put it together and published it... I just want to know that it's future proof and will be updated and maintained. Anyone who knows more about that?

apis/nucleus/spec/spec.conf.js
```diff
@@ -12,19 +12,19 @@ module.exports = {
   },
   parse: {
     types: {
-      'enigma.Doc': {
+      'EngineAPI.IApp': {
         url: 'https://github.com/qlik-oss/enigma.js/blob/master/docs/api.md#generated-api',
       },
-      'qae.GenericObjectLayout': {
+      'EngineAPI.IGenericObjectLayout': {
         url: 'https://core.qlik.com/services/qix-engine/apis/qix/definitions/#genericobjectlayout',
```
veinfors, line 19: While we are touching this, we can update these links, which are not working, as well. So instead of pointing towards core.qlik... we can point this to https://qlik.dev/apis/json-rpc/qix/schemas#%23%2Fdefinitions%2Fschemas%2Fentries%2FGenericObjectLayout for instance (since qlik core has been deprecated).

streamside:
> I like this idea of pointing the "engine types" towards `EngineAPI.IGenericObjectLayout` etc. It feels like the right thing to do.
> 
> I didn't know about those typescript defs being available for the QIX Api though and I'm curious about how "official"/reliable those definitions are. Are they being published by Qlik or is it some third party/private person who have just put it together and published it... I just want to know that it's future proof and will be updated and maintained. Anyone who knows more about that?

It was decided a while back that the type defs for enigma shouldn't be included in its code base (see [PR](https://github.com/qlik-oss/enigma.js/issues/26)). It was promoted that the type defs for enigma should be published to the github project DefinitelyTyped.

The type definitions at [@types/qlik-engineapi](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c5982f2ba79c1bebb301211e9709c979788d23f4/types/qlik-engineapi/index.d.ts) looks to me to be of good quality. It would be good if someone with engine knowledge could verify this.

The other option is to omit references to the Engine API and accept "any" as type instead. That would impact the benefit the type definitions offer the developers.

apis/nucleus/spec/spec.conf.js
streamside: I looked for these urls but couldn't find any suitable ones. I'll up them!

veinfors (approved):

PR #751, opened 2022-02-02
refactor(listbox): range selections

johanlahti:
## Motivation

Changes:
- Range selections with instant feedback
- Unit tests

Possible improvement (for later):
- A tiny side-effect of having direct response on toggle is that: when you start selecting a range – and that first value is already selected – it will initially be rendered as unselected *until you start dragging the range*
- This can be improved later on by setting the `isRangeSelection` flag already when moving the cursor beyond a certain "click threshold" (this seems to be how it behaves in the existing Listbox)

https://user-images.githubusercontent.com/5780544/152346586-bd21519f-544e-46c4-85d1-c7910a19379e.mov


apis/nucleus/src/components/listbox/listbox-selections.js
```diff
@@ -53,3 +53,49 @@ export async function selectValues({ selections, elemNumbers, isSingleSelect = f
...
+export function getElemNumbersFromPages(pages) {
+  if (!pages || !pages.length) {
+    return [];
+  }
+  const elemNumbersArr = pages.map((page) => {
+    const qElemNumbers = page.qMatrix.map((p) => {
+      const [{ qElemNumber }] = p;
+      return qElemNumber;
+    });
+    return [...qElemNumbers];
```
T-Wizard, line 66:
```suggestion
    return qElemNumbers;
```

T-Wizard (approved):

PR #763, opened 2022-02-14
fix(jsdoc): mark mount and unmount  as instance functions

T-Wizard:


streamside:
Does these changes result in any changes in the TS definitions?

T-Wizard:
> Does these changes result in any changes in the TS definitions?

They should.
(should an update of TS definition file be included in this PR)

streamside:
> > Does these changes result in any changes in the TS definitions?
> 
> They should. (should an update of TS definition file be included in this PR)

I believe it would be nice if they are included directly in the PR. It will make it every commit on master more atomic and possible to make a build from (imho!)

Me and @veinfors just had a chat about adding a precommit hook to ensure that specs and ts defs are up to date. This will catch potential problems as early at possible.

PR #764, opened 2022-02-14
refactor(Listbox): Checkboxes

johanlahti:
## Motivation

Enable displaying listbox values as checkboxes. This can be achieved by setting the Listbox option: ```checkboxes: true```.

### Changes:
- Checkboxes use the `onClick` event instead of `mousedown` and `mouseup` for pre-selection and selection

### Other:
- The `Row` and `Column` components have been merged into one (`ListBoxRowColumn`) since they previously duplicated most logic
- Fix broken locked state
- Unit tests

*Below: Demo of checkboxes, field selections and locked state:*

https://user-images.githubusercontent.com/5780544/154054882-e0758749-929b-4ab7-8fe7-084838cd8979.mov


apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -132,9 +133,12 @@ export default function ListBox({
   }
   const isVertical = listLayout !== 'horizontal';
   const count = layout.qListObject.qSize.qcy;
-  const ITEM_SIZE = isVertical ? 33 : 200;
+  const SIZE_VERTICAL = 40;
+  const ITEM_SIZE = isVertical ? SIZE_VERTICAL : 200;
```
T-Wizard, line 137: It looks like ITEM_SIZE is changed from 33 to 44 even when not using checkboxes. why?

johanlahti: `40px` is from the design I got. I can double-check just to make sure.

johanlahti: I will take a broader look at the design in a subsequent PR when I have gotten more details. I think we can go with this for now.

T-Wizard: I think it should be an opt in to not affect any current use of nebula.js listbox out there

johanlahti: Yes, you're right. We can keep it as it was while awaiting the precise design specs.

johanlahti: @T-Wizard I have changed so that the larger row height (`40px`) is only used for checkboxes.
If the design suggests we should increase the height then we can bring up what you mentioned, whether this is a "breaking" change or not.

johanlahti:
🟢

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -19,7 +18,8 @@ export default function ListBox({
   height,
   width,
   listLayout = 'vertical',
-  rangeSelect = true,
+  rangeSelect = false,
+  checkboxes = false,
```
T-Wizard, line 22: I am unsure if I think we should enable rangeSelect by default (we should atleast do it for the next major version)
but if not. do enable it for nebula serve.

johanlahti: Ok, I can turn it off for now

apis/nucleus/src/index.js
```diff
@@ -346,8 +346,9 @@ function nuked(configuration = {}) {
            * @param {string=} options.title Custom title, defaults to fieldname
            * @param {string=} [options.direction=ltr] Direction setting ltr|rtl.
            * @param {string=} [options.listLayout=vertical] Layout direction vertical|horizontal
-           * @param {boolean=} [options.search=true] To show the search bar
-           * @param {boolean=} [options.toolbar=true] To show the toolbar
+           * @param {boolean=} [options.search=true] Show the search bar
+           * @param {boolean=} [options.toolbar=true] Show the toolbar
+           * @param {boolean=} [options.checkboxes=false] Show values as checkboxes instead of as fields
```
T-Wizard, line 351: I don't see any rangeSelect option. Is it missing?

johanlahti: ✅

T-Wizard (approved):
The search highlight did look odd for me when checkbox mode was used.

johanlahti:
Ok, I will look at that in an upcoming PR. Thanks for the feedback.

PR #766, opened 2022-02-17
refactor(Listbox): Make search algorithm account for multiple ranges

johanlahti:
## Motivation

The search algorithm returned incorrect segments when multiple ranges are returned from Engine.
Also the highlighting class was applied on the checkbox instead of the label, which is fixed with this PR.

### Changes:
- Account for multiple search ranges in the algorithm
- Display starting or trailing white-space on search
- Fix ellipsis for highlights and non-highlighted labels
- Render highlights with white-space intact but without wrapping lines
- Unit tests

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxRowColumn.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxRowColumn.jsx
@@ -49,7 +49,6 @@ const useStyles = makeStyles((theme) => ({
   // The leaf node, containing the label text.
   labelText: {
     flexBasis: 'max-content',
-    fontSize: '0.9rem',
     lineHeight: '16px',
     userSelect: 'none',
     whiteSpace: 'pre', // to keep white-space on highlight
@@ -149,6 +148,7 @@ export default function RowColumn({ index, style, data, column = false }) {
   const getValueField = ({ lbl, ix, color, highlighted = false }) => (
     <Typography
       component="span"
+      variant="body2"
       key={ix}
       className={[classes.labelText, !!highlighted && classes.highlighted]
         .filter((c) => !!c)
--- a/apis/nucleus/src/components/listbox/ListBoxRowColumn.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxRowColumn.jsx
@@ -12,6 +12,7 @@ const useStyles = makeStyles((theme) => ({
   row: {
     flexWrap: 'nowrap',
     borderBottom: `1px solid ${theme.palette.divider}`,
+    color: theme.palette.text.primary,
     '&:focus': {
       boxShadow: `inset 0 0 0 2px ${theme.palette.custom.focusOutline}`,
       outline: 'none',
@@ -20,6 +21,7 @@ const useStyles = makeStyles((theme) => ({
   column: {
     flexWrap: 'nowrap',
     borderRight: `1px solid ${theme.palette.divider}`,
+    color: theme.palette.text.primary,
     '&:focus': {
       boxShadow: `inset 0 0 0 2px ${theme.palette.custom.focusOutline}`,
       outline: 'none',
@@ -32,16 +34,14 @@ const useStyles = makeStyles((theme) => ({
     alignItems: 'center',
     minWidth: 0,
     flexGrow: 1,
+    paddingLeft: '6px',
+    paddingRight: '6px',
     '& span': {
+      fontSize: '0.9rem',
       whiteSpace: 'nowrap',
-      fontSize: '12px',
       lineHeight: '16px',
       userSelect: 'none',
     },
-  },
-  valueLabel: {
-    paddingLeft: '6px',
-    paddingRight: '6px',
   },
   icon: {
     display: 'flex',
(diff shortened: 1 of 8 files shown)


apis/nucleus/src/components/listbox/listbox-search.js
```diff
@@ -0,0 +1,59 @@
...
+ * @param {string} label The label we want to create segments out of.
+ * @param {Range[]} ranges The ranges defining indices for cutting the string into segments.
+ * @returns {Segment[]} An array of segments in the same order as the ranges.
+ */
+export default function getSegmentsFromRanges(label, ranges) {
+  if (!ranges.length) {
+    return [];
+  }
+  const labels = ranges.reduce((acc, curr, ix) => {
+    // First non highlighted segment
```
T-Wizard, line 44: remove comment? (I don't think is is correct anymore)

apis/nucleus/src/components/listbox/ListBoxRowColumn.jsx
```diff
@@ -115,25 +119,34 @@ export default function RowColumn({ index, style, data, column = false }) {
...
+      ) : (
+        lbl
+      );
     return (
       <FormControlLabel
         color={color}
         control={cb}
-        className={classes.checkboxLabel}
-        label={<Typography>{lbl}</Typography>}
+        className={[classes.checkboxLabel].join(' ').trim()}
```
T-Wizard, line 136: why there should be no to create an array with a single element and then join and trim it

T-Wizard:
some of the styling changes makes the checkboxes of center in my testing

johanlahti:
I will clean up and add some more unit tests as well.

apis/nucleus/src/components/listbox/ListBoxRowColumn.jsx
```diff
@@ -20,36 +28,62 @@ const useStyles = makeStyles((theme) => ({
...
-      userSelect: 'none',
+    // Note that this padding is overridden when using checkboxes.
+    paddingLeft: '12px',
+    paddingRight: '12px',
+  },
+
+  // The leaf node, containing the label text.
+  labelText: {
+    flexBasis: 'max-content',
+    fontSize: '0.9rem',
```
veinfors, line 52: Any specific reason to why you are using rem here instead of px?
(At some point I guess we'd like to tie it to the theme as well...)

apis/nucleus/src/components/listbox/listbox-highlight.js
```diff
@@ -0,0 +1,57 @@
...
+ * @property {string} segment The sub-string/segment cut out from the original label.
+ * @property {boolean} highlighted A flag which tells whether the segment should be highlighted or not.
+ */
+
+/**
+ * @param {string} label The label we want to create segments out of.
+ * @param {Range} range The indexes which define how to create the segments.
+ * @param {number=} [startIndex] An optional index which tells where we want to start the first segment from
+ *   (only relevant for creating the first unhighlighted segment of a string/sub-string).
+ * @returns {Segment} A segment.
```
veinfors, line 18:
```suggestion
 * @returns {Segment} An array of segments.
```

apis/nucleus/src/components/listbox/listbox-highlight.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * @typedef {object} Range
+ * @property {number} qCharPos The (absolute) index where the highlighted range starts.
+ * @property {number} qCharCount The length of the sub-string (starting from qChartPos) that should be highlighted.
+ */
+
+/**
+ * @typedef {Array<{segment: string, highlighted: boolean}>} Segment
```
streamside, line 8: Defining objects inline will cause problems when generating typescript definitions. It looks like the typedef can be changed to:

```
@typedef {object} Segment
```

And when using it around line 18:

```
@returns {Segment[]} The segments
```

apis/nucleus/src/components/listbox/ListBoxRowColumn.jsx
johanlahti: Not really, but `rem` will automatically adapt to the root font-size automatically. Is there any reason we want `px` instead of `rem`?

apis/nucleus/src/components/listbox/listbox-highlight.js
johanlahti: I will change.

apis/nucleus/src/components/listbox/ListBoxRowColumn.jsx
johanlahti: I think the general advice is to use adaptive font-sizes but we tend to only use `px` in our implementations. Do you know if we have any "best practices" for this at Qlik?

streamside: I try to stick with the theme configured for Material UI. In this case there is a set a typography variants specified so I would pick a suitable one from the list and apply its styling. If the size, colors or other styling is desired to be altered the theme config can be updated.

In the storybook for sprout there is a list of available typography variants:
http://rd-sprout.qliktech.com/sprout/?path=/story/mui-components-typography--variants

johanlahti: I've switched to using `variant="body2"` which seems to be closest to what I already used (`0.875rem` instead of `0.9rem`).

veinfors (approved):

PR #767, opened 2022-02-17
docs: add missing context in createconfiguration

LiKang6688:
## Motivation

```
embed(app, {
  constraints: {
    active: true, // do not allow interactions
  }
})
```
should be
```
embed(app, {
  context: {
    constraints: {
      active: true, // do not allow interactions
    }
  }
})
```
```
const d = m.createConfiguration({
 theme: 'dark'
});
```
should also be
```
const d = m.createConfiguration({
  context: {
    theme: 'dark'
  }
});
```

### DIFF
--- a/apis/stardust/api-spec/spec.json
+++ b/apis/stardust/api-spec/spec.json
@@ -43,7 +43,7 @@
             "type": "#/entries/embed"
           },
           "examples": [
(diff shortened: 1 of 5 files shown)


apis/stardust/api-spec/spec.json
```diff
@@ -790,13 +790,25 @@
...
+                },
+                "checkboxes": {
+                  "description": "Show values as checkboxes instead of as fields",
+                  "optional": true,
+                  "defaultValue": false,
+                  "type": "boolean"
+                },
+                "rangeSelect": {
+                  "description": "Enable range selection",
                   "optional": true,
```
LiKang6688, line 812: I think that there are coming from https://github.com/qlik-oss/nebula.js/pull/764#partial-pull-merging

veinfors (approved):

apis/stardust/api-spec/spec.json
T-Wizard: the type definition file should probably also be updated

T-Wizard (approved):

apis/stardust/api-spec/spec.json
LiKang6688: Good point. Thank you. Updated.

PR #768, opened 2022-02-22
refactor: Enable global Material UI CSS classes

johanlahti:
## Motivation

We need global CSS classes to be able to customise the Material UI components when components are not directly exposed. Also, global classes opens up for easier CSS customisations by the Nebula API *users*, in addition to the hard-coded classes that we explicitly add to our own components. However – this is not bullet-proof unless we also expose the prefix to the user.

With the changes in this PR we can do:
```
const MySwitch = styled(Switch)(({ theme }) => {
  const { prefix } = theme;
  return {
    width: '34px',
    height: '30px',
    padding: '7px 0',
    [`& ${prefix}-MuiSwitch-switchBase`]: {
      // ... custom styling
  };
```

Changes:
- Enable global classes
- Expose the prefix (it contains an incremental component and cannot be hard-coded)

Limitations:
- An outside user still cannot access the incremental prefix (now exposed through the theme). If the user needs to do this (which is not supported) they can add an "incrementor" themselves or use CSS selectors based on contained elements.


johanlahti:
Closing this PR since we can use workarounds (in *MUI 4*) such as:
```js
const StyledSwitch = styled(Switch)(({ theme, iconOn, iconOff }) => ({
  '&.qlik-switch-root': {
    opacity: 1,
    width: '34px !important',
    height: '30px !important',
    padding: '7px 0 !important',
  },
  '& .qlik-switch-colorSecondary': {
    color: '#FFFFFF',
    backgroundColor: 'inherit',
    '&.qlik-switch-checked:hover': {
      backgroundColor: 'inherit',
    },
  },
}));
return (
    <StyledSwitch
      classes={{
        root: `qlik-switch-root`,
        track: 'qlik-switch-track',
        thumb: 'qlik-switch-thumb',
        checked: 'qlik-switch-checked',
        switchBase: 'qlik-switch-switchBase',
        colorSecondary: 'qlik-switch-colorSecondary',
      }}
      onChange={onChange}
    />
);
```

PR where this was needed: https://github.com/qlik-oss/nebula.js/pull/771/files#diff-df8cd78ca22919265585910ea0dcc11c784b64cd1004e487a2f8f41de1a6fe7aR70-R77

PR #769, opened 2022-02-22
feat(listbox): dense mode with reduced padding and text size

streamside:
## Motivation

Add `dense` option for listbox component. When running in dense mode the padding and text size is reduced.

Add static class names to the search input and for each value. The class names are `search` and `value`. This enables possibility to override styling.

### Examples

*Normal mode (as-is today)*
image

*Dense mode*
image

*With custom styling*
image

### DIFF
--- a/apis/nucleus/src/index.js
+++ b/apis/nucleus/src/index.js
@@ -350,6 +350,7 @@ function nuked(configuration = {}) {
            * @param {boolean=} [options.toolbar=true] Show the toolbar
            * @param {boolean=} [options.checkboxes=false] Show values as checkboxes instead of as fields
            * @param {boolean=} [options.rangeSelect=true] Enable range selection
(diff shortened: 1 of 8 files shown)


apis/nucleus/src/index.js
```diff
@@ -350,6 +350,7 @@ function nuked(configuration = {}) {
            * @param {boolean=} [options.toolbar=true] Show the toolbar
            * @param {boolean=} [options.checkboxes=false] Show values as checkboxes instead of as fields
            * @param {boolean=} [options.rangeSelect=true] Enable range selection
+           * @param {boolean=} [options.dense] Reduces padding and text size
```
johanlahti, line 353:
```suggestion
           * @param {boolean=} [options.dense=false] Reduces padding and text size
```

johanlahti (approved):

PR #771, opened 2022-02-23
refactor(Listbox): Toggle switch

johanlahti:
## Motivation

Add a switch (toggle) button to the Listbox.

image

### Capabilities:
- Toggle any supported boolean option (currently `['checkboxes', 'search', 'toolbar', 'rangeSelect', 'dense']`)
- Toggle multiple options at once using an optional `onChange` function which has access to a setter function (`setOptions`)
- Styled with Qlik's green and gray colors
- Possibility of defining label, helper text and custom Switch styling (e.g. icon)


apis/nucleus/src/components/listbox/QlikSwitchButton/QlikSwitch.jsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import { makeStyles } from '@material-ui/core';
+import Switch from '@material-ui/core/Switch';
+
+export default function QlikSwitch({ checked, onChange, iconOn, iconOff }) {
+  const styles = makeStyles((theme) => ({
```
streamside, line 6: I suggest to move `makeStyles()` outside the React component and pass the props to `useStyles(props)`. The [docs](https://mui.com/styles/basics/#adapting-based-on-props) explains in more detail how to do it.

Moving to MUI5 would require this to be written in either `sx` or `styled()`. It would be interesting to try to write this with styled.

apis/nucleus/src/index.js
```diff
@@ -332,6 +332,44 @@ function nuked(configuration = {}) {
          * @typedef {function(function)} ReceiverFunction A callback function which receives another function as input.
          */
 
+        /**
+         * @typedef {object} SwitchButtonConfig
+         * @param {string} SwitchButtonConfig.option The option to be toggled. Currently supporting: 'checkboxes', 'search', 'toolbar', 'rangeSelect', 'dense'.
```
streamside, line 337: I believe `SwitchButtonConfig.` should be omitted from the param names. For example, `SwitchButtonConfig.option` becomes `option`.

Running `yarn spec && yarn ts` in the `<repo>/apis/stardust` updates the spec and ts defs. It could be worth running them and have a look at them so they are correct.

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -29,6 +31,8 @@ export default function ListBoxPortal({ app, fieldIdentifier, stateName, element
 }
 
 export function ListBoxInline({ app, fieldIdentifier, stateName = '$', options = {} }) {
+  const [opts, setOptions] = useState({ ...options });
```
streamside, line 34: As all options are stored as state I would test the component to make sure that it isn't rerendered unnecessary. Another approach would be to only store the configurable settings in state and take the rest directly from options ?

apis/nucleus/src/components/listbox/QlikSwitchButton/QlikSwitch.jsx
johanlahti: I wrote this with `styled` to start with but then we got the problem with the prefixes not being added, which is why I created https://github.com/qlik-oss/nebula.js/pull/768 . Then, we did not want that workaround, and I was advised to use `makeStyles`. So yeah the loop has been completed 😄  
I could potentially break it out but I spent significant time on adapting the styles to makeStyles so feel a bit hesitant to change this again.

apis/nucleus/src/index.js
johanlahti: I think it could be done like this (at least that's how we do it in another repo) by adding a shorthand variable, e.g. `c`:
```
@typedef {object} c SwitchButtonConfig
@param {string} c.option
```
Good to know that command as well.

apis/nucleus/src/index.js
```diff
@@ -344,6 +344,44 @@ function nuked(configuration = {}) {
          * @typedef {function(function)} ReceiverFunction A callback function which receives another function as input.
          */
 
+        /**
+         * @typedef {object} c SwitchButtonConfig
```
streamside, line 348: For now `@interface` is needed to be used for get the typescript definitions generated properly in the Nebula.js repo.

```suggestion
         * @interface SwitchButtonConfig
         * @property {string} option The option to be toggled. Currently supporting: 'checkboxes', 'search', 'toolbar', 'rangeSelect', 'dense'.
         * @property {boolean=} [invert=false] Invert the switch button's value when applied on the option.
         * @property {boolean=} [startOn=false] Start the switch turned on (note that this will not change the initial value of the option).
         * @property {string} [label] A label for the switch.
         * @property {string} [helperText] A helper text label for the switch.
         * @property {object} [iconOn] CSS styling for the switch's on state (applied on span::after).
         * @property {object} [iconOff] CSS styling for the switch's off state (applied on span::before).
         * @property {function} [onChange] A function for defining a custom callback or for setting multiple properties using the provided setOptions function.
```

streamside (changes requested):

apis/nucleus/src/components/listbox/ListBoxRowColumn.jsx
```diff
@@ -142,15 +142,15 @@ export default function RowColumn({ index, style, data, column = false }) {
         [c] = page.qMatrix[index - area.qTop];
       }
     }
+    const selected = !!c && (c.qState === 'S' || c.qState === 'XS' || c.qState === 'L');
```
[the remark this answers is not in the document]
johanlahti: These are copy and pastes from old code but makes sense 😄

apis/nucleus/src/components/listbox/ListBoxRowColumn.jsx
```diff
@@ -147,7 +147,7 @@ export default function RowColumn({ index, style, data, column = false }) {
       }
     }
     setClassArr(clazzArr);
-  }, [cell && cell.qState]);
+  }, [cell && cell.qState, checkboxes]);
```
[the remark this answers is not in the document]
johanlahti: Otherwise we won't refresh the selection state when switching between checkboxes: true/false when using the switch. We will get checkbox fields with a green background e.g.

apis/nucleus/src/index.js
johanlahti: Ok, yes I noticed the discussion around this 👍

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -134,7 +134,7 @@ export default function ListBox({
 
   useEffect(() => {
     fetchData();
-  }, [layout]);
+  }, [layout, checkboxes]);
```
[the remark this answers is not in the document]
johanlahti: Same here, it is needed to enforce the new selection state on the fields, since checkbox fields should not have a green/gray background color, we need to trigger a refresh when switching between these two selection modes.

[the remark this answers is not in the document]
johanlahti: Had we triggered checkboxes: true/false from outside (using the Stardust API) we would also have run this effect. But since we are changing selection mode internally (from within the Listbox) we need to trigger this to happen by adding this flag to the effect watch array.

streamside (approved):

apis/nucleus/src/components/listbox/ListBoxRowColumn.jsx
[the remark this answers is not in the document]
T-Wizard: the checkbox option will need to be split (maybe not in this PR) as for analytic creation we want checkboxes and showing the state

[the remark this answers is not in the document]
johanlahti: Interesting, good to know 😄. Yes, not in the scope of this PR I think.

Caele:
The styling should follow sprout and the sprout theme. While Nebula doesn't use sprout directly (we have a copy), we will do that in the future when sprout becomes public.

PR #774, opened 2022-03-02
fix: update jsdocs for listbox

streamside:
## Motivation

Update JSDoc to improve API specification and typescript definitions. Changes:

* Add types for ListBox options containing fixed set of valid configs. For example, `FrequencyMode` and `Direction`.
* Use `@interface` to describe objects with a number of properties (instead of `@typedef`)


T-Wizard (approved):
I am OK with it but,
the change to @interface from @typedef should not be needed.
It is used a lot in picasso.js and if generating type from @typedef is broken we should look into it.

veinfors (approved):

PR #775, opened 2022-03-02
refactor(Listbox): Search and style improvements

johanlahti:
## Motivation

Changes:
- New option: ``` search: 'toggle' ``` which hides the search field initially and shows it when clicking the search button/icon
- New option: ``` focusSearch: {Boolean} ``` which when `true` initialises the Listbox with the search input focused (defaults to `false`). Only applicable when search is true since toggle -> show should always focus the search input, in accordance with best accessibility practices.
- Remove duplicate borders between search and toolbar
- Add tooltips to lock/unlock buttons
- When using search 'toggle' mode, the search toggle button is shown instead of the lock selections button *only when there are no selections*
- When search is activated (in 'toggle' mode) the search input gets focused

- ✅  Verify that search field styling works with the current usage of the Listbox @streamside 

<br />
<br />

*Initial state with search 'toggle':*
image

*Active state with search 'toggle':*
image


apis/nucleus/src/components/listbox/ListBoxSearch.jsx
```diff
@@ -8,9 +8,15 @@ import InstanceContext from '../../contexts/InstanceContext';
 
 const useStyles = makeStyles((theme) => ({
   root: {
+    border: `1px solid ${theme.palette.divider}`,
```
streamside, line 11: When looking at the styling it looks like the fieldset element is used as the level where the borders is applied. Can these styles applied on the root level be moved to the fieldset?

johanlahti: Yes, I could try moving these styles onto the `fieldset` instead. That may be more "correct" actually.

johanlahti: Indeed that looks better as we get the blue highlighting on focus 👍

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -5,9 +5,9 @@ import AutoSizer from 'react-virtualized-auto-sizer';
 import Lock from '@nebula.js/ui/icons/lock';
 import Unlock from '@nebula.js/ui/icons/unlock';
 
-import { IconButton, Grid, Typography } from '@material-ui/core';
-
-import { useTheme } from '@nebula.js/ui/theme';
+import { IconButton, Grid, Typography /* , makeStyles */ } from '@material-ui/core';
```
streamside, line 8:
```suggestion
import { IconButton, Grid, Typography } from '@material-ui/core';
```

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -171,22 +181,37 @@ export function ListBoxInline({ app, fieldIdentifier, stateName = '$', options =
 
   const minHeight = 49 + (search ? 40 : 0) + 49;
```
streamside, line 182: I suppose this statement needs to include logic about if the search is visible. Maybe extract logic if search is visible and reused here and around line 246 where the search box is rendered ?

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -218,9 +243,9 @@ export function ListBoxInline({ app, fieldIdentifier, stateName = '$', options =
           </Grid>
         </Grid>
       )}
-      {search ? (
+      {search === true || (search === 'toggle' && showSearch) ? (
```
streamside, line 246:
```suggestion
      {searchVisible && (
        <Grid item>
          <ListBoxSearch model={model} autoFocus={autoFocusSearch} dense={dense} />
        </Grid>
      )}
```

streamside (changes requested):

apis/nucleus/src/components/listbox/ListBoxSearch.jsx
streamside: Good! I tested it with analytic creation and it looks good in there as well.

apis/nucleus/src/components/listbox/ListBoxInline.jsx
johanlahti: Good input, I have changed.

johanlahti:
@streamside I have updated in accordance with your suggestions (great input!).
Also, I have added a new option `focusSearch`. This is required from sense-client's perspective for focusing the search input from start.

streamside (approved):

PR #776, opened 2022-03-03
feat: get the value of style with the exact given attributes seprated by dots 

LiKang6688:
…

## Motivation
Like in https://github.com/qlik-oss/sn-table/pull/453, sometimes we want to get a style value by exact search without a fallback.

When attributes separated by dots is provided, such as 'hover.color', they are required in the theme JSON file
Ex.
`const hoverFontColorFromTheme = theme.getStyle('object', 'straightTable.content', 'hover.color');`

that would require a hover: { color: ... } } in the theme but still fallback to look for that on multiple levels, supporting a theme with a global hover.color fallback.

and if you want more exact you can just call

`const hoverFontColorFromTheme = theme.getStyle('object', '', 'straightTable.content.hover.color');`

### DIFF
--- a/apis/theme/src/index.js
+++ b/apis/theme/src/index.js
@@ -73,29 +73,31 @@ export default function theme() {
     },
 
     /**
-     * Get the value of a style attribute in the theme by searching in the theme's JSON structure.
-     * The search starts at the specified base path and continues upwards until the value is found.
+     * Get the value of a style attribute in the theme
+     * by searching in the theme's JSON structure.
+     * The search starts at the specified base path
+     * and continues upwards until the value is found.
      * If possible it will get the attribute's value using the given path.
-     * When a default value is provided and there is no value of
-     * a style attribute found with that exactly specified path,
-     * the default value will be returned.
+     * When attributes separated by dots are provided, such as 'hover.color',
+     * they are required in the theme JSON file
      *
      * @param {string} basePath - Base path in the theme's JSON structure to start the search in (specified as a name path separated by dots).
      * @param {string} path - Expected path for the attribute (specified as a name path separated by dots).
-     * @param {string} attribute - Name of the style attribute.
-     * @param {string|null} [defaultValue] - Set a default value if no style value found.
-     * @returns {string|undefined|null} The style value or the default value or undefined
+     * @param {string} attribute - Name of the style attribute. (specified as a name attribute separated by dots).
+     * @returns {string|undefined} The style value or undefined if not found
      *
      * @example
      * theme.getStyle('object', 'title.main', 'fontSize');
+     * theme.getStyle('object', 'title', 'main.fontSize');
+     * theme.getStyle('object', '', 'title.main.fontSize');
      * theme.getStyle('', '', 'fontSize');
      */
-    getStyle(basePath, path, attribute, defaultValue) {
+    getStyle(basePath, path, attribute) {
       if (!styleResolverInstanceCache[basePath]) {
         styleResolverInstanceCache[basePath] = styleResolverFn(basePath, resolvedThemeJSON);
       }
 
-      return styleResolverInstanceCache[basePath].getStyle(path, attribute, defaultValue);
+      return styleResolverInstanceCache[basePath].getStyle(path, attribute);
     },
   };
 
--- a/apis/theme/src/index.js
+++ b/apis/theme/src/index.js
@@ -76,21 +76,26 @@ export default function theme() {
      * Get the value of a style attribute in the theme by searching in the theme's JSON structure.
      * The search starts at the specified base path and continues upwards until the value is found.
      * If possible it will get the attribute's value using the given path.
+     * When a default value is provided and there is no value of
+     * a style attribute found with that exactly specified path,
+     * the default value will be returned.
      *
      * @param {string} basePath - Base path in the theme's JSON structure to start the search in (specified as a name path separated by dots).
      * @param {string} path - Expected path for the attribute (specified as a name path separated by dots).
      * @param {string} attribute - Name of the style attribute.
-     * @returns {string} The style value
+     * @param {string|null} [defaultValue] - Set a default value if no style value found.
+     * @returns {string|undefined|null} The style value or the default value or undefined
      *
      * @example
      * theme.getStyle('object', 'title.main', 'fontSize');
      * theme.getStyle('', '', 'fontSize');
      */
-    getStyle(basePath, path, attribute) {
+    getStyle(basePath, path, attribute, defaultValue) {
       if (!styleResolverInstanceCache[basePath]) {
         styleResolverInstanceCache[basePath] = styleResolverFn(basePath, resolvedThemeJSON);
       }
-      return styleResolverInstanceCache[basePath].getStyle(path, attribute, false);
+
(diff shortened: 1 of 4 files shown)


apis/theme/src/index.js
```diff
@@ -76,21 +76,26 @@ export default function theme() {
...
      * If possible it will get the attribute's value using the given path.
+     * When a default value is provided and there is no value of
+     * a style attribute found with that exactly specified path,
+     * the default value will be returned.
      *
      * @param {string} basePath - Base path in the theme's JSON structure to start the search in (specified as a name path separated by dots).
      * @param {string} path - Expected path for the attribute (specified as a name path separated by dots).
      * @param {string} attribute - Name of the style attribute.
-     * @returns {string} The style value
+     * @param {string|null} [defaultValue] - Set a default value if no style value found.
```
haxxmaxx, line 86: Do we know the type of this? If the value you are looking for is a number, the defaultValue could be a number

apis/theme/src/style-resolver.js
```diff
@@ -94,18 +106,22 @@ export default function styleResolver(basePath, themeJSON) {
      * object.barChart - fontSize
      * object - fontSize
      * fontSize
+     * When defaultValue is provided,
+     * it get the value of a style attribute with the given base path + path
+     * Ex: object.barChart - legend.title - fontSize
```
haxxmaxx, line 111: This doesn't really make sense, something like this instead:
```suggestion
     * it will fallback to defaultValue instead of searching through the tree
```

apis/theme/src/style-resolver.js
```diff
@@ -94,18 +106,22 @@ export default function styleResolver(basePath, themeJSON) {
...
      * @ignore
      *
-     * @param {string} component string of properties seperated by dots to search in
-     * @param {string} attribute to return
-     * @returns {any} value of the resolved path, undefined if not found
+     * @param {string} component String of properties seperated by dots to search in
+     * @param {string} attribute Name of the style attribute
+     * @param {string|null} [defaultValue] Set a default value if no style value found.
+     * @returns {string|undefined|null} The style value of the resolved path, undefined if not found or the default value
```
haxxmaxx, line 117: Same here, can't know what it will return

apis/theme/src/index.js
LiKang6688: It can be 'Opinionated'. I do not think it is good to allow all types and return any types. so here I personally decided that it only accepts string or null.

apis/theme/src/style-resolver.js
LiKang6688: it gets the value of a style attribute with the given base path + pat, if not found, it will fallback to defaultValue

apis/theme/src/index.js
haxxmaxx: If there are properties that are of other types, we should allow both defaultValue and return to be of those types. If it is only string | null than we can keep this

apis/theme/src/style-resolver.js
haxxmaxx: yupp that is better

apis/theme/src/index.js
LiKang6688: Yes, lets keep this

haxxmaxx (approved):
LGTM, but I think another reviewer should have look as well

Caele:
I think I'd rather have this be an option than a default value, so:
`getStyle(basePath, path, attribute, exactMatch)` which will only return the value if the exact path exists, else undefined.
But we should also strive to never actually use it, rather make sure there are some kind of lingo to describe hover-colors. I know we've had issue with it in the past regarding themes, might be time to introduce something proper, possible a new attribute.
- backgroundColor - background
- color - text
- complimentColor - reverse text - used for hovers or reversed text stuff - not sure about the name,

haxxmaxx:
Great input, nice to have you back @Caele.
A new hover-color property would be awesome, I thought we were stuck with the current api.

LiKang6688:
> I think I'd rather have this be an option than a default value, so: `getStyle(basePath, path, attribute, exactMatch)` which will only return the value if the exact path exists, else undefined. But we should also strive to never actually use it, rather make sure there are some kind of lingo to describe hover-colors. I know we've had issue with it in the past regarding themes, might be time to introduce something proper, possible a new attribute.
> 
> * backgroundColor - background
> * color - text
> * complimentColor - reverse text - used for hovers or reversed text stuff - not sure about the name,

The idea of adding new attributes is good. Let me think about the name.

T-Wizard:
is exactMatch really what we want?
another option could be to support for something like:
```
const hoverFontColorFromTheme = theme.getStyle('object', 'straightTable.content', 'hover.color');
```
that would require a hover: { color: ... } } in the theme but still fallback to look for that on multiple levels, supporing a theme with a global hover.color fallback.

and if you want more exact you can just call
```
const hoverFontColorFromTheme = theme.getStyle('object', '', 'straightTable.content.hover.color');
```

LiKang6688:
> is exactMatch really what we want? another option could be to support for something like:
> 
> ```
> const hoverFontColorFromTheme = theme.getStyle('object', 'straightTable.content', 'hover.color');
> ```
> 
> that would require a hover: { color: ... } } in the theme but still fallback to look for that on multiple levels, supporing a theme with a global hover.color fallback.
> 
> and if you want more exact you can just call
> 
> ```
> const hoverFontColorFromTheme = theme.getStyle('object', '', 'straightTable.content.hover.color');
> ```

Thanks, it did not work for me before. But I could be wrong and let me try it again.

Caele:
> is exactMatch really what we want? another option could be to support for something like:
> 
> ```
> const hoverFontColorFromTheme = theme.getStyle('object', 'straightTable.content', 'hover.color');
> ```
> 
> that would require a hover: { color: ... } } in the theme but still fallback to look for that on multiple levels, supporing a theme with a global hover.color fallback.
> 
> and if you want more exact you can just call
> 
> ```
> const hoverFontColorFromTheme = theme.getStyle('object', '', 'straightTable.content.hover.color');
> ```

I like this, allows for extra levels at attributes.

apis/theme/src/style-resolver.js
```diff
@@ -56,12 +56,19 @@ function getObject(root, steps) {
...
     const target = getObject(theme, pathArray[i]);
-    if (target !== null && target[attribute]) {
-      return target[attribute];
+    if (target !== null) {
+      if (attributeArray.length === 1 && target[attribute]) {
+        return target[attribute];
+      }
+      if (attributeArray.length > 1) {
+        return getObject(target, attributeArray);
+      }
```
T-Wizard, line 68: why not just
```suggestion
      return getObject(target, attributeArray);
```
(the result of getObject needs to be changes from null to undefined (when not found) to match the jsdoc)

alternative could be to change
```
const target = getObject(theme, pathArray[i]);
```
to
```
return getObject(theme, [...pathArray[i], ...attributeArray]);
```

T-Wizard (commented):
I think it looks OK (but I have not tested it)
Would be nice to also add unit test for it

apis/theme/src/style-resolver.js
LiKang6688: Awesome! I did an update and added a unit test for it.

T-Wizard (approved):

PR #778, opened 2022-03-08
refactor(Listbox): Keyboard navigation

johanlahti:
## Motivation

Enable keyboard navigation, permitting the following features:
- **Space**: Select values with space
- **Arrow buttons**: Move to previous/next field
- **Arrow buttons + Shift**: Select a range
- **Enter**: Confirm selections
- **Escape**: Remove focus from `document.activeElement` and cancel selections

Already supported:
- **Tab**: Supported "natively" using tabIndex where needed
- Focus search field on toggle


Caele:
LGTM, make sure to update from master as I updated the linting rules.

johanlahti:
I have merged in latest master and fixed some breaking code & unit tests in response to this.

johanlahti:
I have merged in latest master and fixed some breaking code & unit tests in response to this.

apis/nucleus/src/components/listbox/ListBoxRowColumn.jsx
```diff
@@ -153,10 +157,13 @@ export default function RowColumn({ index, style, data, column = false }) {
     checkboxes = false,
     dense = false,
     frequencyMode = 'N',
+    actions,
     frequencyMax = '',
     histogram = false,
   } = data;
 
+  const handleKeyDownCallback = useCallback(getKeyboardNavigation(actions), [actions]);
```
T-Wizard, line 165: I think the useCallback here will have very limited effect as the the parent component send in a new actions object every times it re-render. (It still could help for the case when only this and not the parent component is updated)

apis/nucleus/src/components/listbox/listbox-keyboard-navigation.js
```diff
@@ -0,0 +1,63 @@
...
+        }
+        break;
+      case KEYS.ENTER:
+        confirm();
+        break;
+      case KEYS.ESCAPE:
+        cancel();
+        if (document.activeElement) {
+          document.activeElement.blur();
+        }
```
T-Wizard, line 53: In most real use cases I expect that you want to move the focus somewhere else (outside the listbox) and not just remove it. (but the listbox itself will not have enought information)

T-Wizard (approved):
With this pressing enter in a popover listbox confirm the selection but do not close the listbox.
I think that is a bit confusion and I woud expect it to close the popover.

apis/nucleus/src/components/listbox/ListBoxRowColumn.jsx
johanlahti: Yeah agreed. I wrap this mostly as a best practice/habit than intentional optimisation.

Caele:
@johanlahti I now noticed you are not using the accessibility settings at all? Not sure how it works in the listbox but you need to make use of the keyboard hook I think. 
It has a lot of functions to be used for recieving focus and bluring, currently only used in the table.

Caele:
> @Caele I tried using it in sense-client but since this is not a Supernova I can't call hooks from there. (I'm using the `field` method to create the Listbox). Is the `useKeyboard` hook intended for use from within Nebula as well? I don't see any usage yet inside Nebula.js, maybe because it's new…

It was develop for usage in the table, so no other places yet. You should be able to get hold of the object it uses to track things at least, but I'd have to test a bit to know the details.

What is important is that the listbox follows the same principles as other supernovas do, when turning accessibility on/off and how blur/focus is handled.

apis/nucleus/src/components/listbox/listbox-keyboard-navigation.js
johanlahti: Yeah, maybe this is where the `useKeyboard` can help. I think @Caele might have a look next week and see how we can use the keyboard. I am far from 100% sure about the accessibility best practices here so may be more holes to fill. We can improve iteratively.

PR #795, opened 2022-03-16
chore: add long running test

Caele:
## Motivation

Adds a test for long runnning queries by adding a delay to the mocked engine calls. It currently doesn't properly cancel the call when you press cancel (or the test presses cancel), but it covers the functionality.

@streamside Feel free to clean up this code if you like, had to do some light hacking to get the delay to work without slowing everything down too much.

### DIFF
--- a/apis/enigma-mocker/src/__tests__/prop.spec.js
+++ b/apis/enigma-mocker/src/__tests__/prop.spec.js
@@ -67,15 +67,38 @@ describe('getPropFn', () => {
   });
 
   describe('async is enabled', () => {
+    let clock;
+
+    before(() => {
+      clock = sinon.useFakeTimers();
+    });
+
+    afterEach(() => {
+      clock.restore();
+    });
+
     it('returns promise', async () => {
       const prop = sinon.stub();
       prop.returns(500);
       const fn = getPropFn(prop, { async: true });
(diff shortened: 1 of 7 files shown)


streamside:
The intent when writing this was to include all properties on the `genericObjects` passed to the enigma mocker. The property `delay` is more of a setting on how it should behave. Down the road there might be conflicts if engine also provides a property with the same name. I suggest to introduce an options object - which for now includes only the delay setting - to the exposed API:

```js
fromGenericObjects(genericObjects, options)
```
The delay option would be "global" and effect all generic objects being mocked. If more granular control is desired the dev would be able to delay individual objects / functions.

The code looking up the generic object by its qId didn't become that pretty with the `findAsync()`. Another option would be to return the qId along with the mock when creating the mocks (around line 32 in `get-object-mock.js`). The lookup to find the correct object would become easier and no need to add qId on the mock.

Let me know your thoughts about it and if I should make any changes to the code.

streamside:
@Caele I have updated the PR with two changes. (1) `delay` added in options object the enigma-mocker which effects all async calls, and (2) lookup of the mocks via `qId` is simplified. Feel free to take a look!

streamside (approved):

PR #799, opened 2022-03-18
Fix(Listbox) optimistic update when single select

DanielS-Qlik:
Optimistic unselect all values except newly selected when in single select mode (qOneAndOnlyOne).
Add radio buttons for single select.

Before
before
After
after
After non dense
non_dense


apis/nucleus/src/components/listbox/ListBoxRadioButton.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+  return (
+    <Radio
+      checked={checked}
+      value={label}
+      name={label}
+      inputProps={{ 'aria-labelledby': label }}
+      className={dense ? styles.denseRadioButton : styles.radioButton}
+      style={{ backgroundColor: 'transparent' }}
+      disableRipple
```
johanlahti, line 30: This could also be set as a global property override in `theme.js` but I think is fine too.

johanlahti:
Looks good!
Check other optimisation options for `memo` (using `qElementNumber` instead of `JSON.stringify()`) if it works.

johanlahti (approved):

DanielS-Qlik:
> Looks good! Check other optimisation options for `memo` (using `qElementNumber` instead of `JSON.stringify()`) if it works.

Thanks for input :) I will look for optimizations in a new PR.

PR #815, opened 2022-04-06
refactor(listbox): add showGray option to disable gray colors

johanlahti:
## Motivation

Allow not showing gray colors (Excluded, Alternative) through a new option: `showGray {Boolean}`.


T-Wizard (approved):
I think the specific changes are OK.
But as we add more and more to the listbox. I think we are making a mistake by not making it a supernova object like all other charts.

PR #824, opened 2022-05-11
chore: change id generation

Caele:
## Motivation

Using Date.now() as a identifier led to multiple objects having the same ID, so added a more robust ID generator.

### DIFF
--- a/apis/enigma-mocker/src/mocks/create-session-object-mock.js
+++ b/apis/enigma-mocker/src/mocks/create-session-object-mock.js
@@ -1,12 +1,12 @@
 /* eslint-disable no-underscore-dangle */
 import extend from 'extend';
 
+// To cover test
 const crt = global.crypto || { getRandomValues: () => 123456 };
 
 // https://stackoverflow.com/questions/105034/how-to-create-a-guid-uuid
 // Not using crypto.randomUUID due to missing safari support < 15
 function uuidv4() {
-  // To cover test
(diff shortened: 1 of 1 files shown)


apis/enigma-mocker/src/mocks/create-session-object-mock.js
```diff
@@ -1,10 +1,26 @@
...
+    (c ^ (crt.getRandomValues(new Uint8Array(1))[0] & (15 >> (c / 4)))).toString(16)
+  );
+}
+
 function CreateSessionObjectMock() {
   return (props) => {
-    const mockedInclusions = props ? props._mock : null;
+    props = props || {}; // eslint-disable-line no-param-reassign
+    props.qInfo = props.qInfo || {}; // eslint-disable-line no-param-reassign
+    props.qInfo.qId = props.qInfo.qId || `mock-${uuidv4()}`; // eslint-disable-line no-param-reassign
```
T-Wizard, line 19: this now changes the passed in props (the old code did not).
that could maybe cause problem in some edge cases. (I expect it to be OK most of the time)

T-Wizard (approved):

apis/enigma-mocker/src/mocks/create-session-object-mock.js
Caele: hmm yea, maybe better to explicitly copy them first. Depends a bit on the use case, think it could be a valid scenario to pass in props, have them update to reflect the charts defaults, and then alter the object I sent in a bit.

PR #826, opened 2022-05-18
feat: confirm selection when interacting outside of listbox

DanielS-Qlik:
## Motivation

End/-confirm selection when user interacts outside of listbox.
More info: https://jira.qlikdev.com/browse/QLIK-104239?filter=34689&jql=parent%20%3D%20QLIK-104108%20AND%20status%20!%3D%20Closed%20ORDER%20BY%20priority%20DESC

Before, inconsistent behavior
before2

After
after2


johanlahti (approved):

apis/nucleus/src/components/listbox/useConfirmUnfocus.jsx
```diff
@@ -0,0 +1,39 @@
+import { useRef, useEffect, useState } from 'react';
+
+/**
+ * Hook that confirms selection when
+ * user interact outside of passed element ref.
+ *
+ * @param ref - Ref to HTMLelement
+ * @param selections
+ */
```
Caele, line 9: This would get added ot the public API, I don't think we want that?

DanielS-Qlik: That would be unnecessary, will remove. Thanks for input!

Caele:
@DanielS-Qlik Another questions, did you test how this change behaves in a mashup?

DanielS-Qlik:
@Caele, yes you get the same behaviour.
end-selection

PR #832, opened 2022-05-24
fix(ac): looks

DanielS-Qlik:
## Motivation

Make the rows even denser when dense in enabled. Remove row divider when histogram and dense is enabled.

listbox-small-row


johanlahti (approved):

Caele:
@johanlahti @DanielS-Qlik FYI: be careful with this kind of change. While I think this one is probably fine, we do have active users that will get impacted by any visual change.

PR #833, opened 2022-05-24
feat(ListBox): Add function for filtering values

johanlahti:
## Motivation

We have a requirement to filter out certain value states (based on the `qState`). This PR adds a new function (not exposed to users) for achieving this in a generic way, called: `filterValues`. It receives `pages` and must return `pages`.
The `applyFilter` function then adapts the `qHeight` to reflect the decreased number of values and applies this on the List element, to prevent ghost values (without text).

Other: Increased scroll timeout as it reduces the amount of calls executed while scrolling, without really affecting performance.


apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -144,15 +158,19 @@ export default function ListBox({
                 }))
               )
               .then((p) => {
+                let losPaginos = p;
+                if (filterValues && p && p.length) {
+                  losPaginos = applyValueFilter(p, filterValues);
```
DanielS-Qlik, line 163: Is it possible to modify `p` directly instead of creating `losPaginos` ?

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -112,8 +113,21 @@ export default function ListBox({
     [layout, pages]
   );
 
+  const applyValueFilter = (logPaginos, filterFunc) => {
```
DanielS-Qlik, line 116: Rename `logPaginos` to receivedPages?

DanielS-Qlik (approved):

apis/nucleus/src/components/listbox/ListBox.jsx
johanlahti: Possible, but the lint would complain and immutability is usually preferred.

apis/nucleus/src/components/listbox/ListBox.jsx
johanlahti: Yeah, and the correct spelling should also be `losPaginos` 😄 .
My variable naming clearly needs some help here. Thanks!

johanlahti:
I've tried summing up the height over multiple pages but this does not work when performing selections. Will not spend more time on this since we will anyways need a "real" solution, which must be implemented in the backend.

PR #846, opened 2022-06-02
fix: remove light-gray checkboxes in alternative state

DanielS-Qlik:
## Motivation

Before:
cb-white-before

After:
cb-white-after


johanlahti (approved):

Caele:
Can this wait? I'm getting the MUI v5 update in soon and it will break all of this.
If this has to go in, we'll make a 2.x branch for it, but please wait a bit.

Caele (dismissed):
Wait!

T-Wizard:
Do we want this inconsistency of coloring between normal listbox and listbox with checkboxes in nebula.js?
where normal listbox has a specific (light gray) color for alternatives and listbox with checkboxes don't.

(could the specific use case be solved by using a theme where the selected.alternative is changed to white?)

PR #847, opened 2022-06-07
feat(ListBox): postProcessPages and calculatePagesHeight options

johanlahti:
## Motivation

Adaptations to DQ requirements:
- Re-calculate `qHeight` client-side to adjust for filtered values (enabled with option `calculatePagesHeight {Boolean}`)
- Limit number of items to render by telling the `FixedSizeList` component how many rows to render

The latter limit is applied only when there is one page and there are filtered values in that page. This is derived from:
`hasBeenFiltered = number_of_values_in_page < 100` since there are maximum 100 values (i.e. rows) in a page.

When there are 100 values in a page, no filtering has been made.

## Before
image

## After (no empty rows)
image

## S + X should not filter values but if no selections exist in field, then X should be filtered.

https://user-images.githubusercontent.com/5780544/173596803-e1a80419-cfd2-483d-bac9-66082d493864.mov


T-Wizard:
> ## Motivation
> Adaptations to DQ requirements, where we need to adjust `qHeight` client-side of each returned page, since it is incorrect after the backend calculation. We then have to explicitly tell the Listbox to use the accumulated sum of page heights (corrected using client-side override with `calculatePagesHeight`) to decide how many items to show in the List.

this is very questionable. 
the accumulated sum of page heights is not represent the same value the count.
the "accumulated sum of page heights" represent the number of values fetched
and "count" is the number of values possible to fetch.
and this change should break data paging.

Caele:
Its starting to get very messy with all these DQ adjustments, could we do an overview of all of them? After which we might want to just break it out into its own implementation instead of polluting the standard ListBox?

johanlahti:
Indeed, and the only purpose is to cover up for limitations which should be fixed in Engine (`qHeight` not reflecting number of rows per page and `qSize.qcy` not being reliable). Very sceptical about this over all.

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -218,10 +218,18 @@ export default function ListBox({
...
-  const getAccumulatedHeight = (ps) => ps.reduce((acc, p) => acc + p.qArea.qHeight, 0);
-
-  const listCount = pages && pages.length && calculatePagesHeight ? getAccumulatedHeight(pages) : count;
+  const PAGE_MAX_HEIGHT = 100;
+
+  const getCalculatedHeight = (ps) => {
+    // If values have been filtered in the currently loaded page, we want to
+    // prevent rendering empty rows by assigning the actual number of items to render
+    // since count (qcy) does not reflect this in DQ mode currently.
+    const h = ps[0].qArea.qHeight;
```
T-Wizard, line 227: why are you checking only ps[0] and not ps[1] (when it exists)

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -218,10 +218,18 @@ export default function ListBox({
   const isVertical = listLayout !== 'horizontal';
   const count = layout.qListObject.qSize.qcy;
 
-  const getAccumulatedHeight = (ps) => ps.reduce((acc, p) => acc + p.qArea.qHeight, 0);
-
-  const listCount = pages && pages.length && calculatePagesHeight ? getAccumulatedHeight(pages) : count;
+  const PAGE_MAX_HEIGHT = 100;
```
T-Wizard, line 221: I think this is from `minimumBatchSize={100}`.
consider also useing this constant as arguemnt to minimumBatchSize.
(as it is a minimun it could maybe be wring for a extremely large listbox. with more then 100 values visible at the same time)

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -214,6 +217,19 @@ export default function ListBox({
 
   const isVertical = listLayout !== 'horizontal';
   const count = layout.qListObject.qSize.qcy;
+
+  const PAGE_MAX_HEIGHT = 100;
+
+  const getCalculatedHeight = (ps) => {
```
T-Wizard, line 223: have you verified this on a use case that needs to page data and you scroll to where the data is filtered?
(My understanding is that `itemCount` should be the total number of items in the list not the number of items fetched the last time.)

apis/nucleus/src/components/listbox/ListBox.jsx
johanlahti: Yes, we should probably apply this to unlimited pages. My original guess/assumption was that only the first page (index == 0) will be affected when filtering applies. But we could of course have more than 100 (or `max_items_per_page`) of filtered values, which suggests more than one page is affected. Also, this should then affect the last page rather than the first.

Let me push a change and hear what you think about it.

apis/nucleus/src/components/listbox/ListBox.jsx
johanlahti: I have tried to get such a test case but for now we have to rely on intuition of how it works. What you say sounds reasonable. I just wonder what it would look like and if it's something we can live with for the first release – or if we could accumulate the height. Is `page.qArea.qTop` reliable for this?

apis/nucleus/src/components/listbox/ListBox.jsx
johanlahti: Yes, seems so. I will use that one instead.

apis/nucleus/src/components/listbox/ListBox.jsx
T-Wizard: My current understaning (that I am unsure about) is that you can have one or two pages.
For two pages ps[1] is the last requeted page and ps[0] the previous requested page (requested by react-window-infinite-loader).
That the requested pages can be overlapping or have space beween them.
And that any of the pages could start before the other.

apis/nucleus/src/components/listbox/ListBox.jsx
T-Wizard: I would expect `page.qArea.qTop` and `page.qArea.qHeight` to be reliable.
but I rember hearing that engine did not have a reliable `page.qArea.qHeight` (in some versions? it is used in this PR)
so it may be best to check with the one who implemented the backend fix, on what the client can use.
(I have not had any direct communiction on what engine support in direct query apps)

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -214,6 +219,17 @@ export default function ListBox({
...
   const count = layout.qListObject.qSize.qcy;
+
+  const getCalculatedHeight = (ps) => {
+    // If values have been filtered in the currently loaded page, we want to
+    // prevent rendering empty rows by assigning the actual number of items to render
+    // since count (qcy) does not reflect this in DQ mode currently.
+    const hasFilteredValues = ps.some((page) => page.qArea.qHeight < MINIMUM_BATCH_SIZE);
+    const h = ps.reduce((tot, page) => tot + page.qArea.qHeight, 0);
+    return hasFilteredValues ? h : count;
+  };
```
johanlahti, line 230: Can we use `qTop` to account for rows prior to the ones just loaded:
```
const nbrOfPriorRows = pages[0].qArea.qTop;
const h = nbrOfPriorRows + ps.reduce((tot, page) => tot + page.qArea.qHeight, 0);
```
?

johanlahti: 🤔 💭

T-Wizard: Try `const h = Math.max(...ps.map((page) => page.qArea.qTop + page.qArea.qHeight));`

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -224,7 +240,7 @@ export default function ListBox({
       itemCount={count}
```
T-Wizard, line 240: I think this also should be using listCount
```suggestion
      itemCount={listCount}
```

johanlahti: Yes it should.

apis/nucleus/src/components/listbox/ListBox.jsx
johanlahti: They also believe it should be reliable in DQ mode so let's proceed with this.

T-Wizard (approved):

apis/nucleus/src/components/listbox/ListBox.jsx
johanlahti: Unfortunately this does not work in the DQ scenario I've tested. When we perform selections in a field which does not have selections from before – previously excluded values (`qState == 'X'`) are not filtered anymore. In other words: X + S (in same field) => X-values are not filtered.

When we clear the selections of the field the Loader does not ask for more data than it currently has knowledge of (which is based on current pages). Maybe it's possible to send an additional request to find this out but I we should keep `qcy` for now.

T-Wizard: For that case should not 'getCalculatedHeight' detect that there is no filtering anymore and return count that is `qcy`?

PR #848, opened 2022-06-07
feat: data count to hosting app

DanielS-Qlik:
## Motivation
Write the length/count of the list as a parameter to a passed function.


apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -181,6 +182,9 @@ export default function ListBox({
 
   useEffect(() => {
     fetchData();
+    if (setCount !== undefined && layout) {
```
johanlahti, line 185:
```suggestion
    if (typeof setCount === 'function' && layout) {
```

johanlahti (approved):

T-Wizard (approved):

PR #855, opened 2022-06-20
fix: add italic to checkbox text when excluded

DanielS-Qlik:
## Motivation

Before:
italic-before

After:
italic-after


Caele:
Is there a specific reason to make the text italic?

DanielS-Qlik:
It's a decision from UX :) I believe it's to make it more clear what data is not being used in the selection.

Caele (approved):

Issue #860, opened 2022-06-23
Add new property to `ListBoxInline` to enable `useConfirmUnfocus` and set disabled it as default


haxxmaxx:
## 🚀 Feature

Add new property to `ListBoxInline` to enable `useConfirmUnfocus` and set disabled it as default.

### Motivation

In most cases the click outside is handled in the context it is used. `useConfirmUnfocus` should be used in special cases

PR #861, opened 2022-06-23
fix(listbox): only enable confirming selection on blur when enable

haxxmaxx:
## Motivation

fixes https://github.com/qlik-oss/nebula.js/issues/860

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`
    - [x] No changes
     ***OR***

### DIFF
--- a/apis/nucleus/src/components/listbox/useConfirmUnfocus.jsx
+++ b/apis/nucleus/src/components/listbox/useConfirmUnfocus.jsx
@@ -5,7 +5,7 @@ import { useRef, useEffect, useState } from 'react';
  * user interact outside of passed element ref.
  */
 
-export default function useConfirmUnfocus(ref, selections, enabledConfirmOnBlur) {
+export default function useConfirmUnfocus(ref, selections, shouldConfirmOnBlur) {
   const [_isConfirmed, _setIsConfirmed] = useState(false);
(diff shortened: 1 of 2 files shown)


apis/nucleus/src/components/listbox/useConfirmUnfocus.jsx
```diff
@@ -15,6 +15,8 @@ export default function useConfirmUnfocus(ref, selections) {
   };
 
   useEffect(() => {
+    if (!shouldConfirmOnBlur) return () => {};
```
T-Wizard, line 18: there should be no need to return an empty function here
I would go with just `return;` (or `return undefined;` if a return value are needed by lint rules)

T-Wizard (approved):

apis/nucleus/src/components/listbox/useConfirmUnfocus.jsx
haxxmaxx: ok I'll try to remove it

DanielS-Qlik:
You might want to add it to the `DoNotUseOptions` here:  https://github.com/qlik-oss/nebula.js/blob/f44fc1a173621da9ba05ef5b6913407e793f3cf5/apis/nucleus/src/index.js#L140

Issue #863, opened 2022-06-23
Refactor(listbox): Move handling confirm selection on blur out of nebula


haxxmaxx:
## 🚀 Feature

Move handling confirm selection on blur out of nebula. This should not be a part of the listbox itself, but rather the context it is used in, since we can't now what selection toolbar it will use

### Motivation

Keep the listbox generic.

PR #866, opened 2022-06-27
fix(listbox): prop to disable confirming selections on blur events

haxxmaxx:
## Motivation

Temporary fix to disable confirming selections on blur events. This should eventually be refactored, see issue #863

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`
    - [x] No changes
     ***OR***

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -59,7 +59,7 @@ export default function ListBoxInline({ app, fieldIdentifier, stateName = '$', o
     sortByState = 1,
     scrollState = undefined,
     setCount = undefined,
-    shouldConfirmOnBlur,
+    shouldConfirmOnBlur = undefined,
   } = options;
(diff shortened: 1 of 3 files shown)


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -59,6 +59,7 @@ export default function ListBoxInline({ app, fieldIdentifier, stateName = '$', o
     sortByState = 1,
     scrollState = undefined,
     setCount = undefined,
+    shouldConfirmOnBlur = undefined,
```
veinfors, line 62: Maybe it's just my personal preference, but I think it feels nicer calling it just `confirmOnBlur`.

haxxmaxx: or actually I just merged the other PR so easier to stick with this

veinfors: that's fine, just a detail

PR #872, opened 2022-06-29
fix: filterBox react on qsize.qcy change when searching

mhs30:
## Motivation

Listbox fetch It’s only listening to changes on layout.. but since layout.qListObject.qSize.qcy is deep inside the object structure it’s not reacting on qSize.qcy changing when searching

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`
    - [ ] No changes
     ***OR***

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBox.jsx
+++ b/apis/nucleus/src/components/listbox/ListBox.jsx
@@ -196,7 +196,7 @@ export default function ListBox({
     if (typeof setCount === 'function' && layout) {
       setCount(layout.qListObject.qSize.qcy);
     }
-  }, [layout && layout.qListObject.qSize.qcy]);
+  }, [layout, layout && layout.qListObject.qSize.qcy]);
 
   useEffect(() => {
     if (!instantPages || isLoadingData) {
--- a/apis/nucleus/src/components/listbox/ListBox.jsx
+++ b/apis/nucleus/src/components/listbox/ListBox.jsx
@@ -196,7 +196,7 @@ export default function ListBox({
     if (typeof setCount === 'function' && layout) {
       setCount(layout.qListObject.qSize.qcy);
     }
-  }, [layout]);
+  }, [layout && layout.qListObject.qSize.qcy]);
 
   useEffect(() => {
     if (!instantPages || isLoadingData) {


apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -196,7 +196,7 @@ export default function ListBox({
     if (typeof setCount === 'function' && layout) {
       setCount(layout.qListObject.qSize.qcy);
     }
-  }, [layout]);
+  }, [layout && layout.qListObject.qSize.qcy]);
```
haxxmaxx, line 199: this is weird I would expect layout to be a new instance when something on there is updated. Anyway, I would put this as a const above to not have logic in the dep arrray

mhs30: i missed [layout, layout && layout.qListObject.qSize.qcy]

veinfors: @haxxmaxx When we get layout changes back from engine as deltas, I think we "patch" those changes on an existing layout object, that is the explanation to this (I think :)).

veinfors (approved):

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -196,7 +196,7 @@ export default function ListBox({
     if (typeof setCount === 'function' && layout) {
       setCount(layout.qListObject.qSize.qcy);
     }
-  }, [layout]);
+  }, [layout, layout && layout.qListObject.qSize.qcy]);
```
MTjody, line 199: Is it safe to chain this? Looks succeptible to "Cannot read properties of 'undefined'"

mhs30: I would say so since we are checking for layout, manually tested it and looks good.

veinfors: It seems to work fine, yes.
I would have preferred to write it using optional chaining, like: `layout?.qListObject...` but it's not allowed in nebula at the moment.

MTjody: The ugly long version of the above would be
`layout && layout.qListObject && layout.qListObject.qSize && layout.qListObject.qSize.qcy`

Up to you 🙈

veinfors: We are already assuming that `qListObject.qSize.qcy` exist once layout is in place (line 197) today, so this adds no extra risk the way I see it.

PR #873, opened 2022-06-29
fix: issue with empty search results in listbox

StefanEnberg:
## Motivation

Made this fix very explicit so that it is not hard to miss (+1 on line 222 would have looked slicker)

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`
    - [ ] No changes
     ***OR***

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBox.jsx
+++ b/apis/nucleus/src/components/listbox/ListBox.jsx
@@ -196,7 +196,7 @@ export default function ListBox({
     if (typeof setCount === 'function' && layout) {
       setCount(layout.qListObject.qSize.qcy);
     }
-  }, [layout]);
+  }, [layout, layout && layout.qListObject.qSize.qcy]);
 
   useEffect(() => {
     if (!instantPages || isLoadingData) {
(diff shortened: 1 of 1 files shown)


apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -196,7 +196,7 @@ export default function ListBox({
     if (typeof setCount === 'function' && layout) {
       setCount(layout.qListObject.qSize.qcy);
     }
-  }, [layout]);
+  }, [layout, layout && layout.qListObject.qSize.qcy]);
```
veinfors, line 199: This diff is confusing me a little... this is actually what it looks like on master.. maybe something is messed up with this branch or the PR. It's probably fine but a little confusing that we see this diff.

veinfors, line 199: This diff is confusing me a little... this is actually what it looks like on master.. maybe something is messed up with this branch or the PR. It's probably fine but a little confusing that we see this diff.

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -218,7 +218,14 @@ export default function ListBox({
   }
 
   const isVertical = listLayout !== 'horizontal';
-  const count = layout.qListObject.qSize.qcy;
+
+  let count = layout.qListObject.qSize.qcy;
+
+  // Should not allow a count of zero since it will stop executing loadMoreItems
+  // this could happen if a seach returns an empty result
+  if (count === 0) {
```
veinfors, line 226: I tested this and it seems to work as expected and solve the problem!
Would be nice with a unit test covering this though.

veinfors (approved):

apis/nucleus/src/components/listbox/ListBox.jsx
StefanEnberg: https://github.com/qlik-oss/nebula.js/pull/886

PR #875, opened 2022-06-30
fix: merge passed properties. remove sortByState option

DanielS-Qlik:
## Motivation

When using `properties` to override `listdef.qListObjectDef`, all members of `qListObjectDef` was overwritten by the passed `properties` object, even if you only wanted to change one property.
With this PR the passed `properties` is merged with `listdef`.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`
    - [x] No changes
     ***OR***


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -56,7 +57,6 @@ export default function ListBoxInline({ app, fieldIdentifier, stateName = '$', o
     postProcessPages = undefined,
     calculatePagesHeight,
     showGray = true,
-    sortByState = 1,
```
Caele, line 59: Still need to support options.sortByState so that we don't break the API.

DanielS-Qlik: I added options.sortByState to the `DoNotUseOptions` not too long ago (may 30, 2022). Do you still think it's necessary to keep it?

Caele: no, if it isn't official then it is fine to remove it.

Caele (approved):

PR #881, opened 2022-07-01
fix: enable enter key to apply selection in searches

StefanEnberg:
## Requirements checklist

- [ ] Api specification
  - [x] Ran `yarn spec`
    - [x] No changes
     ***OR***


apis/nucleus/src/components/listbox/ListBoxSearch.jsx
```diff
@@ -29,9 +29,8 @@ export default function ListBoxSearch({ model, keyboard, dense = false, visible
   const onKeyDown = (e) => {
     switch (e.key) {
       case 'Enter':
-        // This is disabled to "solve" QB-10955 with the lowest risk
-        // await model.acceptListObjectSearch(TREE_PATH, true);
-        // setValue('');
```
StefanEnberg, line 34: await should not be part of the original code that is commented it got committed by mistake

LiKang6688 (approved):

PR #882, opened 2022-07-11
feat: listbox disable search prop

mhs30:
## Summary

Added disableSearch prop to be able to control if we should build the search input or not

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -60,6 +60,7 @@ export default function ListBoxInline({ app, fieldIdentifier, stateName = '$', o
     scrollState = undefined,
     setCount = undefined,
     shouldConfirmOnBlur = undefined,
+    disableSearch = false,
   } = options;
   let { frequencyMode, histogram = false } = options;
 
@@ -318,9 +319,11 @@ export default function ListBoxInline({ app, fieldIdentifier, stateName = '$', o
           </Grid>
         </Grid>
       )}
-      <Grid item ref={searchContainerRef}>
-        <ListBoxSearch model={model} dense={dense} keyboard={keyboard} visible={searchVisible} />
-      </Grid>
+      {!disableSearch && (
+        <Grid item ref={searchContainerRef}>
+          <ListBoxSearch model={model} dense={dense} keyboard={keyboard} visible={searchVisible} />
+        </Grid>
+      )}
       <Grid item xs>
         <div ref={moreAlignTo} />
         <AutoSizer>


qlikossbuild:
Hi, and thanks for contributing!

Note: If you are part of Qlik R&D, you can apply to the `open-source` team in our _internal_ GitHub organization, and ignore the rest of these instructions.

Before we can handle your pull request we need you to sign our contributor license agreement.
This is a one-time process and enables you to contribute without hassle to any of the qlik-oss repositories in the future.

And it only takes a few minutes! Here are some useful links to get you started:

* [Sign the CLA](https://github.com/qlik-oss/open-source/blob/master/sign-cla.md#sign-the-cla)
* [Rationale behind the CLA](https://github.com/qlik-oss/open-source/blob/master/cla-rationale.md)
* [Qlik R&D Open Source repository](https://github.com/qlik-oss/open-source#open-source-at-qlik-rd)
* [Official Slack team](http://branch.qlik.com/slack)

Having troubles? Feel free to reach out to us either here or on Slack!

**Note:** Once the CLA is sorted, you should rebase/touch this pull request to trigger an update of this CLA status.

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -318,9 +319,11 @@ export default function ListBoxInline({ app, fieldIdentifier, stateName = '$', o
           </Grid>
         </Grid>
       )}
-      <Grid item ref={searchContainerRef}>
-        <ListBoxSearch model={model} dense={dense} keyboard={keyboard} visible={searchVisible} />
-      </Grid>
+      {!disableSearch && (
+        <Grid item ref={searchContainerRef}>
```
a-m-dev, line 323: This part looks ok, but i think maybe this could case issues on line https://github.com/mhs30/nebula.js/blob/34115f7c02e6ccba0699a3a4a731ffc64e414980/apis/nucleus/src/components/listbox/ListBoxInline.jsx#L223
while trying to access of `searchContainer.current`. i think it should throw a "trying to access the value of undefined" error if we fall in that case

a-m-dev: not actually, that if case is above the usage of `searchContainer.current`, i would rather say one more check for the value either in if case like: 
```tsx
   if ( searchContainer?.current && ((search && focusSearch) || (search === 'toggle' && showSearch))) {
       // ...
    }
```
or inside of it:
```tsx
    if ((search && focusSearch) || (search === 'toggle' && showSearch)) {
      const input = searchContainer?.current.querySelector('input');
      input && input.focus();
    }
```

I would go with 2nd solution since it already checks for `input` value first (incase of `searchContainer` being undefined -> `input` becomes undefined) and then focuses on.

MTjody (approved):
🚢

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -318,9 +319,11 @@ export default function ListBoxInline({ app, fieldIdentifier, stateName = '$', o
           </Grid>
         </Grid>
       )}
-      <Grid item ref={searchContainerRef}>
-        <ListBoxSearch model={model} dense={dense} keyboard={keyboard} visible={searchVisible} />
-      </Grid>
+      {!disableSearch && (
```
niekvanstaveren, line 322: Why adding another condition here? Would it not be possible to use the already existing searchVisible? Just add the disableSearch check there?

a-m-dev: if we forcefully pass `visible={false}` to `ListBoxSearch` component, we don't even need to modify `searchVisible` in order to get our desired results || adding another prop to the component. 
Since the `<ListBoxSearch />` already checks for:
```tsx
if (!visible) {
    return null;
  }
``` 
inside of itself, then i'm pretty sure that should do the work.

the only remaining thing is the search icon which comes from sense-client. it still re-renders the react part when clicking on it

Screenshot 2022-07-12 at 09 55 38

I think we will need another PR for it in sense client in order to hide it temporarily

MTjody: I think `disableSearch` and `searchVisible` refer to different things, that's why the prop was added and not re-used

MTjody: because the listbox has a toggle-able searchfield.

apis/nucleus/src/components/listbox/ListBoxInline.jsx
MTjody: You won't get there due to the dependency array of the useEffect

a-m-dev: ✅

MTjody (changes requested):
Like Niek pointed out, we can disable the search from Sense-client instead by passing the showSearch property as false. That will fix removing the clickable search icon as well. We should close this PR

PR #893, opened 2022-08-15
chore(dev-mashup): Fix broken yarn start

johanlahti:
## Motivation

With parcel v.2 we need to explicitly specify type="module" for `<script>` tags and cannot specify browser, module, main with an `.html` extension (parcel will throw an error).

Reference: https://parceljs.org/blog/rc0/#improved-library-builds

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`
    - [ ] No changes
     ***OR***
    - [x] API changes has been formally approved


LiKang6688:
Could you clean your commits a bit, please?

examples/dev-mashup/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "local-mashup",
   "version": "1.0.0",
   "description": "Simple local mashup",
-  "main": "index.html",
+  "main": "index.js",
```
LiKang6688, line 5: Do you know why the main is needed? I do not find the main section in https://github.com/qlik-oss/nebula.js/blob/master/commands/create/templates/mashup/_package.json.

johanlahti: It's for npm or for the compiler/build tool to know which file is the entry point of the application. So no real use for it in this case. We could also remove it completely but it will still default to `index.js` which is why I decided to keep it (no strong opinion about it though).

LiKang6688: Very good to know. Then it is fine to keep it.

LiKang6688 (approved):
There are too many unneeded commits that can be rebased, otherwise looks nice.

PR #895, opened 2022-08-16
feat: add variable and expression support to serve props

Caele:
## Motivation
Currently the cli-serve dev environment doesn't show property panel controls for all properties. For example all q-Props are skipped. This PR adds 
- exception for qStringExpressions: these will show up as a text field with a '=' appended
- support for variables of specific property structure:
```
variable: { name: '', value: '' }
```
Where value will be a qStringExpression trying to evaluate the variable.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`
    - [x] No changes
     ***OR***

### DIFF
--- a/commands/serve/lib/webpack.build.js
+++ b/commands/serve/lib/webpack.build.js
@@ -24,7 +24,7 @@ const cfg = ({ srcDir, distDir, dev = false, serveConfig = {} }) => {
       eHub: [path.resolve(srcDir, 'eHub')],
       fixtures: [path.resolve(__dirname, './fixtures.js')],
     },
-    devtool: dev ? 'eval-cheap-module-source-map' : false,
+    devtool: dev ? 'source-map' : false,
     output: {
       path: distDir,
       filename: '[name].js',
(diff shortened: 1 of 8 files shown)


commands/serve/web/components/FieldsPopover.jsx
```diff
@@ -142,7 +142,7 @@ export default function FieldsPopover({ alignTo, show, close, onSelected, type }
       open={show}
       onClose={close}
       anchorEl={alignTo.current}
-      marginThreshold={theme.spacing(1)}
+      marginThreshold={16} // theme.spacing(1)
```
Caele, line 145: Fixes a MUI warning

commands/serve/lib/webpack.build.js
```diff
@@ -24,7 +24,7 @@ const cfg = ({ srcDir, distDir, dev = false, serveConfig = {} }) => {
       eHub: [path.resolve(srcDir, 'eHub')],
       fixtures: [path.resolve(__dirname, './fixtures.js')],
     },
-    devtool: dev ? 'eval-cheap-module-source-map' : false,
+    devtool: dev ? 'source-map' : false,
```
Caele, line 27: Source maps wheren't working correctly.

T-Wizard:
I would expect that adding `--flags PP_EXPRESSIONS` to the `nebula serve` command on the command line to work (It does not)

Caele:
> I would expect that adding `--flags PP_EXPRESSIONS` to the `nebula serve` command on the command line to work (It does not)

Same here, not sure why it doesn't. WIll look into it. Have been using the config file instead.

T-Wizard (approved):

PR #910, opened 2022-09-05
feat(build): add system js build

streamside:
## Motivation

Add build using systemjs format which for example can be used in MFE context.

Changes:

* Add systemjs build which is performed when the property `systemjs` in `package.json` specifies where to output the bundle. No build performed in case no filename is present.
* Configure external dependencies. Specify an array at path `systemjs.external` in `nebula.config`.
* Refactor: Inject variations in behaviour per build format to `config()` in order to prevent if/elseif/else for each format. Each format specifies, for example, what the output file name should be.
* Refactor: Rename UMD build to `umd` (was `minified` which made more sense when there was only one format to build) 

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`
    - [x] No changes
     ***OR***

### DIFF
--- a/commands/build/lib/build.js
+++ b/commands/build/lib/build.js
@@ -29,10 +29,6 @@ const resolveReplacementStrings = (replacementStrings) => {
   return replacementStrings;
 };
 
-const mergeArray = (a1 = [], a2 = []) =>
-  // Simple merge and deduplication
-  [...a1, ...a2].filter((v, i, a) => a.indexOf(v) === i);
-
 const setupReactNative = (argv) => {
   const { reactNative } = argv;
   let reactNativePath;
@@ -224,8 +220,8 @@ const systemjs = async (argv, core) => {
     behaviours: {
       getExternal: ({ config: cfg }) => {
         const defaultExternal = ['@nebula.js/stardust', 'picasso.js', 'picasso-plugin-q', 'react', 'react-dom'];
-        const { external = [] } = cfg.systemjs || {};
-        return mergeArray(defaultExternal, external);
+        const { external } = cfg.systemjs || {};
+        return Array.isArray(external) ? external : defaultExternal;
       },
       getOutputFile: ({ pkg }) => pkg.systemjs,
       getOutputName: () => undefined,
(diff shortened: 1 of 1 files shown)


commands/build/lib/build.js
```diff
@@ -275,13 +318,19 @@ async function build(argv = {}) {
...
+
+  // Core
   if (argv.core) {
     const core = path.resolve(process.cwd(), argv.core);
     await esm(buildConfig, core);
+    await systemjs(buildConfig, core);
   }
-  return esm(buildConfig);
+
+  return Promise.resolve();
```
T-Wizard, line 333: there is no need to create an Promise here
```suggestion
  return undefined;
```
or remove the return (if it is not needed by lint rules)

streamside: Nice one. Was a bit too eager to not break backward compatibility here.

LiKang6688:
why is a SystemJS module always generated together (at the same time) with an ES Module?

streamside:
Update PR to use `systemjs` property in package.json to control where output of build is placed. No systemjs build is done in case no value is specified.

streamside:
> Do we want a SystemJS module always generated together (at the same time) with an ES Module?

With the latest changes in the PR it's possible for each repo to determine if a systemjs build should be done. The `systemjs` property in package.json controls where the build is placed and no build is done in case no value is specified.

veinfors (approved):

commands/build/lib/build.js
```diff
@@ -224,13 +227,14 @@ const systemjs = async (argv, core) => {
...
-        const { outputFile = 'systemjs/index.js' } = cfg.systemjs || {};
-        return outputFile;
-      },
+      getOutputFile: ({ pkg }) => pkg.systemjs,
       getOutputName: () => undefined,
+      enabled: ({ pkg }) => !!pkg.systemjs,
     },
   });
+  if (!c) {
+    return Promise.resolve();
```
T-Wizard, line 232: use
```suggestion
    return undefined;
```
inside an async function there is no need to create a new empty promise just returning undefined will have the same result.

T-Wizard (approved):

commands/build/lib/build.js
streamside: Fixed this occurrence and similar one when build ESM bundle.

PR #911, opened 2022-09-06
feat: enabled support for background image and color

chinmay-sg:
## Motivation

- Added support for background color in nucleus/src/components/Cell.jsx
- Added support for background image in nucleus/src/components/Supernova.jsx

### DIFF
--- a/apis/nucleus/src/components/Cell.jsx
+++ b/apis/nucleus/src/components/Cell.jsx
@@ -297,6 +297,19 @@ const Cell = forwardRef(
       },
     });
 
+    let bgColor = 'transparent';
+
+    if (
+      layout &&
+      layout.components &&
+      layout.components.find((comp) => comp.key === 'general') &&
+      layout.components.find((comp) => comp.key === 'general').bgColor &&
+      layout.components.find((comp) => comp.key === 'general').bgColor.color &&
+      layout.components.find((comp) => comp.key === 'general').bgColor.color.color
+    ) {
+      bgColor = layout.components.find((comp) => comp.key === 'general').bgColor.color.color;
+    }
+
     useEffect(() => {
       eventmixin(focusHandler.current);
     }, []);
@@ -516,6 +529,7 @@ const Cell = forwardRef(
             xs
             style={{
               height: '100%',
+              backgroundColor: bgColor,
             }}
             ref={contentRef}
           >
--- a/apis/nucleus/src/components/Supernova.jsx
+++ b/apis/nucleus/src/components/Supernova.jsx
@@ -119,11 +119,55 @@ function Supernova({ sn, snOptions: options, snPlugins: plugins, layout, appLayo
     keyboardNavigation,
   ]);
 
+  const imageSizingToCssProperty = {
+    originalSize: 'auto auto',
+    alwaysFit: 'contain',
+    fitWidth: '100% auto',
+    fitHeight: 'auto 100%',
+    stretchFit: '100% 100%',
+    alwaysFill: 'cover',
+  };
+
+  function getBackgroundPosition(data) {
+    let bkgImagePosition = 'center center';
+    if (data && data.components && data.components.length > 0 && data.components !== 'none') {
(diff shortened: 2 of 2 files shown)


apis/nucleus/src/components/Cell.jsx
```diff
@@ -297,6 +297,19 @@ const Cell = forwardRef(
...
 
+    let bgColor = 'transparent';
+
+    if (
+      layout &&
+      layout.components &&
+      layout.components.find((comp) => comp.key === 'general') &&
+      layout.components.find((comp) => comp.key === 'general').bgColor &&
+      layout.components.find((comp) => comp.key === 'general').bgColor.color &&
+      layout.components.find((comp) => comp.key === 'general').bgColor.color.color
```
Caele, line 308: simplify this:
const component = layout?.components?.find((comp) => comp.key === 'general');
if(component?.bgColor?.color) {
...
}
I think you can use the ? notation here, but if not you can at least stop it from running .find() 4 times!

apis/nucleus/src/components/Supernova.jsx
```diff
@@ -119,11 +119,55 @@ function Supernova({ sn, snOptions: options, snPlugins: plugins, layout, appLayo
...
-      style={{ position: 'relative', height: '100%' }}
+      style={{
+        position: 'relative',
+        height: '100%',
+        backgroundImage: `url(${
+          layout &&
+          layout.components &&
+          layout.components.find((comp) => comp.key === 'general') &&
+          layout.components.find((comp) => comp.key === 'general').bgImage &&
+          layout.components.find((comp) => comp.key === 'general').bgImage.mediaUrl
```
Caele, line 165: Same here, way too verbose. Save the varaible above it instead.

apis/nucleus/src/components/Cell.jsx
chinmay-sg: The thing is that for some reason, lint is not allowing to commit with the ? notation, it's throwing the following error.
 
Screenshot 2022-09-06 at 3 12 05 PM

I had done the same thing in a single line earlier, but it didn't allow me to commit

apis/nucleus/src/components/Supernova.jsx
```diff
@@ -119,11 +119,55 @@ function Supernova({ sn, snOptions: options, snPlugins: plugins, layout, appLayo
...
+    fitWidth: '100% auto',
+    fitHeight: 'auto 100%',
+    stretchFit: '100% 100%',
+    alwaysFill: 'cover',
+  };
+
+  function getBackgroundPosition(data) {
+    let bkgImagePosition = 'center center';
+    if (data && data.components && data.components.length > 0 && data.components !== 'none') {
+      const styling = data.components.find((comp) => comp.key === 'general');
```
Caele, line 134: again, this "finding" of the component should be done once per layout change.

chinmay-sg: ok, will change this

apis/nucleus/src/components/Cell.jsx
Caele: ok, but you can atlest simplify it to:
const component = layout &&  layout.components ? layout.components.find((comp) => comp.key === 'general') : null;

apis/nucleus/src/components/Cell.jsx
```diff
@@ -297,6 +297,19 @@ const Cell = forwardRef(
...
+    ) {
+      bgColor = layout.components.find((comp) => comp.key === 'general').bgColor.color.color;
+    }
```
Caele, line 311: Also, don't you need to go through the theme to resolve the color?
Some thing like:
halo.public.theme.getColorPickerColor(bgColor.color)
I assume this is done in the client, but in a slightly different way?

apis/nucleus/src/components/Cell.jsx
chinmay-sg: ok will do that

PR #912, opened 2022-09-07
feat(build): watch systemjs build

streamside:
## Motivation

Add support to output build on SystemJS format when watching.

Examples how to use it:

```sh
// Output build on UMD format (as-is today)
nebula build --watch

// Specify UMD format explicitly
nebula build --watch umd

// Specify SystemJS format
nebula build --watch systemjs
```

In case no `systemjs` bundle is specified in package.json the command fails gracefully informing the user about the problem.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`
    - [x] No changes
     ***OR***

### DIFF
--- a/commands/build/lib/init-config.js
+++ b/commands/build/lib/init-config.js
@@ -1,4 +1,4 @@
-/* eslint global-require: 0 */
+/* eslint global-require: 0, no-param-reassign: 0 */
 const fs = require('fs');
 
 const defaultFilename = 'nebula.config.js';
@@ -13,9 +13,8 @@ const options = {
   },
   watch: {
     description: 'Rebuild the bundle when its source files change on disk',
-    type: 'boolean',
     alias: 'w',
-    default: false,
+    choices: ['umd', 'systemjs', true],
   },
   sourcemap: {
     description: 'Generate source maps',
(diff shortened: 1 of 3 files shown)


commands/build/lib/init-config.js
```diff
@@ -40,22 +39,44 @@ const options = {
   },
 };
 
+// nebula build --watch                - watch umd bundle
+// nebula build --watch umd            - watch umd bundle
+// nebula build --watch systemjs       - watch systemjs bundle
+// nebula build                        - watch disabled
+const watchMiddleware = (argv) => {
```
streamside, line 46: @T-Wizard Ended up using a middleware to get the desired behaviour. Found the [coerce](http://yargs.js.org/docs/#api-reference-coercekey-fn) setting in yargs which would fit nicely with the use case at hand but it turns out that it doesn't work properly with our setup. The options seems to initialised twice and coerce is only run in the first execution.

streamside: An alternate solution to this is to use two options. `watch` controls if watch is enabled and `format` controls what to watch for.

```js
nebula build --watch                     - watch umd bundle
nebula build --watch --format umd        - watch umd bundle
nebula build --watch --format systemjs   - watch umd bundle
```

It's more verbose but to me a bit more explicit.

veinfors: It looks good to me!

veinfors (approved):

PR #917, opened 2022-09-08
refactor: support multiple dims for filterpane 

johanlahti:
## Motivation

During the develoment of the Nebula based Filterpane we need to be able to add more than one dimension. This is not possible in Nebula due to a limitation in the listbox handler, which always assumes one dim.

Visualisations using a hypercube, on the other hand, can add/remove dims and measures using the `qDimensions` and `qMeasures` array – but Listboxes do not have a corresponding array, since they are populated with `createChild()`.

This PR keeps track of dimensions outside of the handler and introduces a new Listbox container/Filterpane handler which makes use of the dimensions handled outside of its scope. 

https://user-images.githubusercontent.com/5780544/189156616-9eaaa971-9af8-4183-ad3e-4bb287f4c677.mov

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`
    - [x] No changes
     ***OR***


apis/nucleus/src/object/lo-handler.js
```diff
@@ -54,13 +43,15 @@ export default function loHandler({ dc: lo, def, properties }) {
     removeMeasure() {},
 
     maxDimensions() {
-      return 1;
+      const { max } = def.dimensions || {};
+      const maxDims = typeof max === 'function' ? max() : max;
+      return maxDims || 1;
```
T-Wizard, line 48: I think this is wrong. addDimension do still not support muliple dimension.
I am questioning if it make sense to extend list object to support multiple dimension. I think it would be more natural to add a new type of data instead.


for filterpane (assuming we keep the current property structure) I think it would be more natural to add an type of data object. Not extending list object (or hyper cube)

johanlahti: Yes, we can/should probably hard-code lo-handler to 1. Note though that the Filterpane uses the `lo-container-handler.js`, so we are not using `lo-handler.js` for multiple dims.

commands/serve/web/components/property-panel/DataCube.jsx
```diff
@@ -27,25 +28,37 @@ const getValue = (data, reference, defaultValue) => {
 };
 
 export default function DataCube({ setProperties, target, properties }) {
-  const createHandler = target.propertyPath.match('/qHyperCube') ? hcHandler : loHandler;
+  // Store dimensions outside of handler to support filterpanes or other containers
+  // holding list objects, which do not store dimensions in an array (like hc.qDimensions).
+  const [dimensions, setDimensions] = useState([]);
```
T-Wizard, line 33: I don't see how this work on existing objects that already have dimensions?

johanlahti: Yeah, it won't. It's not a huge limitation though since this is a dev tool and sufficient for continuing the Filter pane work.
I'm also not sure if we can deduce this in any way from the arguments we have here. In that case that would be a more reliable source for extracting dimensions.

johanlahti:
PTAL

johanlahti:
If there are no *concrete code suggestions* on how to solve this in a better way my arguments for merging this PR are:
- We need this to facilitate the dev of the Filterpane
- It should not destroy any existing logic
- The change applies narrowly to ListObject data supporting multiple dimensions
- This is a dev environment made by developers for developers – let's be pragmatic
- Nothing stops us from improving this later on if we are not happy with the esthetics of the code

T-Wizard:
I do not think that a filterpane that works with this api will also work well inside sense-client (but I have not taken the time to explore it)

johanlahti:
In sense-client they have made a similar workaround as I have here, with the addition that they derive the dimensions from the layout (again, since it's not possible to do from the object definition). I'd rather put the effort in creating that solution in sense-client than in Nebula, for the moment at least.

veinfors:
> If there are no _concrete code suggestions_ on how to solve this in a better way my arguments for merging this PR are:
> 
> * We need this to facilitate the dev of the Filterpane
> * It should not destroy any existing logic
> * The change applies narrowly to ListObject data supporting multiple dimensions
> * This is a dev environment made by developers for developers – let's be pragmatic
> * Nothing stops us from improving this later on if we are not happy with the esthetics of the code

I have looked at the code and I think it feels OK to bring this in. It will add value without screwing up the code quality too much in my opinion.

veinfors:
Remove the `yarn.lock` file... It's probably included in this PR by mistake.

veinfors (approved):

PR #919, opened 2022-09-09
feat(serve): add mfe mode

streamside:
## Motivation

Serve visualisation bundle to use in microfront end (MFE) environment. The MFE mode is activated with the option `--mfe`.

Example starting Nebula serve in MFE mode:

```sh
nebula serve --mfe
```

This allows visualisation developers to spin up nebula serve, override the import map in the MFE environment (for example QCS) and do development against it. When the source code is changed the bundle is automatically rebuilt.

The MFE mode requires the server to run with HTTPS. This in turn requires certificates to be installed on environment spinning up nebula serve.

**Use option `--systemjs` instead of `--mfe`?**

`--mfe` is intended to describe the purpose for the developer while `--systemjs` is how its technically achieved. For the time being systemjs is the format for the bundle. Later on it might change to ESM.

When using `--mfe` some other aspects of nebula serve changes. For example,  the normal nebula serve dev env is disabled and the server requires certificates to be installed. Is this what the user expects when starting with the flag `--systemjs`?

## Requirements checklist

- [x] Api specification
  - [ ] Ran `yarn spec`
    - [x] No changes
     ***OR***

### DIFF
--- a/commands/serve/README.md
+++ b/commands/serve/README.md
@@ -112,7 +112,7 @@ The regular dev environment in nebula serve is disabled when running in this mod
 There are a few ways to install local trusted certificates on your machine, The important
 end result is that there are two files `~/.certs/cert.pem` (the certificate) and
 `~/.certs/key.pem` (the public key). Read about how certificates work
-[here](http://www.steves-internet-guide.com/ssl-certificates-explained/) if you already
+[here](http://www.steves-internet-guide.com/ssl-certificates-explained/). If you already
 have a self-signed and trusted certificate in this location, then skip this guide.
 
 ##### Easy step-by-step guide to install and generate certificates locally
--- a/commands/serve/README.md
+++ b/commands/serve/README.md
@@ -37,6 +37,7 @@ Options:
   --webIntegrationId                                                    [string]
   --fixturePath       Path to a folder that will be used as basis when locating
                       fixtures              [string] [default: "test/component"]
+  --mfe               Serves bundle to use in micro frontend           [boolean]
   -h, --help          Show help                                        [boolean]
 ```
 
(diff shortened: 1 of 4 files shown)


commands/serve/README.md
```diff
@@ -88,3 +89,49 @@ serve({
   s.close(); // close the server
 });
 ```
+
+### Micro Frontend (MFE) Mode
+
+The MFE mode, activated with the `--mfe` option, builds a visualisation bundle which is
+used in a micro frontend environemnt. The bundle is served at:
```
veinfors, line 96:
```suggestion
The MFE mode, activated with the `--mfe` option, builds a visualisation which is
used in a micro frontend environemnt. The visualisation is served at:
```

Avoid possible confusion with the "Visualization bundle" that exist in Qlik Sense

commands/serve/README.md
```diff
@@ -88,3 +89,49 @@ serve({
...
+
+```
+https://<host>:<port>/pkg/<module name>
+```
+
+The module name is by default the name in `package.json` but may be overriden using the
+`--type` option.
+
+The MFE mode requires HTTPS which requires certificates to be installed in the
+environment running nebula serve.
```
veinfors, line 106:
```suggestion
The MFE mode runs in HTTPS which requires certificates to be installed in the
environment running nebula serve.
```

commands/serve/README.md
```diff
@@ -88,3 +89,49 @@ serve({
...
+
+The regular dev environment in nebula serve is disabled when running in this mode.
+
+#### How to install trusted certificates locally
+
+There are a few ways to install local trusted certificates on your machine, The important
+end result is that there are two files `~/.certs/cert.pem` (the certificate) and
+`~/.certs/key.pem` (the public key). Read about how certificates work
+[here](http://www.steves-internet-guide.com/ssl-certificates-explained/) if you already
+have a self-signed and trusted certificate in this location, then skip this guide.
```
veinfors, line 116:
```suggestion
[here](http://www.steves-internet-guide.com/ssl-certificates-explained/). If you already
have a self-signed and trusted certificate in this location, then skip this guide.
```

commands/serve/lib/webpack.serve.js
```diff
@@ -192,8 +218,25 @@ module.exports = async ({
     compiler.hooks.done.tap('nebula serve', (stats) => {
       if (!initiated) {
         initiated = true;
-        const url = `http://${host}:${port}`;
-        console.log(`Development server running at ${chalk.green(url)}`);
+
+        console.log(`     _  _________  __  ____   ___
```
streamside, line 222: https://patorjk.com/software/taag/#p=testall&f=Graffiti&t=Nebula%20serve

veinfors (approved):

T-Wizard (approved):
Accepted but.

I would use `--systemjs` over `--mfe`
as I think that changing the output format of `--mfe` is a breaking change and should be avoided

I think also that an option for serving a bundle over http should be considered.

Caele (approved):

PR #921, opened 2022-09-13
feat: enabled support for background image and color

chinmay-sg:
## Motivation

- Added support for background color in nucleus/src/components/Cell.jsx
- Added support for background image in nucleus/src/components/Supernova.jsx


apis/nucleus/src/components/Supernova.jsx
```diff
@@ -119,11 +122,49 @@ function Supernova({ sn, snOptions: options, snPlugins: plugins, layout, appLayo
...
+    if (data && data.components && data.components.length > 0 && data.components !== 'none') {
+      const styling = data.components.find((comp) => comp.key === 'general');
+      if (styling && styling.bgImage && styling.bgImage.position) {
+        bkgImagePosition = styling.bgImage.position.replace('-', ' ');
+      }
+    }
+    return bkgImagePosition;
+  }
+
+  function getBackgroundSize(data) {
```
Caele, line 145: same here

apis/nucleus/src/components/Supernova.jsx
```diff
@@ -35,6 +35,9 @@ function Supernova({ sn, snOptions: options, snPlugins: plugins, layout, appLayo
     setSnNode(ref);
   }, []);
 
+  const bgComp = layout && layout.components ? layout.components.find((comp) => comp.key === 'general') : null;
+  const bgImage = bgComp && bgComp.bgImage && bgComp.bgImage.mediaUrl ? bgComp.bgImage.mediaUrl : undefined;
```
Caele, line 39: Should we also support expressionUrl?

chinmay-sg: We could do that, but then the `qStringExpression` would need to be evaluated through the expression editor first, right?

chinmay-sg: In that case, could we just put the evaluated URL and the mediaUrl in the same property, `qStaticContentUrlDef` in the mediaURL object and then render the image based on that?

Caele: Given this spec we agreed upon earlier:
```
styles: {
  bgColor: {
    color: { ...  } // support for 'none' 
  },
  bgImage: {
    mode: none, media, expression
    position: top-left, top-center, top-right, center-left, center-center. center-right, bottom-left, bottom-center, bottom-right,
    sizing: original, alwaysFit, alwaysFill, stretch, fitWidth, fitHeight
    mediaUrl: { qStaticContentUrlDef...}
    expressionUrl: { qStringExpression... }
  }
}
```
You should be able to get the image URL either from the mediaUrl property or from the expressionUrl property. This should be based of the mode property.

apis/nucleus/src/components/Cell.jsx
```diff
@@ -290,17 +290,36 @@ const Cell = forwardRef(
...
+      const bgComp = layout && layout.components ? layout.components.find((comp) => comp.key === 'general') : null;
+
+      if (bgComp && bgComp.bgColor && bgComp.bgColor.useColorExpression) {
+        if (bgComp.bgColor.urlColor) {
+          bgComp.bgColor.urlColor = { color: bgComp.bgColor.urlColor };
+        }
+        return bgComp.bgColor.urlColor ? halo.public.theme.getColorPickerColor(bgComp.bgColor.urlColor) : undefined;
+      }
+      if (bgComp && bgComp.bgColor && !bgComp.bgColor.useColorExpression)
+        return bgComp.bgColor.cpColor ? halo.public.theme.getColorPickerColor(bgComp.bgColor.cpColor) : undefined;
```
Caele, line 311: This is incorrect I think?
The property should be:
```
components[N] = {
  key: "general",
  bgColor: {
    color: { ...  } // support for 'none'  
  },
  bgImage: {
    mode: none, media, expression
    position: top-left etc...
    sizing: original, alwaysFit, alwaysFill, stretch, fitWidth, fitHeight
    mediaUrl: { qStaticContentUrlDef...}
    expressionUrl: { qStringExpression... }
  }
}

```
Should be `color` not `cpColor`.

apis/nucleus/src/components/Cell.jsx
```diff
@@ -290,17 +290,40 @@ const Cell = forwardRef(
...
         this.emit(last ? 'focus_toolbar_last' : 'focus_toolbar_first');
       },
     });
 
+    const resolveBgColor = () => {
+      const bgComp = layout && layout.components ? layout.components.find((comp) => comp.key === 'general') : null;
+
+      if (bgComp && bgComp.bgColor && bgComp.bgColor.useColorExpression) {
+        if (bgComp.bgColor.colorExpression) {
+          bgComp.bgColor.colorExpression = { color: bgComp.bgColor.colorExpression };
```
Caele, line 306: We should parse the color here as well, bascially use d3-color. I'm gonna check if we should expose that from "theme" maybe.

chinmay-sg: I guess d3-color is being used by "theme", would I need to add it in nucleus as well, or should we wait for the exposure from theme?

PR #922, opened 2022-09-13
fix(serve): add cors headers to `/info` and `/pkg/<name>`

streamside:
## Motivation

CORS headers needed when serving visualisation artifacts to a MFE environment.

The headers are configured in the options for webpack devserver. These headers are added to responses for requests to `/info` and `/pkg/<name>`.

Question: Would it be useful to allow the headers to be configurable in nebula.config.js?

## Requirements checklist

- [x] Api specification
  - [X] Ran `yarn spec`
    - [ ] No changes
     ***OR***
    - [x] API changes has been formally approved

### DIFF
--- a/commands/serve/lib/serve.js
+++ b/commands/serve/lib/serve.js
@@ -47,7 +47,7 @@ const initiateWatch = async ({ snPath, snName, host }) => {
   const cache = {};
 
   return {
-    addRoutes(app) {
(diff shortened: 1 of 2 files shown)


Caele:
We are in the works to move to Oauth (instead the webIntegrationId cookies), which means that this will also need an update.

PR #926, opened 2022-09-19
feat: add support to run callback when context menu opens from sense-client

Donya-qlik:
## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`
    - [x] No changes
     ***OR***

### DIFF
--- a/apis/stardust/api-spec/spec.json
+++ b/apis/stardust/api-spec/spec.json
@@ -487,32 +487,6 @@
         "import { onTakeSnapshot } from '@nebula.js/stardust';\nimport { useState } from '@nebula.js/stardust';\nimport { useLayout } from '@nebula.js/stardust';\n\nconst layout = useLayout();\nconst [zoomed] = useState(layout.isZoomed || false);\n\nonTakeSnapshot((copyOfLayout) => {\n  copyOfLayout.isZoomed = zoomed;\n  return Promise.resolve(copyOfLayout);\n});"
       ]
     },
-    "onContextMenu": {
-      "description": "Registers a callback that is called when a cell value is copied.",
-      "kind": "function",
-      "params": [
-        {
-          "name": "addItemCallback",
-          "kind": "function",
-          "params": [
-            {
-              "type": "any"
-            }
-          ],
-          "returns": {
-            "type": "Promise",
-            "generics": [
-              {
-                "type": "any"
-              }
-            ]
-          }
-        }
-      ],
-      "examples": [
-        "import { onContextMenu } from '@nebula.js/stardust';\nonContextMenu((menu, event) => {\n  return Promise.resolve(menu);\n});"
-      ]
-    },
     "useRenderState": {
       "description": "Gets render state instance.\n\nUsed to update properties and get a new layout without triggering onInitialRender.",
       "stability": "experimental",
--- a/apis/stardust/api-spec/spec.json
+++ b/apis/stardust/api-spec/spec.json
@@ -487,6 +487,32 @@
         "import { onTakeSnapshot } from '@nebula.js/stardust';\nimport { useState } from '@nebula.js/stardust';\nimport { useLayout } from '@nebula.js/stardust';\n\nconst layout = useLayout();\nconst [zoomed] = useState(layout.isZoomed || false);\n\nonTakeSnapshot((copyOfLayout) => {\n  copyOfLayout.isZoomed = zoomed;\n  return Promise.resolve(copyOfLayout);\n});"
       ]
     },
+    "onContextMenu": {
+      "description": "Registers a callback that is called when a cell value is copied.",
+      "kind": "function",
+      "params": [
+        {
+          "name": "addItemCallback",
+          "kind": "function",
+          "params": [
+            {
+              "type": "any"
+            }
+          ],
+          "returns": {
+            "type": "Promise",
+            "generics": [
+              {
+                "type": "any"
+              }
+            ]
+          }
+        }
+      ],
+      "examples": [
+        "import { onContextMenu } from '@nebula.js/stardust';\nonContextMenu((menu, event) => {\n  return Promise.resolve(menu);\n});"
+      ]
+    },
     "useRenderState": {
       "description": "Gets render state instance.\n\nUsed to update properties and get a new layout without triggering onInitialRender.",
       "stability": "experimental",
--- a/apis/stardust/types/index.d.ts
+++ b/apis/stardust/types/index.d.ts
@@ -154,6 +154,13 @@ export function useOptions(): object;
 export function onTakeSnapshot(snapshotCallback: ($: EngineAPI.IGenericObjectLayout)=>Promise<EngineAPI.IGenericObjectLayout>): void;
 
 /**
+ * Registers a callback that is called when a context menu item is added.
+ * @param addItemCallback
+ */
+ export function onContextMenu(addItemCallback: (menu: Menu, event: any)=>Promise<menu>): void;
+
+ 
+/**
  * Gets render state instance.
  * 
  * Used to update properties and get a new layout without triggering onInitialRender.
@@ -693,5 +700,9 @@ declare namespace stardust {
     interface hyperCubeConversion {
     }
 
+}
+
+export interface Menu {
+    addItem: (input: { translation: string; icon: string; tid: string; select: () => Promise<void> }) => void;
 }
 
--- a/apis/stardust/types/index.d.ts
+++ b/apis/stardust/types/index.d.ts
@@ -154,13 +154,6 @@ export function useOptions(): object;
 export function onTakeSnapshot(snapshotCallback: ($: EngineAPI.IGenericObjectLayout)=>Promise<EngineAPI.IGenericObjectLayout>): void;
 
 /**
- * Registers a callback that is called when the context menu opens
- * @param addItemCallback
- */
- export function onContextMenu(addItemCallback: (menu: any, event: any) => void): void;
-
- 
-/**
  * Gets render state instance.
  * 
  * Used to update properties and get a new layout without triggering onInitialRender.
@@ -701,3 +694,4 @@ declare namespace stardust {
     }
 
 }
+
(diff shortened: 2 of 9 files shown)


apis/stardust/api-spec/spec.json
```diff
@@ -487,6 +487,32 @@
         "import { onTakeSnapshot } from '@nebula.js/stardust';\nimport { useState } from '@nebula.js/stardust';\nimport { useLayout } from '@nebula.js/stardust';\n\nconst layout = useLayout();\nconst [zoomed] = useState(layout.isZoomed || false);\n\nonTakeSnapshot((copyOfLayout) => {\n  copyOfLayout.isZoomed = zoomed;\n  return Promise.resolve(copyOfLayout);\n});"
       ]
     },
+    "onContextMenu": {
```
Donya-qlik, line 490: Should it be added here ?

docs/api-reference.md
```diff
@@ -509,6 +510,21 @@ onTakeSnapshot((copyOfLayout) => {
...
+Registers a callback that is called when a a cell value is copied.
+
+```js
+import { onContextMenu } from '@nebula.js/stardust';
+
+onContextMenu((menu, event) => {
+  menu.addItem(item, index);
+  return Promise.resolve(menu);
+});
+```
```
Donya-qlik, line 526: Should I have this code block here ?

docs/api-reference.md
```diff
@@ -30,6 +30,7 @@ title: API Reference
 - [function: useConstraints()](#function-useconstraints)
 - [function: useOptions()](#function-useoptions)
 - [function: onTakeSnapshot(snapshotCallback)](#function-ontakesnapshotsnapshotcallback)
+- [function: onContextMenu(addItemCallback)](#function-oncontextmenuadditemcallback)
```
Donya-qlik, line 33: .. I guess it should be added or ?

haxxmaxx (commented):
I'll let Tobias have a look as well

apis/stardust/api-spec/spec.json
Caele: hmm, I think we said that it should not be added yet. Thats what the "@ignore" should do. Try re-running yarn spec see if it is removed.

apis/supernova/src/hooks.js
```diff
@@ -990,6 +1003,28 @@ export function onTakeSnapshot(cb) {
   h.fn = cb;
 }
 
+/**
+ * Registers a callback that is called when a context menu item is added.
+ * @entry
+ * @param {function(menu, event): Promise<menu>} addItemCallback
+ * @example
+ * @ignore
```
Caele, line 1011: move this up above example, I think that is what messes with the notation.

Caele, line 1011:
```suggestion
 * @ignore
 * @example
```

apis/stardust/api-spec/spec.json
Caele: found it, your ignore was inside the example attribute

docs/api-reference.md
Caele: Nope

apis/stardust/types/index.d.ts
```diff
@@ -695,3 +702,7 @@ declare namespace stardust {
 
 }
 
+export interface Menu {
+    addItem: (input: { translation: string; icon: string; tid: string; select: () => Promise<void> }) => void;
+}
+
```
Caele, line 708: Not sure where this come from?

apis/stardust/api-spec/spec.json
Donya-qlik: ... so you say it is ok to have it here anyway ?

apis/stardust/types/index.d.ts
Donya-qlik: Ah this one, in the above lines (160) I set menu type as `Menu`, it has certain properties, to set their types I chose to add an interface for it.

apis/supernova/src/creator.js
```diff
@@ -202,8 +203,8 @@ function createWithHooks(generator, opts, galaxy) {
     setSnapshotData(layout) {
       return generator.component.runSnaps(this, layout);
     },
-    focus() {
-      generator.component.focus(this);
+    onContextMenu(menu, event) {
```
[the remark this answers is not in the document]
Donya-qlik: Oh, it wasn't supposed to be removed, I'll bring it back

docs/api-reference.md
Donya-qlik: Ok I'll remove it. btw, the similar hook for snapshot is here, shouldn't the onContextMenu be as well ?

apis/supernova/src/hooks.js
```diff
@@ -990,6 +1003,28 @@ export function onTakeSnapshot(cb) {
   h.fn = cb;
 }
 
+/**
+ * Registers a callback that is called when a context menu item is added.
+ * @entry
+ * @param {function(menu, event): Promise<menu>} addItemCallback
```
T-Wizard, line 1009: why do the a callback return Promise<menu> and not Promise<undefined>?
(It feels just like an unnecessary complication)

and why this API (why a callback)?
and not a object describing a single menu item matching how `useAction` works

apis/stardust/types/index.d.ts
Donya-qlik: Oh..! Ok I'll remove it then!

apis/supernova/src/hooks.js
haxxmaxx: @Caele suggested to implement it similar to `setSnapshotData`, so we followed the same pattern. You pass a callback where you can do whatever you need. As I understand it this might be revisited when there is a context menu implementation for nebula, right @Caele?

apis/stardust/types/index.d.ts
Donya-qlik: BTW @T-Wizard , as I see, the file has been modified manually several times by others! Is it something wrong in my case ?

apis/supernova/src/hooks.js
Donya-qlik: @Caele, Are we right ?

Donya-qlik: @T-Wizard as Max explained here menu is returned since we are implementing this function in sense-client and now the new item (CopyCellValue) has been added to the list of context menu items.

apis/stardust/types/index.d.ts
T-Wizard: you can (and if relevant should) regenerate the file by running `yarn spec` but you should not do any extra edit of file after that as they are lost next time the file is generated

apis/supernova/src/hooks.js
T-Wizard: I can see the reason to use a callback bacause it is similar to how it works in sense-client (and maybe becasue it is similar to setSnapshotData).
(I still think we should take some time to consider the api some more to reduce the need to change it later)

the return values is a separate point and I don't think the callback should br required to return the menu.
(I think it will always be the first argument as I don't see the point of returning it)
returning a promise could be relevant if the callback should be able to do something that takes time.

Donya-qlik: Do you think we should apply this consideration in the current PR? if so then I should discuss it with @haxxmaxx  and @Caele.

Yes. It is already fixed, `export function onContextMenu(addItemCallback: (menu: Menu, event: any) => menu): void;` , I think it is what you meant, (just a small miss spell I found, I'm going to fix it).

T-Wizard: no I meant to change it to `export function onContextMenu(addItemCallback: (menu: Menu, event: any) => Promise<void>): void;`

Donya-qlik: Ok sorry I misunderstood. Umm.. event then I don't think Promise is needed, there is nothing that takes time to do it. I'll try return void and will see.

apis/stardust/types/index.d.ts
```diff
@@ -153,6 +153,13 @@ export function useOptions(): object;
  */
 export function onTakeSnapshot(snapshotCallback: ($: EngineAPI.IGenericObjectLayout)=>Promise<EngineAPI.IGenericObjectLayout>): void;
 
+/**
+ * Registers a callback that is called when a context menu item is added.
```
haxxmaxx, line 157: Not exactly, the callback runs when you open the context menu

apis/stardust/types/index.d.ts
```diff
@@ -695,3 +702,7 @@ declare namespace stardust {
 
 }
 
+export interface Menu {
```
haxxmaxx, line 705: Is this generated or added manually?

apis/supernova/src/hooks.js
```diff
@@ -990,6 +1003,27 @@ export function onTakeSnapshot(cb) {
   h.fn = cb;
 }
 
+/**
+ * Registers a callback that is called when a context menu item is added.
```
haxxmaxx, line 1007: I guess this is where you need to change to when the context menu opens

haxxmaxx (commented):
Code looks good, since it is the same as for snapshots. Not sure about documentation though, I'll let Tobias and Tobias look at that. Also fill in the form in the description

apis/stardust/types/index.d.ts
```diff
@@ -153,6 +153,13 @@ export function useOptions(): object;
  */
 export function onTakeSnapshot(snapshotCallback: ($: EngineAPI.IGenericObjectLayout)=>Promise<EngineAPI.IGenericObjectLayout>): void;
 
+/**
+ * Registers a callback that is called when the context menu opens
+ * @param addItemCallback
+ */
+ export function onContextMenu(addItemCallback: (menu: any, event: any) => void): void;
+
+
```
T-Wizard, line 162: In my opinion this file should only be the unchanged auto generated output.
meaning onContextMenu should not be included as long as it is marked by `@ignore`

Donya-qlik: I don't know how it is still here, I reverted my last commit ! I'll remove it.

haxxmaxx (approved):

T-Wizard (approved):

PR #932, opened 2022-09-20
fix(nucleus): apply styles on header element

streamside:
## Motivation

Modify container styles to apply directly on container element (instead of being applied to sub elements of the container).

Example **without** the fix. Note that there is no padding below the header:

December 2018

Example **with** the fix. Note the 8px padding below the header:

December 2019

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`
    - [x] No changes
     ***OR***

### DIFF
--- a/apis/nucleus/src/components/Header.jsx
+++ b/apis/nucleus/src/components/Header.jsx
@@ -14,11 +14,11 @@ const classes = {
 };
 
 const StyledGrid = styled(Grid)(({ theme }) => ({
-  [`& .${classes.containerStyle}`]: {
+  [`&.${classes.containerStyle}`]: {
     flexGrow: 0,
   },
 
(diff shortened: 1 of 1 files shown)


Caele:
I assume this is a big from the MUIv5 conversion, how did you notice it? As in, are we sure that there was a padding there before?

streamside:
> I assume this is a big from the MUIv5 conversion, how did you notice it? As in, are we sure that there was a padding there before?

I noticed it when upgrading to Nebula@3 in a visualisation repo and the rendering tests failed. When looking closer the images are 8px taller after the upgrade. This was due to that there was no padding below the header, hence the vis got some additional room to render.

The rendering tests runs green when running on Nebula@2.13.0 or earlier. I have also run the tests manually and verified that there was a padding before the upgrade.

Caele (approved):

PR #936, opened 2022-09-23
refactor(ListBox): Hide search and call abort search on selection deactivate

johanlahti:
## Motivation

Part of fix(EN-4079).

#### Changes:

Feature parity with sense-client's Listbox, which means abort search whenever:
- Search deactivates (toggle off)
- Confirm/Cancel selections
- Press Escape

#### Not accounted for (but accounted for in existing PR for sense-client):
- Clicking outside a Listbox will neither deactivate selections nor abort search

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`
    - [x] No changes
     ***OR***


apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -7,46 +7,67 @@ import InstanceContext from '../../../contexts/InstanceContext';
...
+      setSearchApplied(true);
+      return model.searchListObjectFor(TREE_PATH, e.target.value);
     }
+    return undefined;
   };
+
+  const handleFocus = () => {
+    if (!selections.isModal(model)) {
+      selections.begin(['/qListObjectDef']).then(() => {
+        selections.goModal && selections.goModal('/qListObjectDef');
```
T-Wizard, line 51: why?
it looks like selections.begin already are doing this

johanlahti: Yes, it can be removed.

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -7,46 +7,72 @@ import InstanceContext from '../../../contexts/InstanceContext';
 
 const TREE_PATH = '/qListObjectDef';
 
-export default function ListBoxSearch({ model, keyboard, dense = false, visible = true }) {
+export default function ListBoxSearch({ selections, model, keyboard, dense = false, visible = true }) {
   const { translator } = useContext(InstanceContext);
   const [value, setValue] = useState('');
-  const [isSearching, setIsSearching] = useState(false);
+  const [, setAppliedSearch] = useState(false);
```
Caele, line 13: Whats the point of a state that is never used?

johanlahti: Yeah, it looks funny I know. But if you look further down in the code you see how the state is read in the `set` callback. The reason for doing like this is that the state is otherwise not up do date with the latest `appliedSearch`.
Anyhow, I had to refactor this again and got rid of this flag altogether now.

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -7,46 +7,72 @@ import InstanceContext from '../../../contexts/InstanceContext';
...
   const theme = useTheme();
 
+  const cancel = () => selections.isActive() && selections.cancel();
+
   const abortSearch = async () => {
-    await model.abortListObjectSearch(TREE_PATH);
-    setIsSearching(false);
+    if (!selections.isModal()) {
+      return;
+    }
```
T-Wizard, line 20: This only works if the listbox uses model selection.
That is currently true but we may want to change that in the future.
(More specific I think that the use in analytic creation we may want to try using listboxes without modal selection)

johanlahti: Ok, good to know. For this fix I don't think we need to consider these anticipated changes though, since we probably have this dependency elsewhere in our applications that needs to be removed when that day comes. `isActive()` is not a direct equivalent of `isModal` unfortunately, don't know if we have any good proxy for this, do you know?

johanlahti:
?

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -7,46 +7,68 @@ import InstanceContext from '../../../contexts/InstanceContext';
...
     setValue(e.target.value);
-    if (e.target.value.length === 0) {
-      abortSearch();
-    } else {
-      model.searchListObjectFor(TREE_PATH, e.target.value);
-      setIsSearching(true);
+    if (e.target.value.length) {
+      return model.searchListObjectFor(TREE_PATH, e.target.value);
     }
+    return undefined;
```
T-Wizard, line 46: why do onChange no longer do anything when searchstring is changed to empty? (it feels wrong)

johanlahti: We don't want to do anything when the search string is empty.
The `abortSearch` call for empty strings was introduced as a quick fix this summer (not sure who did that). To achieve feature parity with the current Listbox, we should not call abort search in this case. We should only do that when deactivating selections actively or passively, e.g. through clicking outside the viz or pressing Esc.

T-Wizard: the sense-client do call abortListObjectSearch in this case.
how does the feature parity with sense-client work without it?

T-Wizard: Testing the the current selection listbox popover in nebula serve. I can not stop searching without closing the popover and reopening it. (just removing the search string keeps serching for the last letter removed)

T-Wizard: I would expect is to also be possible to remove the search string to end the search. (matching how it works in sense-client)

johanlahti: @T-Wizard How do you search with an empty string in a Filterpane on Stage?

johanlahti: For the records, the case referred to here is when:
1. performing a search
2. clearing the search (no submit needed)

I have adapted the code to accomodate for this case too. Thanks @T-Wizard for finding this.

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -7,46 +7,68 @@ import InstanceContext from '../../../contexts/InstanceContext';
...
+    } finally {
+      setValue('');
+    }
   };
 
   useEffect(() => {
-    if (!visible && isSearching) {
-      abortSearch();
+    if (!visible) {
+      cancel(); // cancel and abort when toggling off search
```
T-Wizard, line 30: why cancel selections here?

in client (old filterpane) it is possible to hide the search without cancel selection
why should this work different?

PR #938, opened 2022-09-26
feat(serve): pass fullscreen callback to nebbie options

DanielS-Qlik:
## Motivation
While using sn-filter-pane in nebula-serve,
the expand-button shown in sn-filter-pane when all Listboxes are collapsed needs a callback function that trigger expand.
fs

## Requirements checklist

- [x] Api specification
  - [ ] Ran `yarn spec`
    - [x] No changes
     ***OR***

### DIFF
--- a/commands/serve/web/components/Cell.jsx
+++ b/commands/serve/web/components/Cell.jsx
@@ -114,6 +114,19 @@ export default function ({ id, expandable, minHeight }) {
     [model, currentThemeName]
   );
 
+  const singleRenderUrl = (modelId) => {
+    if (model || modelId) {
+      return `${document.location.href.replace(/\/dev\//, '/render/')}${window.location.search ? '&' : '?'}object=${
+        model ? model.id : modelId
+      }&theme=${currentThemeName}&language=${language}`;
+    }
+    return '';
+  };
+
+  const openInSingleRender = (modelId) => {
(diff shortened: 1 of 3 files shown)


commands/serve/web/components/Cell.jsx
```diff
@@ -152,13 +165,7 @@ export default function ({ id, expandable, minHeight }) {
...
             className={classes.secondaryIcon}
             disabled={!model}
-            href={
-              model
-                ? `${document.location.href.replace(/\/dev\//, '/render/')}${
-                    window.location.search ? '&' : '?'
-                  }object=${model.id}&theme=${currentThemeName}&language=${language}`
-                : ''
-            }
+            href={singleRenderUrl()}
```
LiKang6688, line 168: The refactor is really good. But it is not relevant to this PR. Would you mind doing it in another PR?

veinfors: I would argue that it is relevant to this PR. The alternative to not doing this refactoring would be to add duplicated code.

LiKang6688: ahh, I see. singleRenderUrl is used in openInSingleRender.

Caele:
Is this only for `serve` and only for dev purposes?

PR #957, opened 2022-10-11
feat: render existing listbox objects

veinfors:
Allow rendering of existing listbox objects (currently it's only possible to render/mount fields and master items which are then created as session objects) by passing in { qId: "generic-object-id" } into mount `embed.field(` function.

--- TODO before merging ---

- [x] Fix and add unit tests
- [x] Regression testing


apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -192,7 +191,7 @@ export default function ListBox({
     if (loaderRef.current) {
       loaderRef.current.resetloadMoreItemsCache(true);
       // Skip scrollToItem if we are in selections, or if we dont sort by state.
```
Caele, line 193: Updat the comment as well

apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+  useEffect(() => {
+    async function fetchObject() {
+      const modelFromStore = modelStore.get(qId);
+      if (modelFromStore) {
+        setModel(modelFromStore);
+      } else {
+        const m = await app.getObject(qId);
+        const key = m ? `${m.id}` : null;
+        modelStore.set(key, m);
```
johanlahti, line 17: If `key === null` should we still store `m` with key `null`? Throw error instead?

johanlahti (approved):
Makes code easier to read with constants added and hopefully better performance/fewer requests. 👍

DanielS-Qlik (approved):

apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
veinfors: I removed the row
`const key = m ? `${m.id}` : null;`

enigma with throw an error if the object isn't found anyway...

PR #959, opened 2022-10-11
refactor(filter-pane-handler): Add filter pane handler without messing with dimension creation

johanlahti:
## Motivation

The same code as in https://github.com/qlik-oss/nebula.js/pull/917 but now does not alter the way we create dimensions (which forced a revert).

Dimensions are now returned from `handler.addDimension()` so that we can manage >1 dim outside of the handler (only relevant for `lo-container-handler` since listbox data does not have an array like `qDimensions` for appending dims).

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`
    - [x] No changes
     ***OR***


apis/nucleus/src/object/lo-handler.js
```diff
@@ -60,7 +62,7 @@ export default function loHandler({ dc: lo, def, properties }) {
       return 0;
     },
     canAddDimension() {
-      return handler.dimensions().length === 0;
+      return handler.dimensions().length < handler.maxDimensions();
```
Caele, line 65: Still should still be strickly for zero dims right? Else you'd end up in the other handler, so a change here is just more confusing? or?

commands/serve/web/components/property-panel/DataCube.jsx
```diff
@@ -27,25 +28,38 @@ const getValue = (data, reference, defaultValue) => {
...
   );
 
   const onDimensionAdded = (a) => {
-    handler.addDimension(typeof a === 'object' ? { qLibraryId: a.qId } : a);
-    setProperties(properties);
+    const dim = typeof a === 'object' ? { qLibraryId: a.qId } : a;
+    const addedDim = handler.addDimension(dim);
+    setProperties(properties).then(() => {
+      setDimensions([...dimensions, addedDim]);
+    });
```
Caele, line 56: Nor sure what this change does?

apis/nucleus/src/object/lo-handler.js
johanlahti: Yes, I will fix.

commands/serve/web/components/property-panel/DataCube.jsx
johanlahti: It adds the dimension to the external dimensions array (`dimensions`) so that the Filter pane can handle multiple dimensions. This `dimensions` array is then sent in to the handler so it can return it whenever the property panel calls `getDImensions()`.

And it's chained so that it only does this if `setProperties` is successful.

apis/nucleus/src/object/lo-handler.js
johanlahti: For the Listbox case we can always assume 1 dim. But we still align to the API where maxDims has to return a value. Therefore, one could argue either way here too but let's be explicit and set it to `0` 😄

commands/serve/web/components/property-panel/DataCube.jsx
johanlahti: We may want to do the same (wait for success of `setProperties`) also when a dimension is removed, further down in this file.

Caele (approved):

Caele:
Need to validate this a bit with Slider/Date picker before release, but else it looks good.

PR #974, opened 2022-10-27
ci: misleading error message

DanielS-Qlik:
## Motivation

When you build with `nebula build` and your peer dependency typescript is missing this error message is shown:
`Please install '@rollup/plugin-typescript' to build using typescript.` even though it's installed.
With this change you get the actual error first which is `Cannot find module 'typescript'`

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`
    - [ ] No changes
     ***OR***

### DIFF
--- a/commands/build/lib/build.js
+++ b/commands/build/lib/build.js
@@ -110,7 +110,7 @@ const config = ({
     try {
       typescriptPlugin = require('@rollup/plugin-typescript'); // eslint-disable-line
     } catch (e) {
-      throw new Error(`Please install '@rollup/plugin-typescript' to build using typescript.`);
+      throw new Error(`${e}\n '@rollup/plugin-typescript' is required to build using typescript.`);
     }
   }
 


commands/build/lib/build.js
```diff
@@ -110,7 +110,7 @@ const config = ({
     try {
       typescriptPlugin = require('@rollup/plugin-typescript'); // eslint-disable-line
     } catch (e) {
-      throw new Error(`Please install '@rollup/plugin-typescript' to build using typescript.`);
+      throw new Error(`${e}\n '@rollup/plugin-typescript' is required to build using typescript.`);
```
johanlahti, line 113: Is `e` always a string? But not a big deal even if it's not as it will become `[object Object]…` in that case. 😄

johanlahti (approved):

veinfors (approved):

PR #977, opened 2022-10-28
refactor(listbox): confirm or cancel selections on outside click

johanlahti:
## Motivation

Clicking outside a Listbox to confirm selections is already supported through the option `shouldConfirmOnBlur`. This PR removes that option and instead binds it to the nebula selections API *only*. Therefore, the identify function was broken out so it can be called for this purpose.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`
    - [ ] No changes
     ***OR***


Caele:
Is this a functional change for the mashup listbox?

johanlahti:
It's the same functionality as before, just rewritten and with unit tests added.
Also, there was a hidden option for toggling this behaviour (exposed through `DO_NOT_USE`). Now this is instead derived from the existence of an external API. We do not confirm/cancel selections on click outside when there is an external API since e.g. sense-client's API does this on it's own.

Caele:
> It's the same functionality as before, just rewritten and with unit tests added. Also, there was a hidden option for toggling this behaviour (exposed through `DO_NOT_USE`). Now this is instead derived from the existence of an external API. We do not confirm/cancel selections on click outside when there is an external API since e.g. sense-client's API does this on it's own.

But the `shouldConfirmOnBlur` was off by default, so anyone embedding a listbox would not get that behaviour? But it will be on by default now? or?

veinfors:
> > It's the same functionality as before, just rewritten and with unit tests added. Also, there was a hidden option for toggling this behaviour (exposed through `DO_NOT_USE`). Now this is instead derived from the existence of an external API. We do not confirm/cancel selections on click outside when there is an external API since e.g. sense-client's API does this on it's own.
> 
> But the `shouldConfirmOnBlur` was off by default, so anyone embedding a listbox would not get that behaviour? But it will be on by default now? or?

That's a good point - this will be a change in the default behaviour of the listbox. I was thinking it's always the wanted behaviour, but maybe it's not?

@Caele would you prefer to have this controlled through an option instead? And what behaviour makes most sense to have as default?

johanlahti:
Good point. Yeah, it's not as clear as I thought.
One could argue that selection modal state should always deactivate when clicking outside a Listbox. The external API should then adapt to this. But that too would be a change in default behaviour.

johanlahti:
While it's a change in default behaviour I think this has *always* been the default behaviour in Qlik Sense. So enforcing this behaviour is in a way not worse than enforcing a pre-defined selections API.

Caele:
I think the behaviour makes sense yes, but we need to test properly to make sure it doesn't have any other implications.

johanlahti:
I have verified this branch in sense-client with Direct Query and Analytic Creation as well as with a dev-mashup and I do not see anything strange when enabling this.

apis/nucleus/src/components/listbox/assets/identify.js
```diff
@@ -0,0 +1,9 @@
+export default function identify({ qId, options }) {
+  return {
+    // External or internal ("on the fly") session model.
+    hasExternalSessionModel: !!(qId || options.sessionModel),
```
veinfors, line 4: I think it's a little bit confusing/misleading calling it `hasExternalSessionModel`, since it's not an external session model if a qId is specified.

Maybe it's better to call it `isOnTheFlyModel`?

Caele: Both are a bit confusing, what does on the fly model mean?

johanlahti: It's the opposite of an `onTheGroundModel` 😄

veinfors: This is what I'm thinking of when I say "on the fly" in this case: https://qlik.dev/libraries-and-tools/nebulajs/rendering#render-on-the-fly

That is the opposite of an existing object you could say...

But there might be some better naming for it.

apis/nucleus/src/components/listbox/__tests__/list-box-inline.spec.jsx
```diff
@@ -113,6 +112,7 @@ describe('<ListboxInline />', () => {
       search: true,
       focusSearch: false,
       toolbar: true,
+      element: 'element',
```
veinfors, line 115: I don't see this element being used...

apis/nucleus/src/components/listbox/assets/identify.js
johanlahti: Why not just hasExistingSessionModel – i.e. `existing` instead of `external`. Would that be more correct?

veinfors: To me, there are two scenarios. When there is an object that already exist in engine, or not.
When passing in either a `qId` or an existing model (`options.sessionModel`), the object exist, otherwise it should be created/generated.

I don't think we need to include sessionModel in the naming, the relevant thing is to indicate if it's an existing object or an object should be created/generated. Therefore, I would rather call it just `existingObject` or something like that.

johanlahti: Sounds good 👍

veinfors (approved):

PR #1010, opened 2022-11-17
feat: add dense as a property

DanielS-Qlik:
## Motivation

To be able to set the dense mode from the property panel in sense client, a dense property is added.
This way you also avoid re rendering the Listbox when toggling dense mode, compared to passing dense as an option.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
+++ b/apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
@@ -6,6 +6,16 @@ export default function useExistingModel({ app, qId, options = {} }) {
   const [modelStore] = useModelStore();
   const { sessionModel } = options;
 
+  const validateOptions = (opts) => {
+    if (opts.dense) {
+      throw new Error('Option "dense" is not avaliable when rendering existing model.');
+    }
+  };
+
+  useEffect(() => {
+    validateOptions(options);
+  }, []);
+
   useEffect(() => {
     async function fetchObject() {
       const modelFromStore = modelStore.get(qId);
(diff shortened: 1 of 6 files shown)


apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
```diff
@@ -6,6 +6,16 @@ export default function useExistingModel({ app, qId, options = {} }) {
...
 
+  const validateOptions = (opts) => {
+    if (opts.dense) {
+      throw new Error('Option "dense" is not avaliable when rendering existing model.');
+    }
+  };
+
+  useEffect(() => {
+    validateOptions(options);
+  }, []);
```
johanlahti, line 17: Can we just skip the `useEffect`?

johanlahti: We should probably iterate through a list of blacklisted options and do the same as you do here. But can wait with that.

apis/stardust/api-spec/spec.json
```diff
@@ -917,7 +917,7 @@
                   "type": "boolean"
                 },
                 "dense": {
-                  "description": "Reduces padding and text size",
+                  "description": "Reduces padding and text size. (Not applicable when rendering existing model)",
```
johanlahti, line 920:
```suggestion
                  "description": "Reduces padding and text size (not applicable when rendering existing model).",
```

apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
DanielS-Qlik: Yes, there are more to come. But Im thinking its a part of the property-panel work.

johanlahti (approved):

Caele (approved):

PR #1013, opened 2022-11-28
feat: sheet embed support

Caele:
## Motivation

Adds support for sense sheets to Nebula
Also fixes/adds:
- Expose a viz object model on Viz
- Fix support for "none" color in theme palettes
- Sets api gov to support pre-releases

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/nucleus/src/components/Sheet.jsx
```diff
@@ -0,0 +1,103 @@
...
+    - sheet title + bg + logo etc + as option
+    - sheet exposed classnames for theming
+  */
+
+  if (cells && cells.length) {
+    const height = Number.isNaN(layout.height) ? '100%' : `${Number(layout.height)}%`;
+    const promises = cells.map((c) => c.mountedPromise);
+    Promise.all(promises).then(() => {
+      onMount();
+    });
```
T-Wizard, line 81: should there be safegaurds to only run onMout once?

apis/nucleus/src/components/Sheet.jsx
```diff
@@ -0,0 +1,103 @@
...
+          let mounted;
+          const mountedPromise = new Promise((resolve) => {
+            mounted = resolve;
+          });
+          return {
+            model: vs[i].model,
+            id: c.name,
+            bounds: c.bounds,
+            cellRef: React.createRef(),
+            currentId: uid(),
```
T-Wizard, line 51: It feels wrong to create a new currentId each time the sheet layout changes. (but maybe it is OK)

Caele: yea I agree, need to think about how to handle the lifecycle of the sheet better.

PR #1021, opened 2022-12-05
feat: enable frequency count property

CostInCode:
For testing, pick also the change in list-objects at https://github.com/qlik-oss/sn-list-objects/pull/30

https://user-images.githubusercontent.com/13185716/205690587-29985edf-88c2-4018-80a3-f14e1a80151a.mov

### DIFF
--- a/apis/nucleus/src/index.js
+++ b/apis/nucleus/src/index.js
@@ -416,7 +416,7 @@ function nuked(configuration = {}) {
            * @param {string=} options.title Custom title, defaults to fieldname
            * @param {Direction=} [options.direction=ltr] Direction setting ltr|rtl.
            * @param {ListLayout=} [options.listLayout=vertical] Layout direction vertical|horizontal
-           * @param {FrequencyMode=} [options.frequencyMode=N] Show frequency none|value|percent|relative
+           * @param {FrequencyMode=} [options.frequencyMode=none] Show frequency none|value|percent|relative
            * @param {boolean=} [options.histogram=false] Show histogram bar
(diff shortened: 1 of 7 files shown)


apis/nucleus/src/index.js
```diff
@@ -416,7 +416,7 @@ function nuked(configuration = {}) {
            * @param {string=} options.title Custom title, defaults to fieldname
            * @param {Direction=} [options.direction=ltr] Direction setting ltr|rtl.
            * @param {ListLayout=} [options.listLayout=vertical] Layout direction vertical|horizontal
-           * @param {FrequencyMode=} [options.frequencyMode=none] Show frequency none|value|percent|relative
+           * @param {FrequencyMode=} [options.frequencyMode=N] Show frequency none|value|percent|relative
```
veinfors, line 419: This is a little bit confusing, because the available values for `FrequencyMode` are:
none, value, percent and relate according to the spec:
```
"FrequencyMode": {
      "kind": "alias",
      "items": {
        "kind": "union",
        "items": [
          {
            "kind": "literal",
            "value": "'none'"
          },
          {
            "kind": "literal",
            "value": "'value'"
          },
          {
            "kind": "literal",
            "value": "'percent'"
          },
          {
            "kind": "literal",
            "value": "'relative'"
          }
        ]
      }
    },
```

veinfors: So I think you can just keep this as `none`... Sure, the corresponding property might be `N`, but the option is still `none` the way I see it.

johanlahti (approved):

PR #1028, opened 2022-12-13
feat: add checkboxes, histogram as properties

DanielS-Qlik:
## Motivation
To be able to set the checkboxes and histogram from the property panel in sense client.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
+++ b/apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
@@ -11,6 +11,9 @@ export default function useExistingModel({ app, qId, options = {} }) {
     frequencyMode: options.frequencyMode,
     checkboxes: options.checkboxes,
     histogram: options.histogram,
+    title: options.title,
+    stateName: options.stateName,
+    listLayout: options.listLayout,
   };
 
   let usedInvalidOption = null;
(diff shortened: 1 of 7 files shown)


apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
```diff
@@ -6,12 +6,25 @@ export default function useExistingModel({ app, qId, options = {} }) {
...
+    stateName: options.stateName,
+    listLayout: options.listLayout,
+  };
+
+  let usedInvalidOption = null;
+  Object.entries(invalidOptions).forEach(([key, value]) => {
+    if (value) {
+      usedInvalidOption = key;
+    }
+  });
```
johanlahti, line 24: So it will only consider the options invalid if its value is `truthy`? An option with value `false` will pass as ok.

I think something like this could work:
```
const forbiddenOptions = ['dense', 'frequencyMode', 'checkboxes'…];
const usedOptions = Object.keys(options);

// get intersection of arrs
const invalidOptions = usedOptions.filter(Set.prototype.has, new Set(forbiddenOptions));

throw new Error(`Options ${invalidOptions.join(', ')} are invalid…`);

```

veinfors: `usedOptions.filter(Set.prototype.has, new Set(forbiddenOptions));` might be an elegant solution, but it's not very easy to read and understand IMO...

johanlahti: Yes, it is good to wrap that in a self-explanatory variable. Could even add a comment about it.

johanlahti (approved):

Issue #1025, opened 2022-12-07
Listbox: Show a disclaimer when there are no search matches


johanlahti:
## 🚀 Feature

Show a disclaimer in the Listbox when there are no search matches.
It should appear like in the old listbox (i.e. also in popovers). See below:

![image](https://user-images.githubusercontent.com/5780544/206198189-8168784d-8690-4503-87ac-683f2cea4eea.png)

### Motivation

Feature parity. https://jira.qlikdev.com/browse/QB-16970

CostInCode:
The linked PR above implements the feature. See the screens below for both sense client and Nebula mashup.

<img width="601" alt="Screenshot 2023-03-10 at 09 36 12" src="https://user-images.githubusercontent.com/13185716/224266357-3aecf213-e1df-40c8-8e35-48a4f6783393.png">

<img width="354" alt="Screenshot 2023-03-10 at 09 45 11" src="https://user-images.githubusercontent.com/13185716/224268244-e8075f8c-a204-4c5e-a026-d8943fb2c8ee.png">

PR #1030, opened 2022-12-15
feat: add disclaimer listbox search

CostInCode:
Show a disclaimer in Listbox when search has no results

https://github.com/qlik-oss/nebula.js/issues/1025

Unit test will be updated/added in an another PR

https://user-images.githubusercontent.com/13185716/208091034-3eb0a85a-cd4b-463b-ae33-d1b5ee81636e.mov

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBox.jsx
+++ b/apis/nucleus/src/components/listbox/ListBox.jsx
@@ -67,7 +67,7 @@ function getSizeInfo({ isVertical, checkboxes, dense, height }) {
 function SearchMessage(width) {
   const styleSearchMsg = {
     fontSize: '16px',
-    padding: '16px',
+    padding: '16px 0',
     textAlign: 'center',
     minWidth: width,
   };
--- a/apis/nucleus/src/components/listbox/ListBox.jsx
+++ b/apis/nucleus/src/components/listbox/ListBox.jsx
@@ -12,6 +12,7 @@ import useLayout from '../../hooks/useLayout';
 import useSelectionsInteractions from './hooks/selections/useSelectionsInteractions';
 
 import RowColumn from './components/ListBoxRowColumn';
+import ListBoxDisclaimer from './components/ListBoxDisclaimer';
 
 const PREFIX = 'ListBox';
 const scrollBarThumb = '#BBB';
@@ -62,17 +63,6 @@ function getSizeInfo({ isVertical, checkboxes, dense, height }) {
     itemSize,
     listHeight,
   };
-}
-
-function SearchMessage(width) {
-  const styleSearchMsg = {
-    fontSize: '16px',
-    padding: '16px 0',
-    textAlign: 'center',
-    minWidth: width,
-  };
-
-  return <div style={styleSearchMsg}>There are no matches for your search.</div>;
 }
 
 export default function ListBox({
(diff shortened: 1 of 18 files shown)


apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -64,6 +64,17 @@ function getSizeInfo({ isVertical, checkboxes, dense, height }) {
...
+  const styleSearchMsg = {
+    fontSize: '16px',
+    padding: '16px 0',
+    textAlign: 'center',
+    minWidth: width,
+  };
+
+  return <div style={styleSearchMsg}>There are no matches for your search.</div>;
+}
+
```
johanlahti, line 77: Move out in to a new file (`SearchDisclaimer.jsx` or something) and import it from ListBox.jsx. It will make the upcoming merge conflict easier to solve 😄

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -64,6 +64,17 @@ function getSizeInfo({ isVertical, checkboxes, dense, height }) {
...
 
+function SearchMessage(width) {
+  const styleSearchMsg = {
+    fontSize: '16px',
+    padding: '16px 0',
+    textAlign: 'center',
+    minWidth: width,
+  };
+
+  return <div style={styleSearchMsg}>There are no matches for your search.</div>;
```
DanielS-Qlik, line 75: Wrap the string in `{translator.get('There are no matches for your search.')}` like in ListboxInline.jsx, and maybe you can  reuse the translation string from the old listbox?

CostInCode: I'll do this way, looks like the translation from the old listbox ('Listbox.NoMatchesForYourTerms') will not work

veinfors:
The text doesn't seem to be centered horizontally in the listbox, that needs to be fixed.

ashrles (approved):
Translation LGTM

apis/nucleus/src/components/listbox/__tests__/list-box.test.jsx
```diff
@@ -253,7 +256,7 @@ describe('<Listbox />', () => {
     test('should prevent InfiniteLoader to get itemCount == 0', async () => {
       layout.qListObject.qSize.qcy = 0;
       await render();
-      expect(infiniteProps.itemCount).toBe(1);
+      expect(infiniteProps.itemCount).toBe(2);
```
johanlahti, line 259: This happens because it's testing the previous test case: It never renders the `<InfiniteLoader />` since we are showing a disclaimer instead. Since we are showing a text "No matches for your terms." we need to verify that this only happens when we:
1. Search for something
2. We get no matches

Are there no other cases where we might end up with `listCount === 0`?

CostInCode:
> The text doesn't seem to be centered horizontally in the listbox, that needs to be fixed.

This should be fixed now (the previous videoscreen was not updated with the fix) by https://github.com/qlik-oss/nebula.js/pull/1030/commits/8f7877707ea7d7601816a8cffc2e1f5dbcfc1452

apis/nucleus/src/components/listbox/__tests__/list-box.test.jsx
johanlahti: I cannot see any problems with this right now. I will merge this.

PR #1033, opened 2022-12-16
chore: add support for engine ARGB colors

Caele:
## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/utils/__tests__/background-props.test.js
+++ b/apis/nucleus/src/utils/__tests__/background-props.test.js
@@ -18,7 +18,7 @@ describe('Background property resolver', () => {
     bgCompLayout = {
       key: 'general',
       bgColor: {
-        useColorExpression: true,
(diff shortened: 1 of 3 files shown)


T-Wizard:
why?
It looks like the color is an expression that would work is engine evaluaded it but it was not evaluated by engine.

Caele:
> why? It looks like the color is an expression that would work is engine evaluaded it but it was not evaluated by engine.

Given the expression =argb(100,200,Sum(sales),255) engine will calculate a dual:
qNum: 29875577 // numeric rgba representation
qText: "ARGB(100,200,123,255)"
So when we are using a qStringExpression we only get the qText back. Most other applications of color by expression is on the hyperCube, and then we get both qNum and qText and tend to use qNum.

T-Wizard (approved):

PR #1058, opened 2023-01-18
chore: update templates to use inline ext

Caele:
## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


T-Wizard:
with inline ext you can/should also remove the `--ext src/ext.js` parameter in the sense commant in the template package.json
here: https://github.com/qlik-oss/nebula.js/blob/6695ee87314e095a040da7132f528f01a91d5e28/commands/create/templates/sn/none/_package.json#L21

veinfors (approved):

T-Wizard (approved):

PR #1059, opened 2023-01-18
feat(Listbox): overflow disclaimer

DanielS-Qlik:
## Motivation

Since react-window only can display a fixed amount of items we have to inform the user about that.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

Grid-column:
grid-column4

Grid-row:
grid-row4

Single column:
single-column4

Dense:
dense


apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -64,6 +65,13 @@ export default function ListBox({
   const { setStoreValue } = useDataStore(model);
   const loadMoreItems = useCallback(itemsLoader.loadMoreItems, [layout]);
 
+  const overflowDisclaimerReducer = (state, action) => ({ ...state, ...action });
+  const [overflowDisclaimer, setOverflowDisclaimer] = useReducer(overflowDisclaimerReducer, {
+    show: false,
+    dismissed: false,
+  });
+  const dismissOverflowDisclaimer = () => setOverflowDisclaimer({ dismissed: true });
```
johanlahti, line 73: Instead of a reducer (which by convention should *dispatch* an action `type` instead of a setter) you could simply use a useState with a function:
```
const [overflowDisclaimer, setOverflowDisclaimer] = useState({
    show: false,
    dismissed: false,
});
const dismissOverflowDisclaimer = () => setOverflowDisclaimer((state) => ({ ...state, dismissed: true }));
```

Otherwise, I think using reducers is a nice idea for more complex state management. I saw this guide: https://beta.reactjs.org/learn/extracting-state-logic-into-a-reducer

apis/nucleus/src/components/listbox/assets/get-list-sizes.js
```diff
@@ -41,6 +41,16 @@ export default function getListSizes({ layout, width, height, listCount, count,
     }
   }
 
+  columnCount = dataLayout === 'singleColumn' ? 1 : columnCount;
+  rowCount = dataLayout === 'singleColumn' ? count : rowCount;
+  const maxRowCount = layoutOptions.dense ? 838000 : 577000; // Styling breaks on items above this number: https://github.com/bvaughn/react-window/issues/659
+  rowCount = Math.min(rowCount, maxRowCount);
+  const maxColumnCount = Math.floor(33550000 / columnWidth);
+  columnCount = Math.min(columnCount, maxColumnCount);
```
johanlahti, line 49: Should we add a comment about the max columns as well? Or where does that number come from? (A lot of columns though 😮)

apis/nucleus/src/components/listbox/components/grid-list-components/setOverflowDisclaimer.js
```diff
@@ -0,0 +1,26 @@
...
+}) {
+  const isColumnLayout = layoutOptions?.layoutOrder === 'column' && layoutOptions.dataLayout !== 'singleColumn';
+  const index = getIndex(renderProps, isColumnLayout);
+  const stopIndex = isColumnLayout ? maxCount.column : maxCount.row;
+  const count = isColumnLayout ? columnCount : rowCount;
+  const overflowPossible = count >= stopIndex;
+  if (!overflowPossible && overflowDisclaimer.state.show) {
+    overflowDisclaimer.set({ show: false });
+  } else if (index >= stopIndex - 1) {
+    overflowDisclaimer.set({ show: true });
```
johanlahti, line 24:
```suggestion
const show = overflowPossible || overflowDisclaimer.state.show || index >= stopIndex - 1;
overflowDisclaimer.set({ show });
```
You may have to check and modify this suggestion though 😄

apis/nucleus/src/components/listbox/assets/get-list-sizes.js
DanielS-Qlik: I named that number in a const, to clarify what is is. Also added a comment to the limitation bug.

apis/nucleus/src/components/listbox/components/grid-list-components/setOverflowDisclaimer.js
DanielS-Qlik: 👌

johanlahti (approved):

PR #1067, opened 2023-01-27
feat: listbox popover api

Caele:
## Motivation

Adds an experimental listbox popover API under DO_NOT_USE (for now). Also fixes some issues with React portals and missing keys,

## Requirements checklist

- [x] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxPopoverWrapper.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxPopoverWrapper.jsx
@@ -1,7 +1,15 @@
 import React, { useState } from 'react';
 import ListBoxPopover from './ListBoxPopover';
 
-export default function ListBoxPopoverWrapper({ app, fieldIdentifier, stateName, element, options = {} }) {
+export default function ListBoxPopoverWrapper({
+  app,
+  fieldIdentifier,
+  stateName,
+  element,
+  anchorOrigin,
+  transformOrigin,
+  options = {},
+}) {
   const [showState, setShowstate] = useState(!!options.show);
   const handleCloseShowState = () => {
     setShowstate(false);
@@ -12,6 +20,8 @@ export default function ListBoxPopoverWrapper({ app, fieldIdentifier, stateName,
       show={showState}
       app={app}
       alignTo={{ current: element }}
+      anchorOrigin={anchorOrigin}
+      transformOrigin={transformOrigin}
       close={handleCloseShowState}
       fieldName={fieldIdentifier}
       stateName={stateName}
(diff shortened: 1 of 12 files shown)


apis/nucleus/src/components/listbox/ListBoxPopoverWrapper.jsx
```diff
@@ -0,0 +1,30 @@
...
+  const [showState, setShowstate] = useState(!!options.show);
+  const handleCloseShowState = () => {
+    setShowstate(false);
+  };
+
+  return (
+    <ListBoxPopover
+      show={showState}
+      app={app}
+      alignTo={{ current: element }}
```
haxxmaxx, line 22: Is this some ref trick? why not just pass `element` as it is?

apis/nucleus/src/index.js
```diff
@@ -482,6 +485,23 @@ function nuked(configuration = {}) {
       getRegisteredTypes: types.getList,
       __DO_NOT_USE__: {
         types,
+        popover(anchorElement, fieldIdentifier, options = { show: true }) {
+          if (api._popoverInstance) {
+            root.remove(api._popoverInstance);
+            api._popoverInstance = null;
+          }
```
haxxmaxx, line 492: Son now when there is an existing pop-over instance, we remove it and create a new one. I thought we wanted to keep the pop-over instance, and replace the contents (i.e. a different field and position).

What is the difference between this and just mounting and unmounting the popover everytime it opens and closes?

apis/nucleus/src/components/listbox/ListBoxPopoverWrapper.jsx
a-m-dev: reason is `<ListboxPopover />` component has been used in nebula itself. I would assume that in internal use case of nebula , for `alignTo` prop, then entire ref object has been passed so we are already using it as `alignTo.current` inside of `<ListboxPopover />`component.


**on the other hand:** when we expose popover function -> we expect to receive an actual element as an anchor point, and not a ref (because not every environment has to be react to have a ref object, it could be mashup for example ). 

so we accept the overhead of wrapping actual element into ref format for internal usage like `{ current: element }`
(or pls correct me if I'm wrong in this case)

Caele: looks like a ref trick yes, but I think I just copied this from somewhere... but now I can't find it..

apis/nucleus/src/index.js
Caele: yes, I wanted to keep it, but ran into all sorts of issues. So made it this way for now, with the intention of revisiting. Bascially the API should be stable, but we ca change the implementation

veinfors (approved):

haxxmaxx (approved):

PR #1079, opened 2023-02-14
feat: create api

Caele:
## Motivation

Adds new APIs to better support creating new charts outside of Nebula:
- embed.create: Allows creating a new engine object OR getting the property structure for one
- ~useEmitter: Allows a chart to communicate with the outside~: now in separate PR: #1168 

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/object/create-object.js
+++ b/apis/nucleus/src/object/create-object.js
@@ -1,0 +1,58 @@
+import populateData from './populator';
+import { modelStore } from '../stores/model-store';
+
+/**
+ * @typedef {string | EngineAPI.INxDimension | EngineAPI.INxMeasure | LibraryField} Field
+ */
+
+/**
+ * @interface CreateConfig
+ * @description Rendering configuration for creating and rendering a new object
+ * @property {string} type
+ * @property {string=} version
+ * @property {(Field[])=} fields
+ * @property {EngineAPI.IGenericObjectProperties=} properties
+ */
+export default async function createObject(
+  { type, version, fields, properties, extendProperties /* , options, plugins, element */ },
+  halo,
+  generateOnly
+) {
+  let mergedProps = {};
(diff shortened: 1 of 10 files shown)


apis/nucleus/src/object/create-object.js
```diff
@@ -0,0 +1,58 @@
...
+          sn,
+          properties: mergedProps,
+          fields,
+        },
+        halo
+      );
+    }
+    if (properties && sn && sn.qae.properties.onChange) {
+      sn.qae.properties.onChange.call({}, mergedProps);
+    }
```
niekvanstaveren, line 39: What is this?

niekvanstaveren:
Would probably be better to split up the create api and the emitter work into two separate PRs.
Also not completely sure if we should expose the create part in one API function. Since the returned object (model vs. property definition) is completely different it might be better to split them up

apis/nucleus/src/object/create-object.js
Caele: Good question, it runs a onChange callback that you can put on you qae def. I think it is only used in a single place: https://github.com/qlik-oss/sn-mekko-chart/blob/master/src/index.js#L47
That thing updates coloring props when switching dims/meas.

apis/nucleus/src/object/create-object.js
```diff
@@ -0,0 +1,59 @@
...
+  } catch (e) {
+    // error = e;
+    // minimal dummy object properties to allow it to be created
+    // and rendered with the error
+    mergedProps = {
+      qInfo: {
+        qType: type,
+      },
+      visualization: type,
+    };
```
T-Wizard, line 50: maybe not for generateOnly?

T-Wizard (approved):

PR #1081, opened 2023-02-15
feat: add property for disabling listbox search

CostInCode:
## Motivation
Add `searchEnabled` property so that the user is able to disable the search setting in listbox.

https://user-images.githubusercontent.com/13185716/219059306-343e30d6-6a7a-468f-a8ce-6aba13cb9e2c.mov

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
+++ b/apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
@@ -24,6 +24,7 @@ export default function ListBoxSearch({
   dense = false,
   visible = true,
   wildCardSearch = true,
+  searchEnabled,
 }) {
   const { translator } = useContext(InstanceContext);
   const [value, setValue] = useState('');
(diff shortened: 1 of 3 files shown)


apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -125,47 +126,49 @@ export default function ListBoxSearch({
...
-          border: 'none',
-          borderRadius: 0,
-          backgroundColor: 'transparent',
-          '& fieldset': {
-            border: `1px solid ${theme.palette.divider}`,
-            borderWidth: '1px 0 1px 0',
-            borderRadius: 0,
-          },
-          '&:hover': {
+    searchEnabled !== false && (
```
veinfors, line 129: Maybe cleaner to move this check into the if statement on line 124 instead?
like this:
```
if (!visible || searchEnabled === false) {
  return null;
}
```

johanlahti (approved):

PR #1082, opened 2023-02-20
docs: add load to config

Caele:
## Motivation

Documents the global load fallback function on config. 
Additionally moves some jsdoc around and adds examples to improve the docs.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/index.js
+++ b/apis/nucleus/src/index.js
@@ -48,9 +48,9 @@ const DEFAULT_SNAPSHOT_CONFIG = /** @lends SnapshotConfiguration */ {
 /**
  * @interface Configuration
  * @property {LoadType=} load Fallback load function for missing types
- * @property {Context=} context
- * @property {Array<TypeInfo>=} types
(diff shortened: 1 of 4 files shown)


apis/nucleus/src/index.js
```diff
@@ -67,33 +47,69 @@ const DEFAULT_SNAPSHOT_CONFIG = /** @lends SnapshotConfiguration */ {
 
 /**
  * @interface Configuration
+ * @property {LoadType=} load Fallback load function for missing types
```
T-Wizard, line 50: how does this use interact with getRegisteredTypes? (and should there be anything in the docs about it?)

T-Wizard (approved):

apis/nucleus/src/index.js
Caele: hmm, have to think about that. I think I added that to solve something with charts in tooltips, but not sure. For coming projects we will need it however, but in an expanded sense. So might be that we need another config function to make it fully compatible.

PR #1083, opened 2023-02-22
feat: listbox search icon button changes

CostInCode:
https://user-images.githubusercontent.com/13185716/220606265-20232a51-f9f2-446f-95e1-695c16f922bc.mov

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -209,10 +192,14 @@ export default function ListBoxInline({ options = {} }) {
           <Grid item>
             {isLocked ? (
               <IconButton tabIndex={-1} onClick={unlock} disabled={!isLocked} size="large">
-                <Lock title={translator.get('Listbox.Unlock')} />
+                <Lock title={translator.get('Listbox.Unlock')} style={iconStyle} />
```
johanlahti, line 195: Does it work to use: `sx={{ fontSize: '12px' }}` instead of `style`? If it does, use that instead.

johanlahti (approved):
For later: check if we can vertically align the search icon.

PR #1090, opened 2023-02-28
chore: move more popover back into container

Caele:
## Motivation

To get the look'n'feel back to previous iteration and tidy up the code.
Looks like this now:
image
image
image

 Previously:
image

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/ActionsToolbar.jsx
+++ b/apis/nucleus/src/components/ActionsToolbar.jsx
@@ -128,7 +128,12 @@ function ActionsToolbar({
     return buttons[last ? buttons.length - 1 : 0];
   };
 
-  useEffect(() => () => setShowMoreItems(false), [popover.show]);
+  useEffect(
+    () => () => {
+      setShowMoreItems(false);
+    },
+    [popover.show, show, selections.show]
+  );
 
   useEffect(() => {
     if (!focusHandler) return;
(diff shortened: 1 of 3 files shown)


apis/nucleus/src/components/ActionsToolbarMore.jsx
```diff
@@ -55,6 +55,11 @@ const More = React.forwardRef(
           ref={ref}
           open={show}
           anchorEl={alignTo.current}
+          getContentAnchorEl={null}
+          container={alignTo.current}
+          disablePortal
+          hideBackdrop
+          style={{ pointerEvents: 'none' }}
```
johanlahti, line 62: Instead of `style` it could send it in as a property and set through the `styled(Popover)` call.

apis/nucleus/src/components/ActionsToolbar.jsx
```diff
@@ -212,7 +213,7 @@ function ActionsToolbar({
         <More
           show={showMoreItems}
           actions={moreActions}
-          alignTo={moreRef}
+          alignTo={moreAlignTo}
```
johanlahti, line 216: Does it need a fallback value?

johanlahti (approved):

apis/nucleus/src/components/ActionsToolbar.jsx
Caele: hmm, maybe

a-m-dev (approved):

PR #1094, opened 2023-03-01
fix: correctly hide hidden fields in app selections

Caele:
## Motivation

Hides hidden fields in the app selections

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/selections/SelectedFields.jsx
+++ b/apis/nucleus/src/components/selections/SelectedFields.jsx
@@ -37,7 +37,9 @@ function getItems(layout) {
   if (layout.alternateStates) {
     layout.alternateStates.forEach((s) => collect(s.qSelectionObject, fields, s.stateName));
   }
-  return Object.keys(fields).map((key) => fields[key]);
+  return Object.keys(fields)
+    .map((key) => fields[key])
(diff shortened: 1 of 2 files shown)


apis/nucleus/src/components/selections/SelectedFields.jsx
```diff
@@ -37,7 +37,9 @@ function getItems(layout) {
   if (layout.alternateStates) {
     layout.alternateStates.forEach((s) => collect(s.qSelectionObject, fields, s.stateName));
   }
-  return Object.keys(fields).map((key) => fields[key]);
+  return Object.keys(fields)
+    .map((key) => fields[key])
+    .filter((f) => !f.selections.some((s) => s.qIsHidden));
```
T-Wizard, line 42: maybe just check qIsHidden on the first value in selection
(I think all of them always should have the same qIsHidden value)
```suggestion
    .filter((f) => !f.selections[0].qIsHidden);
```

T-Wizard (approved):

apis/nucleus/src/components/selections/SelectedFields.jsx
Caele: Using some here actually handles the case of an empty selections array as well, which I think can happen.

PR #1095, opened 2023-03-01
chore: remove all old doc files

Caele:
## Motivation

Remove all the old documentation that is no longer updated, instead just link to qlik.dev

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


LiKang6688 (approved):
👍  https://qlik.dev/libraries-and-tools/nebulajs should be the only doc portal to nebula.js

PR #1106, opened 2023-03-03
fix: do not show Search box on selection when there is no title

quanho:
To fix https://github.com/qlik-oss/sn-list-objects/issues/103 by checking selection state and layout.title.

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -159,7 +159,8 @@ export default function ListBoxInline({ options = {} }) {
     : [];
 
   const showTitle = true;
-  const shouldShowSearch = ((search === 'toggle' || (search === 'inSelection' && (!layout.title || !toolbar))) && showSearch);
+  const shouldShowSearch =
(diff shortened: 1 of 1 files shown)


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -161,7 +161,7 @@ export default function ListBoxInline({ options = {} }) {
   const showTitle = true;
 
   const searchVisible =
-    (search === true || ((search === 'toggle' || search === 'inSelection') && showSearch)) && !selectDisabled();
+    (search === true || ((search === 'toggle' || (search === 'inSelection' && !layout.title)) && showSearch)) && !selectDisabled();
```
DanielS-Qlik, line 164: The same goes for when toolbar is disabled. Something like
```suggestion
    (search === true || ((search === 'toggle' || (search === 'inSelection' && (!layout.title || !toolbar))) && showSearch)) && !selectDisabled();
```

quanho: Good to know! Updated!

DanielS-Qlik (approved):

PR #1108, opened 2023-03-06
fix: only select with main mouse button

T-Wizard:
(maybe also abort selection if multiple buttons are pressed?)

## Motivation

fix qlik-oss/sn-list-objects#68

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/components/ListBoxRowColumn/ListBoxRowColumn.jsx
+++ b/apis/nucleus/src/components/listbox/components/ListBoxRowColumn/ListBoxRowColumn.jsx
@@ -104,10 +104,6 @@ function RowColumn({ index, rowIndex, columnIndex, style, data }) {
 
   const preventContextMenu = useCallback(
     (event) => {
-      if (checkboxes) {
-        // Event will not propagate in the checkbox/radiobutton case
-        onClick(event);
-      }
       event.preventDefault();
     },
     [checkboxes]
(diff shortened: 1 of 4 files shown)


apis/nucleus/src/components/listbox/components/ListBoxRowColumn.jsx
```diff
@@ -379,10 +379,6 @@ function RowColumn({ index, rowIndex, columnIndex, style, data }) {
   );
 
   const preventContextMenu = (event) => {
-    if (checkboxes) {
-      // Event will not propagate in the checkbox/radiobutton case
-      onClick(event);
-    }
```
T-Wizard, line 385: @johanlahti do you remember why this call to onClick was added? (in #808)

T-Wizard:
Anyone that have any ide why the unit test fails?
I do work when I run it locally on my machine.

DanielS-Qlik (approved):

Caele:
@T-Wizard I've made changes to this in my react18 update as well. The main problem there was the usage of onClick for the checkboxes. Basically a label/input pair will fire double click events when you click one of them, which with the react update somehow manifested. I've changed it to use onChange instead which seems to work better.
Maybe this could wait until that update?

apis/nucleus/src/components/listbox/components/ListBoxRowColumn.jsx
johanlahti: Maybe it could be implemented differently instead of this workaround.

Caele: Looking into this now as well, got a bit messy with the React update

Caele (approved):

PR #1114, opened 2023-03-08
fix(Listbox): use fallback title

DanielS-Qlik:
## Motivation

When rendering a Listbox on the fly (`embed.field(…).mount()`) and no title is provided in options.
Set a default title.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/hooks/useOnTheFlyModel.jsx
+++ b/apis/nucleus/src/components/listbox/hooks/useOnTheFlyModel.jsx
@@ -6,12 +6,17 @@ export default function useOnTheFlyModel({ app, fieldIdentifier, stateName, opti
   const [fieldDef, setFieldDef] = useState('');
   const [isFetching, setIsFetching] = useState(true);
   const [model, setModel] = useState();
+  const [title, setTitle] = useState(options.title);
+  const setFallbackTitle = (fallbackTitle) => {
+    setTitle((currentTitle) => currentTitle ?? fallbackTitle);
+  };
 
   useEffect(() => {
     async function fetchMasterItem() {
       try {
         const dim = await app.getDimension(fieldIdentifier.qLibraryId);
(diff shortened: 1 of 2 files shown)


apis/nucleus/src/components/listbox/hooks/useOnTheFlyModel.jsx
```diff
@@ -6,12 +6,17 @@ export default function useOnTheFlyModel({ app, fieldIdentifier, stateName, opti
   const [fieldDef, setFieldDef] = useState('');
   const [isFetching, setIsFetching] = useState(true);
   const [model, setModel] = useState();
+  const [title, setTitle] = useState(options.title);
+  const setFallbackTitle = (fallbackTitle) => {
+    setTitle((currentTitle) => currentTitle ?? fallbackTitle);
+  };
```
T-Wizard, line 12: I expect this change to break the ability to update the title by updating options.title
(but I have not tested)

veinfors: hmm... do we even have the possibility to update options without re-rendering the listbox completely? How/what scenario should we do to test that?

Or maybe you have a suggestion on how we should re-write this to avoid that problem (that might only exist in theory at the moment...) @T-Wizard ?

T-Wizard: if you can not update the options then it is not a problem (or at lest not now)
and even if you can update the options it could be a limitation we can live with.

as for re-write suggestion, mabe something like this:

```javascript
const [fallbackTitle, setFallbackTitle] = useState();
const title = options.title ?? fallbackTitle;
```

T-Wizard (approved):

veinfors (approved):

PR #1115, opened 2023-03-08
fix: assign id to all session listboxes

Caele:
## Motivation

When mounting multiple listboxes in the same mashup and using the exact same definition, they will also share a model -> meaning they also share the modal selection state.
They should get unique IDs.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/hooks/__tests__/useOnTheFlyModel.test.jsx
+++ b/apis/nucleus/src/components/listbox/hooks/__tests__/useOnTheFlyModel.test.jsx
@@ -49,13 +49,22 @@ describe('useExistingModel', () => {
   });
 
   describe('from field name', () => {
-    test('should use provided frequencyMode', async () => {
+    test('should assign qID', async () => {
       const options = {
         frequencyMode: 'value',
       };
       const fieldIdentifier = 'Alpha';
       await render(useOnTheFlyModel, { app, fieldIdentifier, stateName: '$', options });
       expect(useSessionModel.mock.lastCall[0].qListObjectDef.qFrequencyMode).toBe('V');
+    });
+
+    test('should use provided frequencyMode', async () => {
+      const options = {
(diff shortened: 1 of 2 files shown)


apis/nucleus/src/components/listbox/hooks/__tests__/useOnTheFlyModel.test.jsx
```diff
@@ -49,7 +49,7 @@ describe('useExistingModel', () => {
   });
 
   describe('from field name', () => {
-    test('should use provided frequencyMode', async () => {
+    test('should assign qID', async () => {
```
veinfors, line 52:
```suggestion
   test('should use provided frequencyMode', async () => {
```

apis/nucleus/src/components/listbox/hooks/useOnTheFlyModel.jsx
```diff
@@ -61,6 +62,7 @@ export default function useOnTheFlyModel({ app, fieldIdentifier, stateName, opti
   const listdef = {
     qInfo: {
       qType: 'njsListbox',
+      qId: uid(),
```
T-Wizard, line 65: I think this is a bad ide
and that it will casue that useSessionModel will create a new model all the time

maybe use

```javascript
const [qId] = useState(() => uid());
```

or

```javascript
const id = useRef();
if (!id.current) { id.current = uid(); }
```

to get a more stable random id

Caele: agree, that is a bit stupid. It is also strange that the ListBoxWrapper is a React component at all as the input properties never change. If you mount again we create a new one, so all that hooking stuff with useBlaBlaModel could be vastly simplified. There seesm to be a tendency to write more hooks than actually needed.

T-Wizard (approved):

PR #1116, opened 2023-03-08
refactor(ListBox): break out components from ListBoxRowColumn

johanlahti:
## Motivation

As a first step to optimise the performance, I'm breaking out components from the ListBoxRowColumn component.
This will make it easier to proceed with the next steps.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


DanielS-Qlik:
Could we move the content from index.jsx to ListBoxRowColumn.jsx?
Then in index.jsx just have one line `export { ListBoxRowColumn } from './ListBoxRowColumn';`
It makes it easier navigating your tabs in your file editor.

PR #1117, opened 2023-03-09
fix: avoid re-render on unchanged-context

T-Wizard:
## Motivation

part of fix for performance problem in qlik-oss/sn-list-objects#105

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`
    - [ ] No changes **_OR_** API changes has been formally approved
- [ ] Unit/Component test coverage
- [ ] Correct PR title for the changes (fix, chore, feat)

When build and tests have passed:

- [ ] Add code reviewers, for example @qlik-oss/nebula-core

PR #1119, opened 2023-03-09
fix: no re-render of inline listbox from extra useLayout updates

T-Wizard:
## Motivation

the internal logic in useLayout (in useRpc) causes a lot of extra re-rendering of the component using it
(some of the updates are relevant if the request status in the second retrun value is used)

this isolates ListBoxInline and prevent the extra updates from re-rendering it

together with #1117 and #1118 this fixes most (but not all) of the performance problem in qlik-oss/sn-list-objects#105

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -44,7 +44,7 @@ const Title = styled(Typography)(({ theme }) => ({
   fontFamily: theme.listBox?.title?.main?.fontFamily,
 }));
 
-export default function ListBoxInline({ options = {} }) {
+function ListBoxInline({ options, layout }) {
   const {
     app,
     direction,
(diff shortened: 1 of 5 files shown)


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -292,6 +291,7 @@ export default function ListBoxInline({ options = {} }) {
             {({ height, width }) => (
               <ListBox
```
quanho, line 292: Should we apply React.memo(ListBox) instead of React.memo(ListBoxInline)

T-Wizard: maybe (or memo both)
but doing a working memo on ListVox would require more work as it has a lot of properties that would need to be stable

Caele (approved):

PR #1121, opened 2023-03-09
refactor: avoid setPreselected many times

quanho:
On click an item, useSelectionsInteractions is run too many times because of some state change.
This PR is to reduce partly this problem by checking the value before setting a state.

### DIFF
--- a/apis/nucleus/src/components/listbox/hooks/selections/useSelectionsInteractions.js
+++ b/apis/nucleus/src/components/listbox/hooks/selections/useSelectionsInteractions.js
@@ -79,7 +79,9 @@ export default function useSelectionsInteractions({
         return;
       }
       const elemNumber = +event.currentTarget.getAttribute('data-n');
-      setPreSelected([elemNumber]);
+      if (!(preSelected.length === 1 && preSelected[0] === elemNumber)) {
+        setPreSelected([elemNumber]);
+      }
       handleSingleSelectKey(event);
     },
     [selectingValues, selectDisabled]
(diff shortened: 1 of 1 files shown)


apis/nucleus/src/components/listbox/hooks/selections/useSelectionsInteractions.js
```diff
@@ -79,7 +79,9 @@ export default function useSelectionsInteractions({
         return;
       }
       const elemNumber = +event.currentTarget.getAttribute('data-n');
-      setPreSelected([elemNumber]);
+      if (!(preSelected.length === 1 && preSelected[0] === elemNumber)) {
+        setPreSelected([elemNumber]);
+      }
```
Caele, line 84: Please explain these a bit more, why does this check make a difference? onClick/onMouseDown should only trigger a single time per click.

quanho: If preSelected is already [elemNumber], setting setPreSelected([elemNumber]) will change the state of preSelected (a new array even the content is the same as the old one). And the state change will trigger a new rerendering.

quanho: The change here will be not needed if we refactor this file. We should check which states are necessary for triggering a new rerendering and keep them other states should be converted to a normal variable.

Caele: I think the problem is that onClick runs twice, which is due to how a checkbox works. I will fix this is the react 18 update as that for some reason causes big issues then.
We can likely skip this update and instead look at a bigger refactor instead.

PR #1123, opened 2023-03-10
feat(Listbox): detached toolbar

DanielS-Qlik:
## Motivation

Show the toolbar detached if there is not enough space inside title bar or the title bar is hidden.
Align icon colors with sense-client.
fixes: https://github.com/qlik-oss/sn-list-objects/issues/63

detached-toolbar

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -124,6 +127,7 @@ function ListBoxInline({ options, layout }) {
   if (layout?.toolbar !== undefined) {
     toolbar = layout.toolbar;
   }
+  const toolbarDetachedOnly = toolbar && layout?.title === '';
```
johanlahti, line 130: Don't if title can be undefined… but:
```suggestion
  const toolbarDetachedOnly = toolbar && !layout?.title?.length
```

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -194,6 +197,16 @@ function ListBoxInline({ options, layout }) {
...
+  const getActionToolbarProps = (isPopover) =>
+    listboxGetActionProps({
+      isPopover,
+      showToolbar,
+      containerRef,
+      isLocked,
+      listboxSelectionToolbarItems,
+      selections,
+    });
+
```
johanlahti, line 209: Move this function outside of the `ListBoxInline` component to prevent cluttering and easier optimisation… if relevant.

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -203,145 +216,135 @@ function ListBoxInline({ options, layout }) {
...
               </Grid>
-            )}
-            <Grid item className={classes.listBoxHeader}>
-              {showTitle && (
-                <Title variant="h6" noWrap>
-                  {layout.title}
-                </Title>
-              )}
+            </Grid>
+            <Grid item xs />
```
johanlahti, line 279: What is this empty item doing here?

apis/nucleus/src/components/listbox/ListBoxInline.jsx
johanlahti: It may even be worth considering wrapping the `<ActionsToolbar />` inside another component that hosts this logic 🤔

apis/nucleus/src/components/listbox/interactions/listbox-get-action-props.js
```diff
@@ -0,0 +1,28 @@
+export default ({ isPopover, showToolbar, containerRef, isLocked, listboxSelectionToolbarItems, selections }) => ({
```
johanlahti, line 1: Named default export?
```suggestion
export default function getListboxActionProps({ isPopover, showToolbar, containerRef, isLocked, listboxSelectionToolbarItems, selections }) {
```

packages/ui/theme/definitions/light.js
```diff
@@ -18,7 +18,7 @@ const light = {
       disabled: 'rgba(0, 0, 0, 0.3)',
     },
     action: {
-      active: colors.grey25, // color for actionable things like icon buttons
+      active: '#545454', // color for actionable things like icon buttons
```
johanlahti, line 21: Was there no predefined color for this?

apis/nucleus/src/components/listbox/ListBoxInline.jsx
DanielS-Qlik: I don't either, but it should only be detached if you explicitly set it to an empty string.

apis/nucleus/src/components/listbox/ListBoxInline.jsx
DanielS-Qlik: Seems to have been there since this file was created, if removed the align of search glass is broken.

johanlahti: 😮 🆗

apis/nucleus/src/components/listbox/ListBoxInline.jsx
johanlahti: Ok, can just move the func outside/to its own file or outside of the component (I recently cleaned another file which had a lot of inline funcs like these 😬).

apis/nucleus/src/components/listbox/ListBoxInline.jsx
johanlahti: I was just thinking of the case when an undefined value slips in here, should this be `true` or `false` then?

apis/nucleus/src/components/ActionsToolbar.jsx
```diff
@@ -176,6 +176,10 @@ function ActionsToolbar({
     action: () => setShowMoreItems(!showMoreItems),
   };
 
+  const handleMouseDown = (e) => {
+    e.stopPropagation(); // Prevent clickOutSide handler confirming selection
+  };
+
```
Caele, line 182: This one is a bit strange? This should affect all charts as well and seems like a rather big deal?

veinfors: I also feel a little bit sceptic about this change. I did test it running with the sn-scatter-plot on a mashup and clicking outside the selection toolbar still confirms the selections (which it should do). So this confuses me a little bit, could you explain some more around this @DanielS-Qlik?

Caele: Should add `'.njs-action-toolbar-popover'` to the included elements list here: https://github.com/qlik-oss/nebula.js/blob/master/apis/nucleus/src/components/listbox/ListBoxPortal.jsx#L76
In that call to useObjectSelections.

DanielS-Qlik: I did try adding `'.njs-action-toolbar-popover'` at first, but the `event.target` here: https://github.com/DanielS-Qlik/nebula.js/blob/a420ff2253fc38557b5ea46e02a7281c27b6bfa9/apis/nucleus/src/hooks/useObjectSelections.js#L37 was an svg, so the `contains()` would return false. Now it works properly though, and I can not reproduce the previous problem. So I changed as you suggested

DanielS-Qlik: It looks like I accidentally removed `pointerEvents: 'auto'` here: https://github.com/DanielS-Qlik/nebula.js/blob/a420ff2253fc38557b5ea46e02a7281c27b6bfa9/apis/nucleus/src/components/ActionsToolbar.jsx#L244 when styling, that's why adding `'.njs-action-toolbar-popover'` in the call to useObjectSelection didn't work at first 😬 .

apis/nucleus/src/components/ActionsToolbar.jsx
```diff
@@ -110,6 +106,10 @@ function ActionsToolbar({
   },
   focusHandler = null,
   actionsRefMock = null, // for testing
+  popoverAnchorOrigin = {
+    vertical: 'top',
+    horizontal: 'right',
+  },
```
Caele, line 112: Why move this code?
And why the theme spacing change below it?
This file is used by all charts, so rather not make changes unless there is a good reason.

DanielS-Qlik: Its moved to be able to override the defaults from ListboxInline.
This is to align with placement and looks in sense-client.

DanielS-Qlik: Can we detect if we are in sense-client, and only apply these changes then?

Caele: Where are you doing this override? and what is the diff between client and outside?

veinfors (approved):

PR #1124, opened 2023-03-10
fix: fix filter pane header change the height on selection

quanho:
In a filter pane, the action toolbar items (to the right of the header) has 32px height, while the title has 28px height.
This causes the problem reported here https://github.com/qlik-oss/sn-list-objects/issues/119.
To avoid the header height being changed on selection, we set its height to the same height of the action toolbar items.
This PR is to fix https://github.com/qlik-oss/sn-list-objects/issues/119.

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -209,7 +209,7 @@ function ListBoxInline({ options, layout }) {
     >
       {toolbar && (
         <Grid item container style={{ padding: theme.spacing(1) }} wrap="nowrap">
-          <Grid item container wrap="nowrap">
+          <Grid item container style={{ height: '32px' }} wrap="nowrap">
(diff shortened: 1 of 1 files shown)


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -209,7 +210,7 @@ function ListBoxInline({ options, layout }) {
     >
       {toolbar && (
         <Grid item container style={{ padding: theme.spacing(1) }} wrap="nowrap">
-          <Grid item container wrap="nowrap">
+          <Grid item container style={{ height: `${headerHeight}px` }} wrap="nowrap">
```
johanlahti, line 213:
```suggestion
          <Grid item container height={headerHeight} wrap="nowrap">
```
or with `sx`.

quanho: Updated!

johanlahti (approved):

PR #1126, opened 2023-03-11
fix: prevent flickering from engine update during selection

T-Wizard:
## Motivation

Refactor client side selection state to prevent flickering from engine update during selection

select without waiting for previous selection to be done (work for range select as local client state is updated)

(I found one edge case (also in old sense-client listbox) if a failed selection is followed by a working selection (before the response that the first failed) that unnecessary also clear both)

Properly wait for update from onging selection before starting whenthe field is calculated. (before it only waited for the selectListObjectValues to be done not for fetching the new layout/getListObjectData)

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/hooks/selections/useSelectionsInteractions.js
+++ b/apis/nucleus/src/components/listbox/hooks/selections/useSelectionsInteractions.js
@@ -1,170 +1,137 @@
-import { useEffect, useState, useCallback } from 'react';
-import {
-  getUniques,
-  selectValues,
-  applySelectionsOnPages,
-  fillRange,
-  getSelectedValues,
-  getElemNumbersFromPages,
-} from './listbox-selections';
+import { useEffect, useCallback, useRef } from 'react';
+import { selectValues, fillRange, getElemNumbersFromPages } from './listbox-selections';
 
-export default function useSelectionsInteractions({
-  layout,
-  selections,
-  pages = [],
-  checkboxes = false,
-  selectDisabled,
-  doc = document,
-  isSingleSelect: singleSelect = false,
-}) {
-  const [instantPages, setInstantPages] = useState(pages);
-  const [mouseDown, setMouseDown] = useState(false);
-  const [selectingValues, setSelectingValues] = useState(false);
-  const [isSingleSelect, setIsSingleSelect] = useState(singleSelect);
-  const [selected, setSelected] = useState([]);
-  const [isRangeSelection, setIsRangeSelection] = useState(false);
-  const [preSelected, setPreSelected] = useState([]);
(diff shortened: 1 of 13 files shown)


apis/nucleus/src/components/listbox/hooks/selections/useSelectionsInteractions.js
```diff
@@ -174,22 +141,24 @@ export default function useSelectionsInteractions({
...
 
-  useEffect(() => {
-    if (selectingValues || !pages || (!checkboxes && !mouseDown)) {
-      return;
-    }
-    // Render pre-selections before they have been selected in Engine.
-    const newPages = applySelectionsOnPages(pages, preSelected, isSingleSelect);
-    setInstantPages(newPages);
-  }, [preSelected]);
+    // selections.on('activated', onActivated);
```
johanlahti, line 153: rm

apis/nucleus/src/components/listbox/hooks/selections/useSelectionsInteractions.js
```diff
@@ -174,22 +141,24 @@ export default function useSelectionsInteractions({
...
+    selections.on('deactivated', onDeactivated);
+    selections.on('cleared', onCleared);
+    return () => {
+      // selections.removeListener('deactivated', onActivated);
```
johanlahti, line 157:
```suggestion
```

apis/nucleus/src/components/listbox/hooks/selections/useSelectionsInteractions.js
johanlahti:
```suggestion
```

test/rendering/listbox/listbox.js
```diff
@@ -5,6 +5,8 @@
       id: `listbox-${+new Date()}`,
       getListObjectData: async () => (fixture ? fixture.getListObjectData() : getMockData()),
       getLayout: async () => (fixture ? fixture.getLayout() : getListboxLayout()),
+      beginSelections: async () => {},
+      selectListObjectValues: async () => ({ qSucsess: true }),
```
johanlahti, line 9:
```suggestion
      selectListObjectValues: async () => ({ qSuccess: true }),
```

T-Wizard: that the test still worked mean I overcomplicated it and it should be `async () => true` (also updated)

Caele (approved):
I've done a test-run in a mashup, works nicely, no apparent issues. CMD-select works fine. 

But I get a console error from ListboxRowColumn.jsx r109 when trying CTRL-click, its a remnant from me missing to remove that onClick. I don't think it is needed anymore?

quanho (approved):

johanlahti (approved):

PR #1127, opened 2023-03-11
fix: hide search icon and scrollbar of a filter pane in edit mode

quanho:
In edit mode, the search icon should be not shown and scrolling should be disabled (or the scroll bar should not be shown).
This PR is to fix https://github.com/qlik-oss/sn-list-objects/issues/72.

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -211,24 +211,26 @@ function ListBoxInline({ options, layout }) {
       {toolbar && (
         <Grid item container style={{ padding: theme.spacing(1) }} wrap="nowrap">
           <Grid item container height={headerHeight} wrap="nowrap">
-            <Grid item sx={{ display: 'flex', alignItems: 'center', width: searchIconWidth }}>
-              {isLocked ? (
-                <IconButton tabIndex={-1} onClick={unlock} disabled={!isLocked} size="large">
-                  <Lock title={translator.get('Listbox.Unlock')} style={{ fontSize: '12px' }} />
-                </IconButton>
-              ) : (
-                searchEnabled !== false && (
-                  <IconButton
-                    onClick={onShowSearch}
-                    tabIndex={-1}
-                    title={translator.get('Listbox.Search')}
-                    size="large"
-                  >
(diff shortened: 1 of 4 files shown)


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -211,24 +211,26 @@ function ListBoxInline({ options, layout }) {
...
-                    onClick={onShowSearch}
-                    tabIndex={-1}
-                    title={translator.get('Listbox.Search')}
-                    size="large"
-                  >
-                    <SearchIcon style={{ fontSize: '12px' }} />
+            {(isLocked || (searchEnabled !== false && !constraints.active)) && (
+              <Grid item sx={{ display: 'flex', alignItems: 'center', width: searchIconWidth }}>
+                {isLocked ? (
+                  <IconButton tabIndex={-1} onClick={unlock} disabled={!isLocked} size="large">
```
T-Wizard, line 217: maybe also disable unlock if constraints.active (or constraints.select)

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -211,24 +211,26 @@ function ListBoxInline({ options, layout }) {
...
-              ) : (
-                searchEnabled !== false && (
-                  <IconButton
-                    onClick={onShowSearch}
-                    tabIndex={-1}
-                    title={translator.get('Listbox.Search')}
-                    size="large"
-                  >
-                    <SearchIcon style={{ fontSize: '12px' }} />
+            {(isLocked || (searchEnabled !== false && !constraints.active)) && (
```
T-Wizard, line 214: I recommend storing this in a variable. (as it will be even longer when support for drilldown icon is added)

apis/nucleus/src/components/listbox/ListBoxInline.jsx
quanho: Updated it!

apis/nucleus/src/components/listbox/ListBoxInline.jsx
quanho: Updated it!

DanielS-Qlik (approved):

PR #1136, opened 2023-03-13
fix: fixed frequency width and hide when width is too small

johanlahti:
### Before:
image

### After:
image

### Before
Values not aligned as "columns"

### After (fixed frequency width)
image

The problem with a fixed frequency width is that we will show ellipsis when the frequency value takes too much space. Don't know if `display: 'grid'` could solve this (would require a rewrite).
Another option is to estimate the width based on actual frequency values, like we do with labels.
This implementation is at par (or slightly better) than the existing Listbox/Filter pane.

## Motivation

- fixes(https://github.com/qlik-oss/sn-list-objects/issues/122)
- fixes("Frequency values do not align properly under one another")
- fixes("values partly hidden when shown in italics")

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


T-Wizard (approved):

apis/nucleus/src/components/listbox/assets/__tests__/get-list-sizes.test.jsx
```diff
@@ -90,6 +93,64 @@ describe('get-list-sizes', () => {
...
+    const sizes = getListSizes(args);
+    expect(sizes).toMatchObject({
+      columnWidth: MIN_WIDTH,
+      maxCount: {
+        column: 599107,
+      },
+    });
+  });
+
+  it('A minimum item width should kick in if text is short and reverve extra space for frequency', () => {
```
DanielS-Qlik, line 137: "reverve" -> reserve, just a typo

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -177,7 +178,13 @@ export default function ListBox({
     model,
   });
 
-  const sizes = getListSizes({ layout, width, height, listCount: unlimitedListCount, count, textWidth });
+  let freqIsAllowed = getFrequencyAllowed({ itemWidth: width, layout, frequencyMode });
+  const sizes = getListSizes({ layout, width, height, listCount: unlimitedListCount, count, textWidth, freqIsAllowed });
+  if (sizes.columnWidth) {
```
DanielS-Qlik, line 183: can this be defined when not in grid mode? Maybe return and use `isGridMode` from `getListSizes`?

johanlahti: Yeah, I checked that up. `columnWidth` is only defined in grid mode, i.e. it's an indicator of grid mode, so it should be ok.

apis/nucleus/src/components/listbox/assets/__tests__/get-list-sizes.test.jsx
johanlahti: I will change in a subsequent PR if I remember 😄

PR #1137, opened 2023-03-13
feat!: change search option default to toggle

veinfors:
Breaking change!
Changes the default look/behaviour of the listbox to not show the search input field but instead can be toggled.

Before:
with_search

After:
without_search

This is how the field renders by default when rendering like this:
```
const fieldInstance = await n.field("Alpha");
fieldInstance.mount(element);
```

(The default value for the `search` option is now changed to `"toggle"` instead of true)


johanlahti (approved):
For a new major release.

Caele:
I don't think I fully understood the change, plus I've given it some more thought:
We don't need to change the default setting, that is not the problem. I think the issue was that we show two search icons when the search is true or false, when that second one is only needed when in toggle mode.

veinfors:
We decided not to make this Api change due to reasons mentioned above, hence I'm closing this PR. Will however hide search icon when search is set to true or false, since it adds no value in that case, but will do it in a separate PR.

PR #1141, opened 2023-03-14
fix: hide search toggle button when not needed

veinfors:
Only show search toggle button when option `search: "toggle"` is used.
When the search field is always visible or hidden (search: true/false), there is no need to show a search toggle button.


Caele (approved):
Nice, can you also add a test that has "toggle" set?

veinfors:
> Nice, can you also add a test that has "toggle" set?

sure, will do that!

PR #1142, opened 2023-03-14
fix: clear old handler when sn-filter-pane is cut/deleted

quanho:
On cutting or deleting a listbox, its handler become invalid so we should clear its handler.
This is to fix https://github.com/qlik-oss/sn-list-objects/issues/133

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


Caele (approved):

T-Wizard (approved):

T-Wizard (approved):
Maybe also add a new unit test to verify the change
(that the model is removed from the store when closed)

PR #1144, opened 2023-03-14
fix(Listbox): rtl mode

PurwaShrivastava:
## Motivation

RTL mode for the ListBox. This is to fix https://github.com/qlik-oss/sn-list-objects/issues/82

Before: Title, Search Icon, Columns & Toolbar were not correctly aligned for Folded as well as normal ListBox.
Number Alignment was not correct & - sign for negative numbers shown backwards.

After: Attached Screenshots
Litbox Title, Search Glass & Columns with Numeric positive negative as well as String values in RTL mode :

Screenshot 2023-03-15 at 11 36 50

Actions Toolbar in RTL Mode :

Screenshot 2023-03-16 at 11 06 18

Folded ListBox in RTL Mode : 

Screenshot 2023-03-15 at 11 37 20

## Requirements checklist

- [x] Api specification
  - [ ] Ran `yarn spec`


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -301,6 +319,7 @@ function ListBoxInline({ options, layout }) {
                 onConfirm: () => {},
                 onCancel: () => {},
               }}
+              direction={isRtl}
```
Caele, line 322: This doesn't look right? 
Could we use `direction` (rtl/ltr) for all properties passed between components and only use `isRtl` for checks witin a component? Its easier to understand if we are consistent in what properties mean.

PurwaShrivastava: ok shall change

apis/nucleus/src/components/ActionsToolbar.jsx
```diff
@@ -45,9 +45,9 @@ const ActionToolbarElement = {
 };
 
 const ActionsGroup = React.forwardRef(
-  ({ ariaExpanded = false, actions = [], first = false, last = false, addAnchor = false }, ref) =>
+  ({ ariaExpanded = false, actions = [], first = false, last = false, addAnchor = false, isRtl = false }, ref) =>
```
Caele, line 48: This using isRtl as a property here.

Caele:
- Look at adding a rendering test to cover the RTL case
- The current rendering tests should not change or?

apis/nucleus/src/components/ActionsToolbar.jsx
PurwaShrivastava: Have forwarded direction in the main element ActionsToolbar, calculated isRTL and using it inside the inner element, seems redundant to pass direction again and recalculate isRtl

Caele:
> I have updated the snapshots for rendering tests, have added unit tests for RTL case. Can look into adding rendering tests.

I meant that the changes you have made should not affect the current rendering tests, so was there an intentional style update there somewhere or?

PurwaShrivastava:
> > I have updated the snapshots for rendering tests, have added unit tests for RTL case. Can look into adding rendering tests.
> 
> I meant that the changes you have made should not affect the current rendering tests, so was there an intentional style update there somewhere or?



> > I have updated the snapshots for rendering tests, have added unit tests for RTL case. Can look into adding rendering tests.
> 
> I meant that the changes you have made should not affect the current rendering tests, so was there an intentional style update there somewhere or?

Have introduced a right padding for header.

PR #1150, opened 2023-03-15
fix: grid mode calculations

johanlahti:
## Motivation

Improve algorithm for grid rows and columns:
- fixes(https://github.com/qlik-oss/sn-list-objects/issues/144)
- fixes(https://github.com/qlik-oss/sn-list-objects/issues/146)
- fixes(https://github.com/qlik-oss/sn-list-objects/issues/127)

Remove ticks in small sizes:
- fixes(https://github.com/qlik-oss/sn-list-objects/issues/145)

### Solution:

I will fall back to auto mode when the algorithm hits a limitation (i.e. when rows/cols cannot fit – as is seen in the **video below**). Otherwise, items will become smaller than the minimum allowed item width. We may need to communicate this to the user somehow.

In 

https://user-images.githubusercontent.com/5780544/225383379-4fd2ca47-0756-4348-a9bd-1ed377cdc650.mov

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


apis/nucleus/src/components/listbox/assets/get-list-sizes/column-mode.js
```diff
@@ -0,0 +1,40 @@
...
+      // If items cannot fit inside height, fall back to auto mode.
+      return calculateColumnMode({
+        maxVisibleRows: { ...maxVisibleRows, auto: true },
+        itemHeight,
+        listCount,
+        listHeight,
+        columnAutoWidth,
+        containerWidth,
+      });
+    }
```
T-Wizard, line 29: I think you would get the same result by limiting the rowCount to not be larger the autoRowCount
```suggestion
    rowCount = Math.min(listCount, maxRows, autoRowCount);
```

T-Wizard (approved):
I think that the logic for to many rows/columns could be done simpler by limiting the user set value to not be larger the value you would get from auto.
(but suggeted solution is OK)

I think it would give the same effect as long as auto try to fit as many rows/columns as possible.
(but maybe we want it to be possible to get smaller columns by setting it manual then you would get by auto, but then I also think you should fallback to that extreme value when set to large)

I also a bit questioning on basing show tick on item size.
maybe a manual setting and/or more complicated logic to hide tick if it avoids the need for scroll would be better.
(but I will accept it as is)

T-Wizard:
consider also adding some grid mode rendering test. (for a better picture)

apis/nucleus/src/components/listbox/assets/get-list-sizes/column-mode.js
johanlahti: Yes, that's essentially doing the same thing but shorter 👍 
I will change in the row calculation too, so we can avoid the recursive call.

johanlahti:
- Simpler code avoiding recursive call  ✅ Changed
- I think it's nice to fall back to auto when we cannot fit values with given min width – or we have to accept this scenario: https://github.com/qlik-oss/sn-list-objects/issues/127 
- Breakpoint for hiding tick mark: This was a request from the design team – but I can imagine we will reiterate over this and possibly add a setting or at least smarter calculation based on text width

PR #1151, opened 2023-03-15
fix: fix set focus on keyboard navigation

quanho:
This is to fix focus on keyboard navigation.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -45,6 +45,12 @@ const StyledGrid = styled(Grid, { shouldForwardProp: (p) => !['containerPadding'
     [`& .${classes.listboxWrapper}`]: {
       padding: containerPadding,
     },
+    '&:focus': {
+      boxShadow: `inset 0 0 0 2px ${theme.palette.custom.focusBorder} !important`,
+    },
+    '&:focus-visible': {
+      outline: 'none',
+    },
   })
 );
 
--- a/apis/nucleus/src/components/listbox/interactions/listbox-keyboard-navigation.js
+++ b/apis/nucleus/src/components/listbox/interactions/listbox-keyboard-navigation.js
@@ -49,7 +49,8 @@ export function getFieldKeyboardNavigation({ select, confirm, cancel, setScrollP
         break;
       case KEYS.ESCAPE:
         cancel();
-        return; // let it propagate to top-level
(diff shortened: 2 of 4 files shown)


apis/stardust/types/index.d.ts
```diff
@@ -584,7 +584,7 @@ declare namespace stardust {
         /**
          * Function used by the visualization to tell Nebula to it wants to relinquish focus
          */
-        blur?(): void;
+        blur?(b?: boolean): void;
```
T-Wizard, line 587: you should not change direct in this file, but instead change the jsdoc source that is ised to generate it (and the json spec)

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -337,7 +343,7 @@ function ListBoxInline({ options, layout }) {
                   rangeSelect={rangeSelect}
                   checkboxes={checkboxes}
                   height={height}
-                  width={width}
+                  width={width - 3}
```
johanlahti, line 346: This will reintroduce the problem of having same margin on both sides in grid mode. It has to be added somewhere else.

apis/nucleus/src/components/listbox/interactions/listbox-keyboard-navigation.js
```diff
@@ -129,6 +130,9 @@ export function getListboxInlineKeyboardNavigation({
       case KEYS.ARROW_DOWN:
         updateKeyScrollOnHover({ down: 1 });
         break;
+      case KEYS.ARROW_LEFT:
+      case KEYS.ARROW_RIGHT:
+        return; // let it propagate to top-level
```
T-Wizard, line 135: this should not be needed as default is also return.
but maybe we still want it for better info

apis/nucleus/src/components/listbox/interactions/listbox-keyboard-navigation.js
```diff
@@ -49,7 +49,8 @@ export function getFieldKeyboardNavigation({ select, confirm, cancel, setScrollP
         break;
       case KEYS.ESCAPE:
         cancel();
-        return; // let it propagate to top-level
+        elementToFocus = event.currentTarget.closest('.listbox-container'); // focus jump from a cell to a listbox parent
+        break;
```
T-Wizard, line 53: why this change?
it looks like it should have bubbled up and be handled in listbox inline keydown before
what that broken?

Have you tested this change in ListBoxPopover case? (that closing the popover case is unchanged)

johanlahti (approved):

PR #1152, opened 2023-03-16
feat: fetch frequencyMax

T-Wizard:
do an extra app.evaluateEx call to fetch frequencyMax when it is set to the value 'fetch'
update useOnTheFlyModel.jsx to only frequencyMax when actually needed
update useOnTheFlyModel.jsx to use it for drilldowns (maybe use it all library dimensions?)

## Motivation

For use with library dimensions as it is not possible for a static qValueExpression to handle that the library dimension changes field it uses (common for drilldowns but possible for any lib dimension)

the use of a marker value to trigger is are because without it there is not enough information in the layout to know if frequencyMax needs to be fetched.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/nucleus/src/components/listbox/hooks/useOnTheFlyModel.jsx
```diff
@@ -122,11 +127,16 @@ export default function useOnTheFlyModel({ app, fieldIdentifier, stateName, opti
     fieldName = fieldIdentifier;
   }
 
-  if (frequencyMode !== 'N' || histogram) {
-    const field = fieldIdentifier.qLibraryId ? fieldDef : fieldName;
-    listdef.frequencyMax = {
-      qValueExpression: `Max(AGGR(Count([${field}]), [${field}]))`,
-    };
+  if (frequencyMode !== 'P' && histogram) {
```
quanho, line 130: Should we update isFrequencyMaxNeeded above from 
`const isFrequencyMaxNeeded = options.histogram || options.frequencyMode !== 'N';`
to
`const isFrequencyMaxNeeded = options.histogram && options.frequencyMode === 'V';`

and use it here?

quanho: No it is in different scope!

PR #1156, opened 2023-03-16
fix: fix getting listbox theme

quanho:
## Motivation

Fix getting theme for listbox.

This is to fix https://github.com/qlik-oss/sn-list-objects/issues/78.

There is still another problem of theme inheriting in nebula.js. E.g. Breeze theme has no font-family defined for listbox so it should inherit from Sense classic theme. This will be fixed in another PR.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


DanielS-Qlik (approved):

T-Wizard:
It broke a lot of rendering test
Maybe need some more thought on how it should work outside sense with only nebula.js default theme.

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -144,6 +144,7 @@ export default function ListBoxSearch({
       sx={[
         {
           border: 'none',
+          fontSize: 14,
```
Caele, line 147: Added this as well to the search field isn't huge compared to the rest

PR #1157, opened 2023-03-17
refactor: correct and simplify frequency bar calculation

johanlahti:
## Motivation

- fixes(https://github.com/qlik-oss/sn-list-objects/issues/152) 

#### Before:

Dynamic bars, calculated from the frequency value, *where padding etc. was subtracted dynamically*, which made it more complicated and easier to break when new features are added into the row components (e.g. checkboxes and altered padding).

#### After:

Using two components:
- A **bar container** with fixed width (to fill the remaining row width)
- A **filled bar**: the visible bar, always having a %-width linearly congruent with the derived or actual frequency percent that we get from the backend

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


quanho (approved):

DanielS-Qlik:
Might want to add rendering test to the histogram as well

DanielS-Qlik (approved):

johanlahti:
Yeah, I will add some rendering tests, we don't have any yet for histogram in nebula.

PR #1159, opened 2023-03-17
feat: search & selectAll shortcuts

T-Wizard:
fix qlik-oss/sn-list-objects#161

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/nucleus/src/components/listbox/interactions/listbox-keyboard-navigation.js
```diff
@@ -66,6 +74,18 @@ export function getFieldKeyboardNavigation({ select, confirm, cancel, setScrollP
         }
         setScrollPosition?.('end');
         break;
+      case KEYS.A:
+        if (ctrlKey || metaKey) {
```
T-Wizard, line 78: Mac question: should this check metaKey?
It does in the old listbox in sense-client, but the crtl-End + ctrl-Home above don't

veinfors (approved):

PR #1161, opened 2023-03-17
fix: support editing listbox title

quanho:
## Motivation

To fix https://github.com/qlik-oss/sn-list-objects/issues/158 by checking constraints.active. If it is true, it means we are in edit mode so a listbox label can be edited.

Editable title

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -238,7 +238,7 @@ function ListBoxInline({ options, layout }) {
     containerPadding = layoutOptions.layoutOrder === 'row' ? '2px 4px' : '2px 6px 2px 4px';
   }
 
-  const titleComponent = constraints?.active ? (
+  const TitleComponent = constraints?.active ? (
     <EditableTitle layout={layout} model={model} />
   ) : (
     <Title variant="h6" noWrap ref={titleRef} title={layout.title}>
@@ -311,7 +311,7 @@ function ListBoxInline({ options, layout }) {
                 </Grid>
               )}
               <Grid item sx={{ justifyContent: isRtl ? 'flex-end' : 'flex-start' }} className={classes.listBoxHeader}>
-                {showTitle && titleComponent}
+                {showTitle && TitleComponent}
               </Grid>
             </Grid>
             <Grid item xs />
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -79,6 +79,7 @@ function ListBoxInline({ options, layout }) {
     calculatePagesHeight,
     showGray = true,
     scrollState = undefined,
+    isInSenseClientAndAlowEdittingTitle = false,
   } = options;
   let { toolbar = true } = options;
 
@@ -238,13 +239,14 @@ function ListBoxInline({ options, layout }) {
     containerPadding = layoutOptions.layoutOrder === 'row' ? '2px 4px' : '2px 6px 2px 4px';
   }
 
-  const TitleComponent = constraints?.active ? (
-    <EditableTitle layout={layout} model={model} />
-  ) : (
-    <Title variant="h6" noWrap ref={titleRef} title={layout.title}>
-      {layout.title}
-    </Title>
-  );
+  const TitleComponent =
+    isInSenseClientAndAlowEdittingTitle && constraints?.active ? (
+      <EditableTitle layout={layout} model={model} />
+    ) : (
+      <Title variant="h6" noWrap ref={titleRef} title={layout.title}>
+        {layout.title}
+      </Title>
+    );
   return (
     <>
       {toolbarDetachedOnly && <ActionsToolbar direction={direction} {...getActionToolbarProps(true)} />}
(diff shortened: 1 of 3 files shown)


apis/nucleus/src/components/listbox/components/EditableTitle.jsx
```diff
@@ -0,0 +1,83 @@
...
+const Title = styled(Typography)(({ theme }) => ({
+  color: theme.listBox?.title?.main?.color,
+  fontSize: theme.listBox?.title?.main?.fontSize,
+  fontFamily: theme.listBox?.title?.main?.fontFamily,
+}));
+
+const StyledInput = styled(InputBase)(({ theme }) => ({
+  color: theme.listBox?.title?.main?.color,
+  fontSize: theme.listBox?.title?.main?.fontSize,
+  fontFamily: theme.listBox?.title?.main?.fontFamily,
```
DanielS-Qlik, line 15: To keep the font while changing from `Title` to `StyledInput`
```suggestion
    fontSize: theme.listBox?.title?.main?.fontSize ?? theme.typography.h6.fontSize,
    fontFamily: theme.listBox?.title?.main?.fontFamily ?? theme.typography.h6.fontFamily,
    fontWeight: theme.listBox?.title?.main?.fontWeight ?? theme.typography.h6.fontWeight,
```

apis/nucleus/src/components/listbox/components/EditableTitle.jsx
```diff
@@ -0,0 +1,83 @@
...
+
+export default function EditableTitle({ layout, model }) {
+  const [properties] = useProperties(model);
+  const [titleDef, setTitleDef] = useState(
+    properties?.title?.qStringExpression?.qExpr ? `=${properties?.title?.qStringExpression?.qExpr}` : properties?.title
+  );
+  const [isTitleFocused, setIsTitleFocused] = useState(false);
+  const isCanceledChange = useRef(false);
+
+  const confirmChange = () => {
```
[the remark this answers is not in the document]
quanho: After the user types something but wants to cancel it by pressing Esc key, we need to cancel the change. That's why I name it isCanceledChange.

apis/nucleus/src/components/listbox/components/EditableTitle.jsx
quanho: Updated!

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -237,6 +238,13 @@ function ListBoxInline({ options, layout }) {
     containerPadding = layoutOptions.layoutOrder === 'row' ? '2px 4px' : '2px 6px 2px 4px';
   }
 
+  const titleComponent = constraints?.active ? (
```
johanlahti, line 241:
```suggestion
  const TitleComponent = constraints?.active ? (
```

johanlahti:
Does it also work with keyboard navigation in general?

quanho:
> Does it also work with keyboard navigation in general?
> 
> Unit tests?

There is no support for keyboard navigation in Edit mode.

apis/nucleus/src/components/listbox/ListBoxInline.jsx
quanho: A bit unclear of naming here. This is a variable to refer to a component so I am not sure if it should be a normal variable name or and component name.
Updated it anyway.

Caele:
Won't this be pretty strange in a mashup?
Say that you want to render without selections etc, then you for some odd reason would be able to edit the title?

Quickest way to get it working would be to pass in an option from the filterpane to enable the behaviour (and not document it to begin with).

quanho:
> Won't this be pretty strange in a mashup? Say that you want to render without selections etc, then you for some odd reason would be able to edit the title?
> 
> Quickest way to get it working would be to pass in an option from the filterpane to enable the behaviour (and not document it to begin with).
Can you use constraints.active in mashup?

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -237,6 +238,13 @@ function ListBoxInline({ options, layout }) {
     containerPadding = layoutOptions.layoutOrder === 'row' ? '2px 4px' : '2px 6px 2px 4px';
   }
 
+  const TitleComponent = constraints?.active ? (
```
T-Wizard, line 241: there need also be a check of permission if the user has can do setProperties on the object

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -78,6 +79,7 @@ function ListBoxInline({ options, layout }) {
     calculatePagesHeight,
     showGray = true,
     scrollState = undefined,
+    isInSenseClientAndAlowEdittingTitle = false,
```
johanlahti, line 82:
```suggestion
    isInSenseClientAndAllowEditingTitle = false,
```

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -237,6 +239,14 @@ function ListBoxInline({ options, layout }) {
...
   }
 
+  const TitleComponent =
+    isInSenseClientAndAlowEdittingTitle && constraints?.active ? (
+      <EditableTitle layout={layout} model={model} />
+    ) : (
+      <Title variant="h6" noWrap ref={titleRef} title={layout.title}>
+        {layout.title}
+      </Title>
+    );
```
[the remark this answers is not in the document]
quanho: We discussed about this and see we don't need to check since we do not support filter pane in storytelling which is the case we concern.

PR #1162, opened 2023-03-19
feat: return promise from field mount fn

veinfors:
Return a promise from the field mount function to indicate when all data fetching is done.
This can be used to determine when the listbox finishes rendering.

Awaits data pages to arrive and maxFrequencyCount when needed.

- [ ] Integration test added for verification.

### DIFF
--- a/apis/nucleus/src/index.js
+++ b/apis/nucleus/src/index.js
@@ -446,13 +446,14 @@ function nuked(configuration = {}) {
            * @param {boolean=} [options.dense=false] Reduces padding and text size (not applicable for existing objects)
            * @param {string=} [options.stateName="$"] Sets the state to make selections in (not applicable for existing objects)
            * @param {object=} [options.properties={}] Properties object to extend default properties with
+           * @returns {Promise<void>} A promise that resolves when the data is fetched.
            *
            * @since 1.1.0
            * @instance
            * @example
            * fieldInstance.mount(element);
            */
-          mount(element, options = {}) {
+          async mount(element, options = {}) {
             if (!element) {
               throw new Error(`Element for ${fieldName || qId} not provided`);
             }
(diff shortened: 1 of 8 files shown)


apis/nucleus/src/index.js
```diff
@@ -446,13 +446,14 @@ function nuked(configuration = {}) {
...
            * @param {object=} [options.properties={}] Properties object to extend default properties with
+           * @returns {Promise<void>} A promise that resolves when the data is fetched.
            *
            * @since 1.1.0
            * @instance
            * @example
            * fieldInstance.mount(element);
            */
-          mount(element, options = {}) {
+          async mount(element, options = {}) {
```
Caele, line 456: Needs a spec update. Normally a very breaking change, but as we didn't return anything before it should be quite safe.

apis/nucleus/src/index.js
```diff
@@ -461,19 +462,31 @@ function nuked(configuration = {}) {
...
-              }),
-              stateName: options.stateName || '$',
+            let renderCount = 0;
+
+            return new Promise((resolve) => {
+              const renderedCallback = () => {
+                renderCount++;
+                if (renderCount === 1) {
+                  resolve();
+                }
```
Caele, line 472: This looks a bit unnecessary?

T-Wizard:
how does it work for supernova objects?
and maybe we should use the same api for fields?

Caele:
> how does it work for supernova objects? and maybe we should use the same api for fields?

They have an onMount thing that I'm not 100% sure why it actually works. I've looked at making it so that the return "Viz" resolves when the chart is rendered (in my createApi PR) but that was quite the hassle.

apis/nucleus/src/index.js
veinfors: Sure, it can be simplified, I've updated it.

apis/nucleus/src/index.js
veinfors: Updated ✅

veinfors:
> > how does it work for supernova objects? and maybe we should use the same api for fields?
> 
> They have an onMount thing that I'm not 100% sure why it actually works. I've looked at making it so that the return "Viz" resolves when the chart is rendered (in my createApi PR) but that was quite the hassle.
> 
> I think the Viz resolves when the onMount callback into Cell resolves, which is done when the loadType function finishes. Then there is a renderCount thing in Supernova that also does stuff... pending more reading...

Yeah, I made the assumption that the promise returned from the `render` function resolves once the supernova has no more outstanding, unresolved, promises (basically the rendering is done...). And then I added this promise return from the `field` -> `mount` function to make it work in an equivalent way.

But maybe the render function doesn't work like I assumed...

T-Wizard:
> > > how does it work for supernova objects? and maybe we should use the same api for fields?
> > 
> > 
> > They have an onMount thing that I'm not 100% sure why it actually works. I've looked at making it so that the return "Viz" resolves when the chart is rendered (in my createApi PR) but that was quite the hassle.
> > I think the Viz resolves when the onMount callback into Cell resolves, which is done when the loadType function finishes. Then there is a renderCount thing in Supernova that also does stuff... pending more reading...
> 
> Yeah, I made the assumption that the promise returned from the `render` function resolves once the supernova has no more outstanding, unresolved, promises (basically the rendering is done...). And then I added this promise return from the `field` -> `mount` function to make it work in an equivalent way.
> 
> But maybe the render function doesn't work like I assumed...

It is hard to track but I think it **don't** wait for the render.
and that you need to use option.onInitialRender to get that for supernova charts

Caele:
So I dug:
`viz = await nebiie.render(...)` will resolve when the Cell has finished loading and creating the chart, not when it is done rendering. So basically when the Viz API is ready. There is an option to pass in (undocumented) called `onInitialRender` that I had trouble getting to work.
In Nebulas tests we use 'data-render-count' attribute to check if it is done.

In client we call the innermost render function directly, which resolves when all the hook-ing is done.

Christians approach here is fine I think, good if it resolves when rendering is done.

Caele (approved):

PR #1163, opened 2023-03-20
fix(Listbox): no box-shadow on hover

DanielS-Qlik:
## Motivation

Don't add box-shadow on hover.
Fixes https://github.com/qlik-oss/sn-list-objects/issues/170

Before:
Kapture 2023-03-20 at 13 51 48

After:
Kapture 2023-03-20 at 13 49 18

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


veinfors (approved):

T-Wizard:
In (maybe) the slightly longer perspective I think we want the more complicated solution of not stealing focus on hover

DanielS-Qlik:
> 

Yes, in the old listbox the "virtual scroll" from sense is used to enable the user to scroll with the keyboard on hover. But it seems to involve quite a lot of work.

Caele (approved):

PR #1165, opened 2023-03-20
feat: pick up enigmaOptions from fixture

veinfors:
Makes it possible to pass in options to enigmaMocker even when using it indirectly through nebula serve by picking up `enigmaOptions` from the root of the fixture.


DanielS-Qlik:
Maybe add a note about it in this example: https://github.com/qlik-oss/nebula.js/blob/1ac8a87c38a97998e04951378661294e2c3cddca/commands/serve/docs/fixture-format.md?plain=1#L76

veinfors:
> Maybe add a note about it in this example:
> 
> https://github.com/qlik-oss/nebula.js/blob/1ac8a87c38a97998e04951378661294e2c3cddca/commands/serve/docs/fixture-format.md?plain=1#L76

Good point, did that!

PR #1167, opened 2023-03-21
fix: do not show search box on selection

quanho:
## Motivation
Currently if the listbox has no title, the search box will be shown on selection activated. However, search box should only be shown if the user clicks the search icon, otherwise it should not be shown.
This is to fix https://github.com/qlik-oss/sn-list-objects/issues/181.
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -150,6 +150,9 @@ function ListBoxInline({ options, layout }) {
     };
     const hide = () => {
       setShowToolbar(false);
+      if (search === 'toggle') {
+        setShowSearch(false);
+      }
     };
     if (selections) {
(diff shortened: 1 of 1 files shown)


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -147,15 +147,9 @@ function ListBoxInline({ options, layout }) {
...
       setShowToolbar(true);
-      if (search === 'toggle' && toolbar === false) {
-        setShowSearch(true);
-      }
     };
     const hide = () => {
       setShowToolbar(false);
-      if (search === 'toggle') {
-        setShowSearch(false);
-      }
```
T-Wizard, line 158: keep this.
we still want to close search when you end the selection mode

DanielS-Qlik (approved):

PR #1168, opened 2023-03-21
feat: add useEmitter and corresponding viz api

Caele:
## Motivation

Adds new useEmitter hook to enable charts to affect mashups and client.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


apis/nucleus/src/viz.js
```diff
@@ -109,6 +112,24 @@ export default function viz({ model, halo, initialError, onDestroy = async () =>
...
     },
+    /**
+     * Listens to custom events from inside the visualization. See useEmitter
+     * @experimental
+     * @param {string} e Event name to listen to
+     * @param {Function} cb Callback function to invoke
+     */
+    on(e, cb) {
+      emitter.on(e, cb);
+    },
```
T-Wizard, line 123: consider longer names for better readability
(node docs use eventName and listener (https://nodejs.org/api/events.html#emitteroneventname-listener)

T-Wizard (approved):
consider if we want to expose a bit more then just on and off to the mashup
ex: once, removeAllListeners, setMaxListeners
but that could be done later when we see a need for it

Caele:
> consider if we want to expose a bit more then just on and off to the mashup ex: once, removeAllListeners, setMaxListeners but that could be done later when we see a need for it

yea, keeping this experimental for now, so just adding bare minimum

niekvanstaveren (approved):

PR #1169, opened 2023-03-21
chore: update enigma

Caele:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


niekvanstaveren:
This is good, but we probably also want to update the schemas that are imported everywhere:
e.g. https://github.com/qlik-oss/nebula.js/blob/master/commands/serve/web/connect.js#L2

Caele:
> 

yea, thought I did that, good catch

veinfors (approved):

PR #1172, opened 2023-03-21
Hva/sn filter pane 184

quanho:
## Motivation

To fix https://github.com/qlik-oss/sn-list-objects/issues/184.
Ticks and lock icons for cells should not be shown in grid mode row and no frequency count. Based on this we update padding for cells.
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/components/ListBoxRowColumn/ListBoxRowColumn.jsx
+++ b/apis/nucleus/src/components/listbox/components/ListBoxRowColumn/ListBoxRowColumn.jsx
@@ -164,7 +164,7 @@ function RowColumn({ index, rowIndex, columnIndex, style, data }) {
   const isFirstElement = index === 0;
   const flexBasisVal = checkboxes ? 'auto' : 'max-content';
 
-  const showLock = isSelected && isLocked;
+  const showLock = isSelected && isLocked && sizePermitsTick;
   const showTick = !checkboxes && isSelected && !isLocked && sizePermitsTick;
   const showIcon = !checkboxes && (showTick || showLock);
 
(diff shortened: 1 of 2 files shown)


apis/nucleus/src/components/listbox/components/ListBoxRowColumn/components/ListBoxRoot.jsx
```diff
@@ -68,7 +68,7 @@ const RowColRoot = styled('div', {
     minWidth: 0,
     flexGrow: 1,
     // Note that this padding is overridden when using checkboxes.
-    paddingLeft: `${CELL_PADDING_LEFT}px`,
```
T-Wizard, line 71: this (and the other changes) should not looks att isGridMode but if the current layout show Ticks (/Icons)

T-Wizard (changes requested):

apis/nucleus/src/components/listbox/components/ListBoxRowColumn/ListBoxRowColumn.jsx
```diff
@@ -164,7 +164,7 @@ function RowColumn({ index, rowIndex, columnIndex, style, data }) {
   const isFirstElement = index === 0;
   const flexBasisVal = checkboxes ? 'auto' : 'max-content';
 
-  const showLock = isSelected && isLocked;
+  const showLock = isSelected && isLocked && sizePermitsTick;
```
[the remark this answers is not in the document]
quanho: Updated!

apis/nucleus/src/components/listbox/components/ListBoxRowColumn/components/ListBoxRoot.jsx
quanho: Updated!

johanlahti (approved):

T-Wizard (approved):
OK. but we should probably also add a rendering test for the fixed cases

PR #1177, opened 2023-03-22
fix: show search icon in edit mode if the search checkbox is on

quanho:
## Motivation

This is to fix https://github.com/qlik-oss/sn-list-objects/issues/183 and also make search icon and lock icon more readable when it is disabled.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -245,7 +245,7 @@ function ListBoxInline({ options, layout }) {
     containerPadding = layoutOptions.layoutOrder === 'row' ? '2px 4px' : '2px 6px 2px 4px';
   }
 
-  const SearchIconComp = constraints?.active ? (
+  const searchIconComp = constraints?.active ? (
     <SearchIcon title={translator.get('Listbox.Search')} size="large" style={{ fontSize: '12px', padding: '7px' }} />
   ) : (
     <IconButton
@@ -260,7 +260,7 @@ function ListBoxInline({ options, layout }) {
     </IconButton>
   );
 
-  const LockIconComp = selectDisabled() ? (
+  const lockIconComp = selectDisabled() ? (
     <Lock size="large" style={{ fontSize: '12px', padding: '7px' }} />
   ) : (
     <IconButton title={translator.get('SelectionToolbar.ClickToUnlock')} tabIndex={-1} onClick={unlock} size="large">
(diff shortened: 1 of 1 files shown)


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -290,18 +290,21 @@ function ListBoxInline({ options, layout }) {
                       onClick={unlock}
                       disabled={selectDisabled()}
                       size="large"
+                      sx={{ '&.Mui-disabled': { color: '#000', opacity: 0.9 } }}
```
DanielS-Qlik, line 293: Can we get this from the theme?
If you change the background color in the theme to #000, you won't be able to change this to #FFF.

quanho: Refactor so we don't need to change color and opacity for icons in edit mode.

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -245,6 +245,33 @@ function ListBoxInline({ options, layout }) {
...
+      disableRipple
+      data-testid="search-toggle-btn"
+    >
+      <SearchIcon style={{ fontSize: '12px' }} />
+    </IconButton>
+  );
+
+  const LockIconComp = selectDisabled() ? (
+    <Lock
+      title={translator.get('SelectionToolbar.ClickToUnlock')}
```
T-Wizard, line 265: maybe change the tooltip in the case. as you can not click it

quanho: Removed it. There is no tooltip for lock icon in edit mode for old filter pane.

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -245,6 +245,33 @@ function ListBoxInline({ options, layout }) {
     containerPadding = layoutOptions.layoutOrder === 'row' ? '2px 4px' : '2px 6px 2px 4px';
   }
 
+  const SearchIconComp = constraints?.active ? (
```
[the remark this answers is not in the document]
quanho: @johanlahti said it should be capital because we can use it as a React component.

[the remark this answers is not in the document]
T-Wizard: no it is not.
Is is a react node, not a function that defines a react component.
it is when you use it like `<SearchIconComp />` that is should start with a capital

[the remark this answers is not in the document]
quanho: Updated it!

T-Wizard (approved):

PR #1182, opened 2023-03-23
fix: generic load function should work for existing charts

Caele:
## Motivation
When rendering an existing chart using a generic load function, the chart would fail to render due to it being not registerd.
Sample setup:
```
{
const nebbie = nebula.embed({
      context: {
        language: 'en-US',
        constraints: {
          select: false,
          passive: false,
          active: false,
        },
        theme: 'light',
      },
      load: ({ name, version }) => {
        if (name === 'barchart' || name === 'bar') {
          return Promise.resolve(bar);
        }
      }
});


nebbie.render({
  id: "lnjsg"
});

```
This change removes a check in Cell.jsx that stopped us getting to the generic load function. That error is now removed and replaced by a dev warning.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/Cell.jsx
+++ b/apis/nucleus/src/components/Cell.jsx
@@ -400,14 +400,6 @@ const Cell = forwardRef(
 
       // Load supernova
       const withVersion = types.getSupportedVersion(layout.visualization, layout.version);
-      if (!withVersion) {
-        if (__NEBULA_DEV__) {
-          // eslint-disable-next-line no-console
-          console.warn(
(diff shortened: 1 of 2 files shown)


T-Wizard (approved):
Should there even be a warning.
**when** you want the fallback to generic load, then the warning is just a distraction

Caele:
> Should there even be a warning. **when** you want the fallback to generic load, then the warning is just a distraction

yea, got rid of it now, instead shows that as an error message if the entire load fails.

PR #1185, opened 2023-03-24
fix(ListBox): show detached toolbar when layout.toolbar is false

quanho:
## Motivation

To fix https://github.com/qlik-oss/sn-list-objects/issues/216.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -141,7 +141,7 @@ function ListBoxInline({ options, layout }) {
   if (layout?.toolbar !== undefined) {
     toolbar = layout.toolbar;
   }
-  const toolbarDetachedOnly = layout?.toolbar === false || layout?.title === '';
+  const toolbarDetachedOnly = layout?.toolbar === false || (toolbar && layout?.title === '');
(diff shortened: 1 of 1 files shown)


DanielS-Qlik (approved):

T-Wizard:
I think we are using `toolbar: false` in analytic creation and there it should not use detached toolbar

veinfors:
> 

Good point, make sure to test so we don't change the behaviour in there before merging this

veinfors:
> I think we are using `toolbar: false` in analytic creation and there it should not use detached toolbar

Good point! Make sure to test so this change doesn't alter the behaviour in there...

DanielS-Qlik:
> 

In the last commit it changed to look at `layout.toolbar`

veinfors:
The way this is implemented now means that the `toolbar` option will get overridden by the toolbar prop on the layout, meaning the option is not "applicable"/reliable for existing objects.

It's best that we clarify that in the spec for the `toolbar` option, and also throw an error in `apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx` when that option is used for existing objects, as we do with other options like that...

T-Wizard (approved):

veinfors (approved):

PR #1195, opened 2023-03-28
feat(Listbox): auto dense

DanielS-Qlik:
## Motivation

When the listbox container is smaller than a grid item, enable dense mode.
fixes https://github.com/qlik-oss/sn-list-objects/issues/214 together with https://github.com/qlik-oss/sn-list-objects/pull/220

Kapture 2023-03-28 at 14 43 32

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


Caele:
There is a lot of logic and properties being passed around here, quite hard to follow its exact purpose.

What is is that triggers the auto dense setting? And does it simply result in dense being true?

DanielS-Qlik:
> There is a lot of logic and properties being passed around here, quite hard to follow its exact purpose.
> 
> What is is that triggers the auto dense setting? And does it simply result in dense being true?

Basically it checks the height of element passed in here https://github.com/qlik-oss/nebula.js/blob/bb82663c84116a6293f0e9621eabfbe5218b0267/apis/nucleus/src/index.js#L458.
If it's in grid mode and that element is smaller than a grid item, it enables dense mode via `autoDense`.

The state needs to be kept in ListboxInline.jsx because it's used by the `ListBoxSearch` there.
Then the state is set in ListBox.jsx via `useSetAutoDense()` because it needs the loaderRef to detect the scollbar presence.

veinfors:
I think we can wait with merging this as we might have better ideas to improve the behaviour instead...
(introducing a setting to the sn-filter-pane to explicitly control wether listboxes should collapse or not could help)

PR #1196, opened 2023-03-28
feat: add useinteractions add deprecate usecontraints

Caele:
## Motivation

- Deprecates useContraints in favor of useInteractions
- Adds an edit interaction state
- interactions are mirrored compared to contraints

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/nucleus/src/components/Cell.jsx
```diff
@@ -286,9 +286,13 @@ const Cell = forwardRef(
...
     const { nebbie } = halo.public;
-    const { disableCellPadding = false } = halo.context || {};
-
-    const { theme: themeName, translator, language, keyboardNavigation } = useContext(InstanceContext);
+    const {
+      theme: themeName,
+      translator,
+      language,
+      keyboardNavigation,
+      disableCellPadding = false,
```
Caele, line 294: I moved `disableCellPadding` into this statement instead of the previous check on halo.context.

apis/nucleus/src/index.js
```diff
@@ -307,8 +314,6 @@ function nuked(configuration = {}) {
           translator: locale.translator,
         };
 
-        halo.context = currentContext;
```
Caele, line 310: Removed this.
Only usage I could find was the `disableCellPadding` in Cell, so my guess is that this is not needed and just leads to incorrect patterns. @T-Wizard can you check? you did the cellPadding thing.

apis/nucleus/src/components/NebulaApp.jsx
```diff
@@ -64,6 +65,10 @@ export default function boot({ app, context }) {
   element.setAttribute('data-app-id', app.id);
   document.body.appendChild(element);
 
+  if (context) {
+    unifyContraintsAndInteractions(context);
```
Caele, line 69: This runs anytime the context updates, and aims to populate both `contraints` and `interactions` correctly. It is especially required for the new `edit` constraint as that should be `true` by default, all other could have stayed as undefined.

apis/nucleus/src/index.js
T-Wizard: how does this affect how you turn on `disableCellPadding` as a user of nebula.js?
(it is ised in the analytics repo)

T-Wizard: I don't remember why it was like that.
but yes it it should only be used in Cell

apis/nucleus/src/components/NebulaApp.jsx
```diff
@@ -110,6 +115,10 @@ export default function boot({ app, context }) {
       context(ctx) {
         (async () => {
           await rendered;
+          // Should be done here, unify contraints and interactions
+          if (ctx) {
+            unifyContraintsAndInteractions(ctx);
```
veinfors, line 120: Is it necessary to do this here, or will it be triggered on line 68 above anyway, making this superfluous? I can understand if this is needed though, to ensure it's kicks in immediately...

apis/stardust/types/index.d.ts
```diff
@@ -133,9 +133,18 @@ export function useAction<A>(factory: ()=>stardust.ActionDefinition<A>, deps?: a
...
  * and should respected by you when implementing the visualization.
+ * @deprecated
  */
 export function useConstraints(): stardust.Constraints;
 
+/**
+ * Gets the desired interaction states that should be applied when rendering the visualization.
+ * 
+ * The interactions are set on the embed configuration before the visualization is rendered
+ * and should respected by you when implementing the visualization.
```
veinfors, line 144:
```suggestion
 * and should be respected when implementing the visualization.

```

veinfors:
I'm thinking if it would be a good idea to make a console.warn etc. when using `useConstraints` instead of the `useInteraction` hook to inform the chart developer about the deprecation. We could make it happen only when running dev mode... Just an idea.

apis/nucleus/src/components/NebulaApp.jsx
Caele: The one on 68 should only run once (on boot), this one is for when someone updates the context by running embed.context(...)

apis/nucleus/src/components/NebulaApp.jsx
veinfors: Oh, of course, you are right. I was thinking of boot as being a react component or hook for some reason, but then I get it! 👍

veinfors (approved):

PR #1203, opened 2023-03-30
refactor: adjust action toolbar position

veinfors:
Moves the ActionToolbar slightly again to mitigate problem with overlapping of values.

Before:
bef

After:
aft

### DIFF
--- a/apis/nucleus/src/components/ActionsToolbar.jsx
+++ b/apis/nucleus/src/components/ActionsToolbar.jsx
@@ -120,7 +120,7 @@ function ActionsToolbar({
   const dividerStyle = useMemo(() => ({ margin: theme.spacing(0.5, 0) }));
 
   const popoverAnchorOrigin = {
-    vertical: 15,
+    vertical: 7,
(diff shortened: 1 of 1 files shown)


apis/nucleus/src/components/ActionsToolbar.jsx
```diff
@@ -120,7 +120,7 @@ function ActionsToolbar({
   const dividerStyle = useMemo(() => ({ margin: theme.spacing(0.5, 0) }));
 
   const popoverAnchorOrigin = {
-    vertical: 15,
+    vertical: 7,
```
T-Wizard, line 123: would it be much harder to only move it when there is no title?

veinfors: No, but we want this to be aligned with the hover menu and selection toolbar in Qlik Sense as well, which complicates it a little bit (the hover menu is coming from within sense-client while the selection toolbar uses this ActionsToolbar in nebula...)

PR #1206, opened 2023-03-31
fix: leave focus unchanged on hover

T-Wizard:
remove focus listbox on hover
and move the logic to global keydown to still work

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


DanielS-Qlik:
With this pr we might be able to remove `:not(:hover)'` from this line: https://github.com/qlik-oss/nebula.js/blob/a6cf04ffd05a61e5c9881620a0c7a4755cccbbfa/apis/nucleus/src/components/listbox/ListBoxInline.jsx#L50

PR #1207, opened 2023-03-31
fix(listbox): shift tab on container should take you to cell

johanlahti:
## Motivation

- fixes(https://github.com/qlik-oss/sn-list-objects/issues/239)

When focus is on a listbox container, the following should happen (implemented with this PR):
- `Shift + Tab` -> Focus on container's cell
- `Tab` -> "Do nothing" (let the event propagate and follow its native route)

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


apis/nucleus/src/components/listbox/interactions/listbox-keyboard-navigation.js
```diff
@@ -234,11 +234,18 @@ export function getListboxInlineKeyboardNavigation({
     }
 
     const container = event.currentTarget.closest('.listbox-container');
+    const inSelection = isModal();
 
     switch (keyCode) {
       case KEYS.TAB:
-        if (shiftKey) {
-          focusRow(container) || focusSearch(container);
+        if (inSelection) {
```
T-Wizard, line 241: I think
```
if (!inSelection) {
   return undefined;
}
```
would be a closer match to how the old filter pane work.
(but maybe that has to may problem from other differences?)

johanlahti: Yeah, that would be closer to how the existing Filter pane works. I discussed it with @pernyl quickly and we think that – for the case when there are no selections – tabbing (without shift) should do the opposite of tabbing with shift. In other words:
- Shift + Tab: Go to parent cell
- Tab: Go into listbox (search field or row)

I think that's an improvement compared to the existing Filter pane.

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -273,7 +273,7 @@ function ListBoxInline({ options, layout }) {
       <StyledGrid
         className="listbox-container"
         container
-        tabIndex={keyboard.outerTabStops ? 0 : -1}
+        tabIndex={-1}
```
T-Wizard, line 276: as this was the only use of outerTabStops, the definition can also be removed

apis/nucleus/src/components/listbox/interactions/listbox-keyboard-navigation.js
```diff
@@ -230,39 +230,41 @@ export function getListboxInlineKeyboardNavigation({
...
+          if (shiftKey) {
+            focusRow(container) || focusSearch(container);
+          } else {
+            focusSearch(container) || focusRow(container);
+          }
+        } else if (shiftKey) {
+          keyboard.blur(true);
```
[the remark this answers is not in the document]
johanlahti: Ok, what problem? Can we do something here to prevent that? (`keyboard.blur` go to the parent cell)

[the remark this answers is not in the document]
quanho: We can deliver this first and then we can fix the thing Tobias mentioned later
It is not so common to use keyboard and then switch to mouse.

PR #1208, opened 2023-03-31
feat(Listbox): tap range select

DanielS-Qlik:
## Motivation
Use two finger tap to select a range on touch devices.
fixes https://github.com/qlik-oss/sn-list-objects/issues/163

Single-column:
single-column

Grid-column:
grid-column

Grid-row:
grid-row

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


T-Wizard (approved):
Have you tested on a real touch device?

I think it will work if you don't move the fingers during the tap.
but I am unsure what happens if you do. (and if that is what we want)

DanielS-Qlik:
> Have you tested on a real touch device?
> 
> I think it will work if you don't move the fingers during the tap. but I am unsure what happens if you do. (and if that is what we want)

Yes I tested it on a iPad. I noticed that the old listbox seems to prevent gestures (like zoom) inside the listbox, so I added that as well. https://github.com/qlik-oss/nebula.js/pull/1208/commits/ffaf757fa5fb8a8948789b4e37074b7aaaa2159c

I have also tried to update the touch positions with `onTouchMove`, to get the latest positions if you move during tap.
But I only get the start position for some reason.

apis/nucleus/src/components/listbox/hooks/selections/useSelectionsInteractions.js
```diff
@@ -16,6 +23,25 @@ export default function useSelectionsInteractions({ selectionState, selections,
     touchRangeSmall: false,
   });
 
+  useEffect(() => {
+    if (!loaderRef.current?._listRef?._outerRef) {
```
T-Wizard, line 27: what is this for?

apis/nucleus/src/components/listbox/hooks/selections/useSelectionsInteractions.js
```diff
@@ -16,6 +23,25 @@ export default function useSelectionsInteractions({ selectionState, selections,
...
+    const listRef = loaderRef.current._listRef._outerRef;
+    listRef.addEventListener('gesturestart', preventGestureStart);
+    listRef.addEventListener('gesturechange', preventGestureChange);
+    listRef.addEventListener('gestureend', preventGestureEnd);
+    return () => {
+      listRef.removeEventListener('gesturestart', preventGestureStart);
+      listRef.removeEventListener('gesturechange', preventGestureChange);
+      listRef.removeEventListener('gestureend', preventGestureEnd);
+    };
+  }, [loaderRef.current?._listRef?._outerRef]);
```
T-Wizard, line 43: this can maybe be a problem as 
loaderRef.current?._listRef?._outerRef could probably without updating the effect
if the update do not also trigger a rerender of the listbox

PR #1213, opened 2023-04-04
fix(listbox): store last focused item

johanlahti:
## Motivation

- fixes(https://github.com/qlik-oss/sn-list-objects/issues/243)

Labelling the last selected row with a CSS class was not working.

The approach in this PR is more solid:

1. Add an `onFocus` listener to rows so we can store the last focused value
2. Pick up this value whenever we need to focus on a row and call `.focus()` on it

While doing this, all relevant methods for focusing were moved into the listbox's own keyboard hook, where we can also keep the required scope for these operations (e.g. `getStoreValue` and `containerRef`).

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


apis/nucleus/src/components/listbox/hooks/useDataStore.js
```diff
@@ -1,12 +1,18 @@
...
 
   const [vizDataStore] = useVizDataStore();
 
-  const setStoreValue = (key, val) => vizDataStore.set(`${keyPrefix}/${key}`, val);
-  const getStoreValue = (key) => vizDataStore.get(`${keyPrefix}/${key}`);
+  const setStoreValue = useCallback(
+    (key, val) => {
+      vizDataStore.set(`${keyPrefix}/${key}`, val);
+    },
+    [vizDataStore]
```
T-Wizard, line 13: also include keyPrefix as dependency here and for getStoreValue
```suggestion
    [vizDataStore, keyPrefix]
```
(just to be extra safe, it will probably never change and therefore also work without)

PR #1221, opened 2023-04-13
fix: fix rtl for histogram and checkbox and frequency count

quanho:
## Motivation

To fix https://github.com/qlik-oss/sn-list-objects/issues/193 and positioning of frequency count and row value when it is a number.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


T-Wizard (approved):
Maybe also a add rtl rendering test

PR #1222, opened 2023-04-14
fix: add right divider to listbox last column items

CostInCode:
## Motivation

Add a right divider to all the items of the last column.
Not adding it to all last items of each row, as in that case will be mis-aligned with the left dividers of previous items in the same column.
Resolves https://github.com/qlik-oss/sn-list-objects/issues/233

Screenshot with latest fix: 
Screenshot 2023-04-14 at 13 51 49

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/components/ListBoxRowColumn/ListBoxRowColumn.jsx
+++ b/apis/nucleus/src/components/listbox/components/ListBoxRowColumn/ListBoxRowColumn.jsx
@@ -169,6 +169,7 @@ function RowColumn({ index, rowIndex, columnIndex, style, data }) {
     justifyContent: valueTextAlign,
   };
 
+  const isLastCell = cellIndex + 1 === pages['0']?.qMatrix?.length;
   const isFirstElement = index === 0;
   const flexBasisVal = checkboxes ? 'auto' : 'max-content';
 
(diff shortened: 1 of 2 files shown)


T-Wizard:
This looks odd. I expect us to want it for all items that are last in there row. not just the very last item.
(In the image that would be X,Y,Z and not just Z)

CostInCode:
> This looks odd. I expect us to want it for all items that are last in there row. not just the very last item. (In the image that would be X,Y,Z and not just Z)

Right, will change this way, all the items on the last column

johanlahti (approved):

veinfors:
> > This looks odd. I expect us to want it for all items that are last in there row. not just the very last item. (In the image that would be X,Y,Z and not just Z)
> 
> Right, will change this way. Edit: adding divider to the items of last column, see description

Can you update the image in the description to show how it looks like with that fixed as well?

CostInCode:
> Can you update the image in the description to show how it looks like with that fixed as well?

The image is updated with the latest fix (so the X is not getting the divider, just the items of last column, Y and Z)

veinfors:
> > Can you update the image in the description to show how it looks like with that fixed as well?
> 
> The image in the description is updated with the latest fix (so the X is not getting the divider, just the items of last column, Y and Z)
> 
> If I add a right border also to X, will look bad as below: Screenshot 2023-04-14 at 14 49 26

Okey... No, that doesn't look good either, I would expect it to look like this ⬆️, but being correctly aligned (the last cell should have the same width as the other cells in that column that is)

CostInCode:
closing as it needs further alignment with UX

PR #1225, opened 2023-04-14
feat: add browser, and preferBuiltins cli args

vcellu:
## Motivation
In some scenarios, integrating third-party packages require no source maps, browser to be set to true and preferBuiltins set to false.
This PR creates the following flags

- disableSourceMaps; will prevent the sourcemaps from being generated
- browser; boolean
- preferBuiltins; boolean

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/commands/build/README.md
+++ b/commands/build/README.md
@@ -26,6 +26,8 @@ Options:
                                  [string] [choices: "production", "development"]
   --core           Set a core build target            [string] [default: "core"]
   --typescript     Enable typescript bundling         [boolean] [default: false]
+  --preferBuiltins Configuration option that specifies whether to use Node.js built-in modules [boolean] [default: false]
+  --browser        Configuration option that specifies whether the bundle is intended to run in a browser environment.  [boolean] [default: false]
 ```
 
 ## Example
--- a/commands/build/README.md
+++ b/commands/build/README.md
@@ -27,7 +27,7 @@ Options:
   --core           Set a core build target            [string] [default: "core"]
   --typescript     Enable typescript bundling         [boolean] [default: false]
   --preferBuiltins Configuration option that specifies whether to use Node.js built-in modules [boolean] [default: true]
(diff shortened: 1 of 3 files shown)


commands/build/lib/build.js
```diff
@@ -86,7 +86,7 @@ const config = ({
   let pkg = require(path.resolve(CWD, 'package.json')); // eslint-disable-line
   const corePkg = core ? require(path.resolve(core, 'package.json')) : null; // eslint-disable-line
   pkg = reactNative ? require(path.resolve(reactNativePath, 'package.json')) : pkg; // eslint-disable-line
-  const { sourcemap, replacementStrings = {}, typescript } = argv;
+  const { sourcemap, replacementStrings = {}, typescript, disableSourceMaps, preferBuiltins, browser } = argv;
```
veinfors, line 89: Can you add these options and descriptions/explanations in the README.md file as well?
And in `commands/build/lib/init-config.js` as well.

T-Wizard, line 89: why is there a need for both a sourcemap option and a disableSourceMaps option?

vcellu: Sorry, my bad, it was test code.  I cleaned it up.

vcellu: Done

commands/build/lib/build.js
```diff
@@ -135,6 +135,8 @@ const config = ({
         }),
         nodeResolve({
           extensions,
+          browser,
+          preferBuiltins,
```
veinfors, line 139: I see that the default value for `preferBuiltins` is `true` for `@rollup/plugin-node-resolve`. So having it default to `false` in nebula build would alter todays behaviour.

I suggest you change it's default value `true` to avoid changing the default build behaviour.

veinfors (approved):

PR #1226, opened 2023-04-17
feat: add filterpane handler for creating filterpane objects

T-Wizard:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/object/create-object.js
+++ b/apis/nucleus/src/object/create-object.js
@@ -20,17 +20,19 @@ export default async function createObject(
   generateOnly
 ) {
   let mergedProps = {};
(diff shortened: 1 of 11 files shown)


apis/nucleus/src/object/create-object.js
```diff
@@ -31,6 +32,7 @@ export default async function createObject(
           sn,
           properties: mergedProps,
           fields,
+          children,
```
veinfors, line 35: I think you should await the `populateData` call now that it is async.

quanho (approved):

veinfors (approved):

PR #1233, opened 2023-04-21
fix: prevent several text lines by using nowrap instead of pre

johanlahti:
## Motivation

`white-space: pre` was used to prevent white space when highlighting text. Changing this to `nowrap` doesn't seem to create any problems anymore, since this component has been refactored after `pre` was introduced.

### After (with search):
image

### After (text with line breaks):
image

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


quanho (approved):

apis/nucleus/src/components/listbox/components/ListBoxRowColumn/components/ListBoxRoot.jsx
```diff
@@ -78,7 +78,7 @@ const RowColRoot = styled('div', {
     userSelect: 'none',
     paddingRight: '1px',
     ...ellipsis,
-    whiteSpace: 'pre', // to keep white-space on highlight
+    whiteSpace: 'nowrap',
```
T-Wizard, line 81: Look it can be removed ('nowrap' already included in ellipsis)

T-Wizard (approved):

PR #1242, opened 2023-04-25
fix: correctly await types in serve

Caele:
## Motivation

Fixes so that serve awaits added types before running.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/commands/serve/web/components/Visualize/Visualize.jsx
+++ b/commands/serve/web/components/Visualize/Visualize.jsx
@@ -80,6 +80,7 @@ export default function Visualize() {
   const [currentId, setCurrentId] = useState();
   const [themeChooserAnchorEl, setThemeChooserAnchorEl] = React.useState(null);
   const [languageChooserAnchorEl, setLanguageChooserAnchorEl] = React.useState(null);
+  const [initialized, setInit] = useState(false);
   const [nebbie, setNebbie] = useState(null);
   const customThemes = Array.isArray(info?.themes) && info.themes.length ? ['light', 'dark', ...info.themes] : [];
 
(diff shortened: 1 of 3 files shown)


commands/serve/web/components/Visualize/Visualize.jsx
```diff
@@ -97,11 +98,16 @@ export default function Visualize() {
...
+      if (info) {
+        await initiateWatch(info);
+        setInit(true);
+      }
+    })();
   }, [info]);
 
   useEffect(() => {
-    if (waiting) return;
+    if (waiting || !initialized) return;
```
linhnguyen-qlik, line 110: Do we need to add `initialized` to dependency list of the useEffect?

linhnguyen-qlik (approved):
Nice, thank you for fixing this :). LGTM I have only a small question.

commands/serve/web/components/Visualize/Visualize.jsx
Caele: we do yes, good catch

commands/serve/web/hot.js
```diff
@@ -62,4 +55,13 @@ export default function initiateWatch(info) {
...
   update();
+  if (info.types) {
+    const proms = info.types.map((t) => {
+      return getModule(t.name, t.url).then((mo) => {
+        window[t.name] = mo;
+      });
+    });
+    return Promise.all(proms);
+  }
+  return Promise.resolve();
```
T-Wizard, line 66: [optional]
alternative you could set window[t.name] to a promise
```
window[t.name] = getModule(t.name, t.url);
```
then you would not need to wait here
and I think that would be closer to how it work in a real mashup

PR #1228, opened 2023-04-19
feat: listbox properties spec

T-Wizard:
Start of ListboxProperties spec

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`
    - [ ] No changes **_OR_** API changes has been formally approved
- [ ] Unit/Component test coverage
- [ ] Correct PR title for the changes (fix, chore, feat)

When build and tests have passed:

- [ ] Add code reviewers, for example @qlik-oss/nebula-core

PR #1245, opened 2023-04-26
feat: listbox properties spec

T-Wizard:
Alternate to #1228 with extracted "@nebula.js/stardust:listbox" api

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/nucleus/src/components/listbox/default-properties.js
```diff
@@ -1,7 +1,32 @@
+/**
```
Caele, line 1: This file should be used by the listbox right?

T-Wizard: is is used by listbox code in nebula.js in useOnTheFlyModel.jsx and filterpane-handler.js

PR #1257, opened 2023-05-03
fix: fix selection close when there are many objects with the same model

quanho:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


veinfors:
The change looks good to me. We just need to verify that it solves the problem we want and do some manual regression testing before we can merge it (+ add unit tests)

quanho:
Moved to https://github.com/qlik-oss/nebula.js/pull/1321 because of rebase issue.

PR #1258, opened 2023-05-04
fix: correct usage of event-emitter

Caele:
## Motivation

Event emitter didn't actually have an off function, so switching to add/remove instead.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


eliseeborn (approved):
👍

PR #1269, opened 2023-05-15
chore: remove storybook

Caele:
## Motivation

We don't really use the storybook, so rather remove it to keep the alerts cleaner.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


LiKang6688 (commented):
Why do you rename .storybook to .storybook_old? Is it intentional to remain `__stories__` folders?

LiKang6688 (approved):

veinfors (approved):

PR #1272, opened 2023-05-16
feat: add Imperaitive Handle to public API

Caele:
## Motivation

Exposes the Imperative handle API in the form of
- useImperativeHandle component hook
- getimperativeHandle viz API function

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/supernova/src/hooks.js
+++ b/apis/supernova/src/hooks.js
@@ -983,15 +983,11 @@ export function useOptions() {
 }
 
 /**
- * TODO before making public - expose getImperativeHandle on Viz
- * Exposes an API to the external environment.
- *
  * This is an empty object by default, but enables you to provide a custom API of your visualization to
  * make it possible to control after it has been rendered.
  *
  * You can only use this hook once, calling it more than once is considered an error.
  * @entry
- * @private
  * @template T
  * @param {function():T} factory
  * @param {Array<any>=} deps
--- a/apis/nucleus/src/__tests__/viz.test.js
+++ b/apis/nucleus/src/__tests__/viz.test.js
@@ -20,6 +20,7 @@ describe('viz', () => {
   let setSnPlugins;
   let takeSnapshot;
   let exportImage;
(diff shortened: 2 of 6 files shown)


apis/supernova/src/hooks.js
```diff
@@ -983,15 +983,11 @@ export function useOptions() {
...
- * Exposes an API to the external environment.
- *
  * This is an empty object by default, but enables you to provide a custom API of your visualization to
  * make it possible to control after it has been rendered.
  *
  * You can only use this hook once, calling it more than once is considered an error.
  * @entry
- * @private
  * @template T
  * @param {function():T} factory
```
T-Wizard, line 992: maybe simplify to
```suggestion
 * @param {function():object} factory
```
The generic type T don't give any value

veinfors:
Looks good to me.
A good idea is to test this change with some existing charts where the `useImperativeHandle` hook is used already, like the sn-bar-chart, sn-pie-chart etc. to make sure we don't break anything for existing extensions.

apis/supernova/src/hooks.js
Caele: hmm, you could have the handle function return whatever you want, shouldn't it be the generic type then? Not sure how a TS setup handles it?

Caele:
> Looks good to me. A good idea is to test this change with some existing charts where the `useImperativeHandle` hook is used already, like the sn-bar-chart, sn-pie-chart etc. to make sure we don't break anything for existing extensions.

There shouldn't be any functional changes, only typings are added and the API exposed to mashups.

veinfors (approved):

T-Wizard (approved):

apis/supernova/src/hooks.js
T-Wizard: as this is the only place where the type is referenced I don't think it gives any value.
(but I also don't there is any real problem with keeping it as it is)

PR #1274, opened 2023-05-17
chore: switch to codeclimate

Caele:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,11 +103,9 @@ jobs:
       - run:
           name: Run unit tests and publish to codeclimate
           command: |
-            curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
-            chmod +x ./cc-test-reporter
-            ./cc-test-reporter before-build
+            ./test/bin/cc-test-reporter before-build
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -101,13 +101,15 @@ jobs:
...
       - run:
-          name: Unit Tests
+          name: Run unit tests and publish to codeclimate
           command: |
-            yarn test:unit --coverage
-            if [ ! -z "$COVERALLS_REPO_TOKEN" ]; then
-              echo "Uploading coverage results to coveralls.io..."
-              cat ./coverage/lcov.info | npx coveralls
-            fi
+            curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
```
niekvanstaveren, line 106: You should not download this without checking the checksum and signature. We had a security issue with this before. Other repos have stored this file in the system actually. Instead we can probably use this github action https://github.com/paambaati/codeclimate-action
@enell asked for permission from SSO and seems to be okay

Caele: I will include the binary

niekvanstaveren (approved):

PR #1278, opened 2023-05-22
chore: split the ci into parts

Caele:
## Motivation

Splits the CI into jobs to make it faster (hopefully) and easier to view

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,7 +95,7 @@ jobs:
           paths:
             - .
 
-  validate:
+  junit:
     <<: *defaults
     steps:
       - attach_workspace:
@@ -114,6 +114,18 @@ jobs:
       - store_test_results:
           path: coverage/junit
 
+  validate:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: ~/project
+      - run:
+          name: Locale
+          command: yarn run locale:verify
+      - run:
+          name: Lint
+          command: yarn run lint
+
   build:
     docker:
       - image: cimg/node:20.1.0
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
T-Wizard, line 1: maybe also a install dependencies job (checkout & yarn)
that build and api-governance can depend on (to remove duplicated work)

.circleci/config.yml
```diff
@@ -1,5 +1,10 @@
 version: 2.1
 
+defaults: &defaults
```
T-Wizard, line 3: where is this used? (or is it not?)

.circleci/config.yml
Caele: I did this initially, but ran into issues, maybe after i've resolved other things I can put it back in.
Not sure the api-gov actually needs the deps installed actually, i'll see.

.circleci/config.yml
```diff
@@ -72,34 +77,21 @@ commands:
...
             <<# parameters.build >> yarn run build <</ parameters.build >>
             <<# parameters.test >>  yarn run test:e2e <</ parameters.test >>
-
 jobs:
-  build:
+  validate:
     docker:
-      - image: cimg/node:20.2.0
+      - image: cimg/node:20.1.0
       - image: browserless/chrome:1-puppeteer-7.1.0
```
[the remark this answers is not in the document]
Caele: hmm, could likely use node-browsers instead, i'll try

.circleci/config.yml
T-Wizard: ok

T-Wizard (approved):

PR #1283, opened 2023-05-25
refactor: dynamic column width for frequency count

johanlahti:
## Motivation

Set the column width for frequency count based on how many characters the longest currently loaded value has.

Also, adapt glyph width estimation to `fontFamily` and `fontSize` provided by theme, if any.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


veinfors:
I haven't looked into detail in the code yet, but when you say "dynamically adjust the with based on the largest number currently loaded" I'm thinking that this solution would mean that the column with could potentially vary while scrolling in a listbox then? Doesn't that give a weird experience?

johanlahti:
Yes, that's a possible outcome of it, although I don't think it has to look that bad. I will try to find a good test case for it.

johanlahti:
I have tried what that will look like by simulating one large value in the pages. While it looks a little ugly when it happens (particularly if the value diff is large (e.g. 10 digits) we could make it happen only once, by always caching the largest value so far.

Anyhow, I made this PR https://github.com/qlik-oss/nebula.js/pull/1284 which uses a simpler approach with `flex-basis`. Not as pretty as if we had an explicit width – but a lot simpler and without the jumping side-effect and extra calculation overhead.

PR #1284, opened 2023-05-26
refactor: use percent instead of fixed width for frequency count

johanlahti:
## Motivation

Improve the space management for frequency count while keeping frequency numbers intact (fixing customer bugs):
- Hard-code frequency width for the percent case, since `100.0%` is the longest string
- Use `flex-basic: max-content` and `min-width: max-content` where appropriate (so far, I have decided to use this in grid mode only, since we need to handle the space more efficiently)
- For absolute mode, increase the width so it can handle 8-10 digits
- Ensure a `min-width` exists for the value label (which also accounts for checkboxes) or it may disappear in some edge cases
- *New*: In single column mode and absolute count mode, use a consistent column width *when the value "touches" the frequency* (right-aligned for LTR and left-aligned for RTL)

*Note:* When Engine supports `qMaxGlyphCount` for frequency, we can switch to use this instead.

#### Demo with LTR: Notice how right-aligned is handled differently in grid mode (filling the space) which is good for using the space but can look slightly off when the frequency values varies a lot in size, thereby causing curved columns.

https://github.com/qlik-oss/nebula.js/assets/5780544/0e29e483-0da7-4e7a-b9e7-134f560fc885

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


johanlahti:
#devbuild

apis/nucleus/src/components/listbox/utils/deduce-frequency-mode.js
```diff
@@ -1,5 +1,5 @@
 export default function deduceFrequencyMode(pages) {
-  const hasPercentSign = pages.some(({ qMatrix }) => qMatrix[0].some((item) => !!item.qFrequency?.match('%')));
+  const hasPercentSign = pages.some(({ qMatrix }) => qMatrix.some(([item]) => !!item.qFrequency?.match('%')));
```
T-Wizard, line 2: why this change?

apis/nucleus/src/components/listbox/components/ListBoxRowColumn/components/ListBoxRoot.jsx
```diff
@@ -165,12 +174,12 @@ const RowColRoot = styled('div', {
...
     ...ellipsis,
-    // Frequency mode 'percent' will never need a text wider than this string: "100.0%"
-    flex: `0 0 ${getFreqFlexBasis({ sizes, frequencyMode, isGridMode })}`,
+    flex: `0 0 ${getFreqFlexBasis({ sizes, frequencyMode, isGridMode, freqHitsValue })}`,
     minWidth: 'auto',
+    // Frequency mode 'percent' will never need a text wider than this string: "100.0%"
     maxWidth: frequencyMode === 'P' ? `${sizes.freqMinWidth}px` : getMaxFreqWidth({ sizes, isGridMode }),
     textAlign: direction === 'rtl' ? 'left' : 'right',
-    paddingLeft: '2px',
+    paddingLeft: direction === 'rtl' ? '10px' : 0,
```
T-Wizard: why is there extra 10px padding only for rtl?

apis/nucleus/src/components/listbox/utils/deduce-frequency-mode.js
```diff
@@ -0,0 +1,5 @@
+export default function deduceFrequencyMode(pages) {
+  const hasPercentSign = pages.some(({ qMatrix }) => qMatrix.some(([item]) => !!item.qFrequency?.match('%')));
+  const OTHER = '-'; // treat all other as '-' (change later if we need to identify other modes)
+  return hasPercentSign ? 'P' : OTHER;
```
T-Wizard, line 4: this could be wrong if all values in pages has no frequency as they are excluded values.
is that OK? out could that cause problems with scrolling?

johanlahti: 🤷 Like I said before, the only workaround I can think of would be to get frequency mode from properties. Should we do that instead?
At the same time, it is more of a **visual** bug than a **functional** bug. I think it should be ok given that it is an edge case and will go away once we have frequency glyph count in the Engine response.

T-Wizard (approved):

apis/nucleus/src/components/listbox/utils/deduce-frequency-mode.js
T-Wizard: It can also be seen when scrolling, but I agree that it is "more of a visual bug than a functional bug"
and we can probably live with it.

PR #1293, opened 2023-06-09
fix: screen reader search

CostInCode:
## Motivation

Add a screen reader text to announce the number of search results available as the user types in the search box.
Fix #QB-14244

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -105,6 +107,26 @@ export default function ListBox({
...
+          break;
+        default:
+          t = 'ScreenReader.ManySearchResults';
+          break;
+      }
+
+      const screenReaderText = translator.get(t, [listCount]);
+      liveInputRegionRef.current.textContent = screenReaderText;
+    }
+  };
```
johanlahti, line 128: If possible. move this function outside of ListBox. Preferably in its own file.

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -328,136 +330,141 @@ function ListBoxInline({ options, layout }) {
...
-                  <DrillDownIcon
-                    tabIndex={-1}
-                    title={translator.get('Listbox.DrillDown')}
-                    size="large"
-                    style={{ fontSize: '12px' }}
-                  />
-                )}
-              </Grid>
-            )}
+      <ListBoxInputContext.Provider value={liveInputRegionRef}>
```
johanlahti, line 333: This context provider currently has one purpose only, and that is to bring the screen reader text to this (higher) level in the DOM hierarchy. Maybe there is no other case where we need to do this currently, but you could still write it so it's easy to make it more generic, if we would need to send more properties upwards like this:

Instead of:
```
const liveInputRegionRef = useRef(null);
```

Write:
```
const listBoxContext = useRef(null);
const { liveInputRegionRef, futureProperty } = listBoxContext;   // `futureProperty` is just made up
```

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -219,6 +240,12 @@ export default function ListBox({
   const { listCount } = sizes;
   setStoreValue('listCount', listCount);
 
+  const inputText = getStoreValue(`inputText`);
```
johanlahti, line 243:
```suggestion
  const inputText = getStoreValue('inputText');
```

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -219,6 +240,12 @@ export default function ListBox({
...
+
+  useEffect(() => {
+    updateScreenReaderSearchText({ inputText, listCount });
+  }, [inputText]);
```
johanlahti, line 247: 🤔 
```suggestion
  }, [inputText, listCount]);
```

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -216,6 +220,7 @@ export default function ListBoxSearch({
         tabIndex: keyboard.innerTabStops ? 0 : -1,
         style: { textAlign: `${inpuTextAlign}` },
         'data-testid': 'search-input-field',
+        'aria-label': 'Search field',
```
johanlahti, line 223: Translation string needed.

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -105,6 +106,26 @@ export default function ListBox({
...
+          break;
+        default:
+          t = 'ScreenReader.ManySearchResults';
+          break;
+      }
+
+      const srText = translator.get(t, [listCount]);
+      setScreenReaderText(srText);
+    }
+  };
```
johanlahti, line 127: If you want to, move some of this function outside so you can write unit tests for it:
```
const getScreenReaderSearchText = ({ inputText, listCount }) => { … }
```
```
const srText = getScreenReaderSearchText({ inputText, listCount });
setScreenReaderText(srText);
```

apis/nucleus/src/components/listbox/components/ScreenReaders.jsx
```diff
@@ -66,3 +66,22 @@ export function ScreenReaderForSelections({ layout }) {
     </div>
   );
 }
+
+export function getScreenReaderSearchText({ inputText, listCount }) {
+  let t;
+  if (inputText) {
```
johanlahti, line 72: I think you should remove this. And if needed, move outside this func.

apis/nucleus/src/components/listbox/ListBox.jsx
```diff
@@ -219,6 +220,14 @@ export default function ListBox({
   const { listCount } = sizes;
   setStoreValue('listCount', listCount);
 
+  const inputText = getStoreValue('inputText');
+
+  useEffect(() => {
+    const srText = getScreenReaderSearchText({ inputText, listCount });
```
johanlahti, line 226:
```suggestion
    if (!inputText) {
      return;
    }
    const srText = getScreenReaderSearchText({ listCount });
```

johanlahti (approved):

PR #1297, opened 2023-06-20
adding context menu (Don't review)

linhnguyen-qlik:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/nucleus/src/components/Cell.jsx
```diff
@@ -320,6 +325,36 @@ const Cell = forwardRef(
...
+              // Other native context menus might behave different.
+              // With this behavior we prevent contextmenu from the backdrop to re-locale existing context menus.
+              initState
+        );
+      }
+    };
+
+    const closeContextMenu = () => {
+      setContextMenu(initState);
+    };
```
Caele, line 357: in general, try to move code out of Cell.jsx so it doesn't grow too much.
Also the "object" here should not be needed, should be something else.

linhnguyen-qlik: Yes. I have moved them out. This is just a quick test. Please don't review this for now since it not ready yet. This is one messy :).

PR #1298, opened 2023-06-21
feat: add menuBuilder

linhnguyen-qlik:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/supernova/src/hooks.js
+++ b/apis/supernova/src/hooks.js
@@ -164,9 +164,9 @@ export function runSnaps(component, layout) {
   return Promise.resolve();
 }
 
-export function runMenu(component, menu, event) {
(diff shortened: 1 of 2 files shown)


apis/supernova/src/hooks.js
```diff
@@ -164,9 +164,9 @@ export function runSnaps(component, layout) {
   return Promise.resolve();
 }
 
-export function runMenu(component, menu, event) {
+export function runMenu(component, menu, event, menuBuilder) {
   try {
-    return Promise.all(component.__hooks.menus.map((h) => Promise.resolve(h.fn(menu, event)))).then(
```
Caele, line 169: Maybe we can change this and the other to just use ...params instead.

T-Wizard (approved):

PR #1321, opened 2023-08-02
fix: fix selection close when there are many objects with the same model

quanho:
If there is a master filterpane and few collapsed filterpane objects are based on this master item, then everytime the user select an item in the listbox popover from a collapsed filterpane the listbox popover will be closed right away.
The problem is that currently we use one object selection for all filterpanes with the same model and then we check if a click is outside a filterpane or not and if it is then we trigger deactivated event which will close the listbox popover.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/hooks/useAppSelections.js
+++ b/apis/nucleus/src/hooks/useAppSelections.js
@@ -1,28 +1,22 @@
 /* eslint no-underscore-dangle: 0 */
 import { useEffect } from 'react';
-import {
-  useAppSelectionsStore,
-  objectSelectionsStore,
-  modalObjectStore,
-  appModalStore,
-} from '../stores/selections-store';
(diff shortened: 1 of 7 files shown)


apis/nucleus/src/hooks/useAppSelections.js
```diff
@@ -65,9 +59,11 @@ function createAppSelections({ app }) {
     isInModal() {
       return !!modalObjectStore.get(key);
     },
-    isModal(object) {
+    isModal(objectSelections) {
```
johanlahti, line 62: Maybe we should return `false` if `objectSelections` is `undefined`:
```
if (!objectSelections) {
  return false;
}
```

johanlahti (approved):
Unit tests? (and fix failing ones)

veinfors (approved):

PR #1327, opened 2023-08-09
chore: clean warnings in build

Caele:
## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


apis/nucleus/src/sn/type.js
```diff
@@ -1,6 +1,6 @@
 import { generator as SNFactory } from '@nebula.js/supernova';
-import { satisfies } from 'semver';
 import extend from 'extend';
+import satisfies from 'semver/functions/satisfies';
```
Caele, line 3: This was an attempt to solve the circular dependency warning detailed here: https://github.com/npm/node-semver/issues/381
Still get the warning however.

PR #1328, opened 2023-08-10
feat: implement auto confirm

johanlahti:
## Motivation

Automatically confirm selections instead of showing the actions toolbar, when `autoConfirm` is true.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -216,7 +216,7 @@ function ListBoxInline({ options, layout }) {
     }
   }, [searchContainer && searchContainer.current, showSearch, search, focusSearch]);
 
-  const { wildCardSearch, searchEnabled, layoutOptions = {} } = layout ?? {};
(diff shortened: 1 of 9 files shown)


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -216,7 +216,7 @@ function ListBoxInline({ options, layout }) {
     }
   }, [searchContainer && searchContainer.current, showSearch, search, focusSearch]);
 
-  const { wildCardSearch, searchEnabled, layoutOptions = {} } = layout ?? {};
+  const { wildCardSearch, searchEnabled, autoConfirm = false, layoutOptions = {} } = layout ?? {};
```
veinfors, line 219: This new setting should be added to the api spec as well

johanlahti: ✅

veinfors (approved):

PR #1329, opened 2023-08-10
fix: race condition for mocked charts

veinfors:
Solves a race condition that has occurred a couple of times when rendering charts using a mocked data model (no real engine), both when using EnigmaMocker and when "manually" mocking the engima app model.

I have no really good explanation to exactly how this solves the problem, but it does. And I see no big risk in introducing this extra await with the setTimeout into `useRpc`, since all of those backend calls are asynchronous anyway - they will always have some kind of delay in real life scenarios.

TODO:
- [x] Some more manual regression testing before merging

### DIFF
--- a/apis/enigma-mocker/src/from-generic-objects.js
+++ b/apis/enigma-mocker/src/from-generic-objects.js
@@ -13,6 +13,7 @@ export default function fromGenericObjects(genericObjects, options = {}) {
     id: `app - ${+Date.now()}`,
     session,
     createSessionObject,
+    destroySessionObject: async () => {},
     getObject,
     getAppLayout,
   };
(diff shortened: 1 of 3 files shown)


apis/enigma-mocker/src/from-generic-objects.js
```diff
@@ -13,6 +13,7 @@ export default function fromGenericObjects(genericObjects, options = {}) {
     id: `app - ${+Date.now()}`,
     session,
     createSessionObject,
+    destroySessionObject: async () => {},
```
veinfors, line 16: Add this empty function here to avoid warnings when using EnigmaMocker

Caele (approved):

PR #1335, opened 2023-08-16
feat: object conversion for readonly charts

quanho:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/viz.js
+++ b/apis/nucleus/src/viz.js
@@ -1,6 +1,7 @@
 /* eslint-disable no-underscore-dangle */
 import EventEmitter from 'node-event-emitter';
 import { convertTo as conversionConvertTo } from '@nebula.js/conversion';
+import { JSONPatch } from '@nebula.js/supernova';
 import glueCell from './components/glue';
 import getPatches from './utils/patcher';
 import validatePlugins from './plugins/plugins';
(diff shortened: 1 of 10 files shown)


apis/nucleus/src/viz.js
```diff
@@ -119,14 +122,20 @@ export default function viz({ model, halo, initialError, onDestroy = async () =>
      * viz.convertTo('barChart');
      */
     async convertTo(newType, forceUpdate = true) {
```
Caele, line 124: I think we'd want to update the API to be clearer as well, so that you more easily can pick between permanent conversion and session based.
+ make sure there is a test and example somewhere.

T-Wizard:
It will most likely not work if the object converted to (or from) has child objects.

quanho:
> It will most likely not work if the object converted to (or from) has child objects.
It is a limitation of nebula conversion.

Caele:
> It will most likely not work if the object converted to (or from) has child objects.

I added an error that gets thrown if you try to convert for example a filterpane

T-Wizard:
> > It will most likely not work if the object converted to (or from) has child objects.
> 
> I added an error that gets thrown if you try to convert for example a filterpane

hmm.
(why do your solution work in that case?)
-> Looked and sn-filter-pane like many other nebula charts only define export/import properties where sense-client look for it not where nebula.js looks for it. (some charts do define both)
(the sense and nebula.js versions also get arguments in different format)
If sn-filter-pane was updated to also specify conversion function for nebula.js you would get the problem case I was thinking of.
Then I would expect it to only work when forceUpdate=false as the update object cases (both of them) do not update child objects.

This is mostly to inform, and I do not expect this PR to do anything for the convert object with child objects case.

Caele:
> This is mostly to inform, and I do not expect this PR to do anything for the convert object with child objects case.

yea, I won't delive into this more for now. Just want to add a base level of protection for users of this. 
I do want to add a way to revert the concersion however, so a bit more tweaking of the API.

Caele (approved):

PR #1339, opened 2023-08-22
feat: enable listbox styling

CostInCode:
## New features:
- Apply styling of header, content and selection states (selected, alternative, excluded)
- This is accomplished using a new property called `components: object[]` which for now only allows you to override the **styling** of the following components: `header`, `content`, `palette` (for selection styling)

The `components` property could get expanded functionality in the future but currently only considering overriding styling.

## Example of using components
Here is how you can specify components to override styling of header, content and selections style:
```
[
    {
        "key": "header",
        "header": {
            "fontColor": {
                "index": -1,
                "color": "#19d956"
            },
            "fontSize": 23
        },
        "fontSize": 24,
        "fontColor": {
            "index": -1,
            "color": "#075900"
        }
    },
    {
        "key": "content",
        "content": {
            "fontSize": 18,
            "fontColor": {
                "index": -1,
                "color": "#ff2047"
            }
        },
        "fontColor": {
            "index": -1,
            "color": "#590c2a"
        },
        "fontSize": 14
    },
    {
        "key": "palette",
        "selected": {
            "index": -1,
            "color": "#4b6d98"
        },
        "alternative": {
            "index": -1,
            "color": "#71a5e6"
        },
        "excluded": {
            "index": -1,
            "color": "#8c8a8a"
        }
    }
]
```

### General component

And for the settings on the Filter pane level, use the general component:
```
{
        "key": "general",
        "title": {
            "main": {
                "fontSize": "16px",
                "fontStyle": [],
                "color": {
                    "index": -1,
                    "color": "#06ab00"
                }
            },
            "subTitle": {
                "fontSize": "14px",
                "color": {
                    "index": -1,
                    "color": "#8a138b"
                },
                "fontStyle": [
                    ""
                ]
            },
            "footer": {
                "color": {
                    "index": -1,
                    "color": "#ed7c4e"
                },
                "fontStyle": [
                    "underline",
                    "bold"
                ],
                "fontSize": "17px"
            }
        },
        "bgColor": {
            "color": {
                "index": 4,
                "color": "#d76c6c"
            }
        }
    },
```

## Motivation (old screen recordings)

Add listbox styling from filterpane styling panel

https://github.com/qlik-oss/nebula.js/assets/13185716/30e1f63a-8e51-4b32-9db4-e6193999469c

Screenshot 2023-09-08 at 07 32 36

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -325,6 +320,16 @@ function ListBoxInline({ options, layout }) {
     renderedCallback?.();
   }
 
+  const Title = styled(Typography)(() => {
```
T-Wizard, line 323: Don't move the definition of Title inside ListBoxInline.
It will make react think it always change and need to be re-mounted.
(you can send in the extra info in props instead)
(or maybe re-use the theme.listbox hack but I don't like it and think it should be removed)

apis/nucleus/src/components/listbox/components/grid-list-components/grid-list-components.jsx
```diff
@@ -103,7 +108,7 @@ export default function getListBoxComponents({
         height={listHeight}
         width={width}
         itemCount={listCount}
-        itemData={{ ...commonItemData, listCount }}
+        itemData={{ ...commonItemData, listCount, contentFontStyle }}
```
johanlahti, line 111: This is need for the `Grid` component too I suppose (i.e. the other component exported from this file).

apis/nucleus/src/components/listbox/hooks/use-style.js
```diff
@@ -0,0 +1,43 @@
...
+
+  for (let i = 0; i < steps.length; ++i) {
+    const step = steps[i];
+
+    if (scoped[step] === undefined) {
+      return fallback;
+    }
+
+    scoped = scoped[step];
+  }
```
johanlahti, line 18: If may be self-explanatory but you could add a short comment about what the function does, and why you return fallback on `undefined`.

apis/nucleus/src/components/listbox/hooks/use-style.js
```diff
@@ -0,0 +1,43 @@
+function getValue(object, path, fallback) {
+  if (object === undefined || path === undefined) {
+    return fallback;
+  }
+
+  const steps = path.split('.');
+
+  let scoped = object;
```
johanlahti, line 8:
```suggestion
  const scoped = {};
  
  // optional check before for-loop
  if (steps.includes(undefined)) {
    return fallback;
  }
```

apis/nucleus/src/components/listbox/hooks/use-style.js
johanlahti: Actually, could you use this instead?
```
utils.getValue(object, 'path', fallback);
```
```

Comment (veinfors) [reply apis/nucleus/src/components/listbox/hooks/use-style.js:18]: `utils.getValue` is in the `conversion` package, right? But I was thinking the same. Just just introduces this helper function one more time in nebula. Don't know the best/preferred way of sharing it between packages within nebula.js though...

Comment (johanlahti) [reply apis/nucleus/src/components/listbox/hooks/use-style.js:18]: Ok, yeah can check if it's possible to re-use somehow then from `api/conversion/src/utils.js` @CostInCode 
Otherwise, add it to a location where can easily import it so we don't duplicate it several times inside Nebula.

File: apis/stardust/package.json [hunk 5]
```diff
@@ -31,7 +31,7 @@
   ],
   "scripts": {
     "build": "cross-env NODE_ENV=production rollup --config ../../rollup.config.js",
-    "build:dev": "rollup --config ../../rollup.config.js",
+    "build:dev": "(rollup --config ../../rollup.config.js && say 'done') || say 'error compiling'",
```
T-Wizard, line 34: don't add any mac-os only commands (like `say`) it will break in other environments

PR #1340, opened 2023-08-24
fix: unresolving promises for missing types

Caele:
## Motivation

Fixes bugs with unresolved promises that stops the viz/sheet API to return properly.
Adds errors for incorrect elements.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/__tests__/viz.test.js
+++ b/apis/nucleus/src/__tests__/viz.test.js
@@ -23,6 +23,8 @@ describe('viz', () => {
   let getImperativeHandle;
   let convertToMock;
 
+  let mockElement;
+
   beforeAll(() => {
     unmountMock = jest.fn();
     setSnOptions = jest.fn();
@@ -66,6 +68,8 @@ describe('viz', () => {
       model,
       halo: { public: {} },
     });
+
+    mockElement = document.createElement('div');
   });
   afterAll(() => {
     jest.resetAllMocks();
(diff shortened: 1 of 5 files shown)


apis/nucleus/src/__tests__/viz.test.js
```diff
@@ -96,16 +100,24 @@ describe('viz', () => {
   });
 
   describe('mounting', () => {
+    test('show throw for invalid elements', () => {
+      try {
+        api.__DO_NOT_USE__.mount('just a string');
+      } catch (error) {
+        expect(error.message).toBe('Provided element is not a proper HTMLElement');
```
cbt1, line 107: If it does not throw the test will pass

apis/nucleus/src/components/Cell.jsx
```diff
@@ -384,7 +389,12 @@ const Cell = forwardRef(
     };
 
     useEffect(() => {
-      if (initialError || !appLayout || !layout) {
+      if (initialError) {
+        // To resolve the Viz promise for missing types
+        onMount();
```
cbt1, line 394: There is no coverage for this line. If this is suppose to fix the issue then consider adding coverage for it.

cbt1 (approved):
LGTM but I'm not entirely sure what the actual issue is or how the `onMount` would help.

PR #1346, opened 2023-08-28
feat: object conversion for readonly charts (#1335)

Caele:
## Motivation

I decided to make this simpler for now just add a parameter to the current api to support patching and keep the new API hidden.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/viz.js
+++ b/apis/nucleus/src/viz.js
@@ -109,11 +109,12 @@ export default function viz({ model, halo, initialError, onDestroy = async () =>
       unmountCell = noopi;
     },
     /**
-     * Converts the visualization to a different registered type. Will update properties if permissions allow, else will patch.
+     * Converts the visualization to a different registered type.
+     *
+     * Will update properties if permissions allow, else will patch (can be forced with forcePatch parameter)
      *
      * Not all chart types are compatible, similar structures are required.
      *
-     * NOTE: Consider using viz.convert.toType instead for session based conversion
      * @since 1.1.0
      * @param {string} newType - Which registered type to convert to.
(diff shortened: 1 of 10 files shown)


apis/nucleus/src/viz.js
```diff
@@ -106,27 +109,43 @@ export default function viz({ model, halo, initialError, onDestroy = async () =>
       unmountCell = noopi;
     },
     /**
-     * Converts the visualization to a different registered type
+     * Converts the visualization to a different registered type. Will update properties if permissions allow, else will patch.
+     *
+     * Not all chart types are compatible, similar structures are required.
+     *
+     * NOTE: Consider using viz.convert.toType instead for session based conversion
```
veinfors, line 116: Maybe we should wait with referencing `viz.convert` until it's part of the api?
Also would be nice to clarify why the user should consider using that instead... some specific benefit?

apis/supernova/src/index.js
```diff
@@ -1,6 +1,7 @@
 import generator from './generator';
+import JSONPatch from './json-patch';
 
-export { generator };
+export { generator, JSONPatch };
```
veinfors, line 4: Feels a little bit weird, the way we import and then export generator and and JSONPatch like this. Maybe we should consider creating some kind of common internal pkg instead for shared stuff, but we don't have to do it right now.

veinfors (approved):

apis/nucleus/src/viz.js
Caele: yes yes, i'll remove that part

apis/supernova/src/index.js
Caele: it is a bit weird, but supernova is an internal package, so no too far off.

PR #1359, opened 2023-09-21
docs: add badges

niekvanstaveren:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


niekvanstaveren:
Codeclimate was actually already running in circleci. However the reporter id that was added in the configuration was wrong. I updated and it is now pushing correctly. Adding only badges here

PR #1376, opened 2023-10-16
feat: toggle data view

niekvanstaveren:
## Motivation

Adds support for toggleDataView. Needed to provide an accessible way to the data in Nebula mashups. 

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/index.js
+++ b/apis/nucleus/src/index.js
@@ -550,7 +550,7 @@ function nuked(configuration = {}) {
       /**
        * Check if certain visualization type is registered
        * @function
-       * @param {string=} type Type to check
+       * @param {string} type Type to check
        * @returns {boolean}
        * @example
        * const isRegistered = n.isRegistered("sn-table")
--- a/apis/nucleus/src/index.js
+++ b/apis/nucleus/src/index.js
@@ -547,6 +547,15 @@ function nuked(configuration = {}) {
        * //]
        */
       getRegisteredTypes: types.getList,
+      /**
+       * Check if certain visualization type is registered
+       * @function
+       * @param {string=} type Type to check
+       * @returns {boolean}
+       * @example
+       * const isRegistered = n.isRegistered("sn-table")
+       */
+      isRegistered: types.isRegistered,
       __DO_NOT_USE__: {
         types,
         popover(anchorElement, fieldIdentifier, options) {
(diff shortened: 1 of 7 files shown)


apis/nucleus/src/viz.js
```diff
@@ -149,6 +150,51 @@ export default function viz({ model, halo, initialError, onDestroy = async () =>
       const propertyTree = await conversionConvertTo({ halo, model, cellRef, newType });
       return propertyTree;
     },
+    /**
+     * Toggles the view of the chart into an accessible table
```
veinfors, line 154: Maybe explain that the type you want to toggle to, sn-table by default, needs to be registered on the embed instance to be able to do this?

niekvanstaveren: Added some more info and also a check if the type is actually registered

apis/nucleus/src/viz.js
```diff
@@ -149,6 +150,44 @@ export default function viz({ model, halo, initialError, onDestroy = async () =>
       const propertyTree = await conversionConvertTo({ halo, model, cellRef, newType });
       return propertyTree;
     },
+    /**
+     * Toggles the view of the chart into an accessible table. The type (default sn-table) needs to be registered in order to use this functionality
+     *
+     * @since 4.6.0
```
niekvanstaveren, line 156: Should it maybe start as experimental?

apis/nucleus/src/index.js
```diff
@@ -547,6 +547,15 @@ function nuked(configuration = {}) {
        * //]
        */
       getRegisteredTypes: types.getList,
+      /**
+       * Check if certain visualization type is registered
+       * @function
+       * @param {string=} type Type to check
```
veinfors, line 553:
```suggestion
       * @param {string} type Type to check
```

apis/nucleus/src/viz.js
```diff
@@ -149,6 +150,44 @@ export default function viz({ model, halo, initialError, onDestroy = async () =>
...
+     * @param {string} newType - The type used for the view data table
```
veinfors, line 157:
```suggestion
     * @param {string=} [newType=sn-table] The type used for the view data table
```

apis/nucleus/src/viz.js
veinfors: Sure, we could start by having it experimental...

veinfors (approved):

Caele:
Some notes:
- Need to decide on the API itself: viz.toggleDataView(type)  has a few options to it how it should look, and hard to change it later. It is tied to conversion so might make sense to place it there viz.convert.toggleDataView . Also is a question if it really should be a toggle or not.
- The created session objects are currently not reused nor destroyed, so new ones are constantly added.

PR #1386, opened 2023-10-24
fix: handle sync calls to session storage

Caele:
## Motivation

Fixes an issue where the alternate states wasn't displayed at all in current selections. 
Instead of using a shared model there where two current-selections session objects, with only one getting the alternate states patch. This looks to be due to a timing issue where the two calls to useCurrentSelectionsModel was done too close to eachother thus not having time to store the rpcCall. This change introduces a sync storage of the call promise instead of waiting fo the actual call to be made.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/hooks/useSessionModel.js
+++ b/apis/nucleus/src/hooks/useSessionModel.js
@@ -7,21 +7,15 @@ export default function useSessionModel(definition, app, ...deps) {
   const [rpcRequestSessionModelStore] = useRpcRequestSessionModelStore();
   const [model, setModel] = useState();
 
-  let rpcShared;
-
-  if (key) {
-    rpcShared = rpcRequestSessionModelStore.get(key);
-  }
-
   useEffect(() => {
     if (!app) {
       return;
     }
     // Create new session object
     const create = async () => {
+      let rpcShared = rpcRequestSessionModelStore.get(key);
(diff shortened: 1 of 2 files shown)


quanho (approved):

apis/nucleus/src/hooks/useSessionModel.js
```diff
@@ -19,10 +29,11 @@ export default function useSessionModel(definition, app, ...deps) {
...
     // Create new session object
     const create = async () => {
-      if (!rpcShared) {
-        const rpc = app.createSessionObject(definition);
-        rpcShared = rpc;
+      if (allocatedResolve) {
+        rpcShared = app.createSessionObject(definition);
         rpcRequestSessionModelStore.set(key, rpcShared);
+        const resolvedModel = await rpcShared;
+        allocatedResolve(resolvedModel);
```
T-Wizard, line 36: I think it would be a better solution to move `rpcRequestSessionModelStore.get(key);` inside the create function instead

if not you should set allocatedResolve to undefined/null when used to not create extra session objects if create run again for an old object

Caele: yea, that is more appropriate and simpler, good

T-Wizard (approved):

PR #1389, opened 2023-10-27
fix: handle unsupported colors and add tests

johanlahti:
## Motivation

The mui function `getContrastRatio` does not support named colors ('red', 'transparent' etc.) and will throw an error if sent in. Therefore, we must convert these colors first and as an extra safety net, wrap the function in a try-catch to be sure to prevent misspelled colors from throwing an error.

Note! This issue blocks the nebula release and bump.

### After fix:

Selections are possible and do not throw.

https://github.com/qlik-oss/nebula.js/assets/5780544/2622b808-5993-4285-aea3-a4422840c4cb

I have also tried to change styling and themes etc.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/assets/styling.js
+++ b/apis/nucleus/src/components/listbox/assets/styling.js
@@ -10,6 +10,7 @@ const TRANSPARENT = 'rgba(255, 255, 255, 0)';
  *
  * @param {string} c The CSS color.
  * @returns {string} Converted color or same color as before.
+ * @private
  * @example
  *  convertNamedColor('red') => '#FF0000'
  */
--- a/apis/nucleus/src/components/listbox/assets/styling.js
+++ b/apis/nucleus/src/components/listbox/assets/styling.js
@@ -39,6 +39,7 @@ export const hasEnoughContrast = (desiredTextColor, backgroundColor) => {
   } catch (err) {
     // The function throws on unsupported or misspelled colors.
     // In these cases, simply return true.
+    isContrastingEnough = true;
   }
   return isContrastingEnough;
 };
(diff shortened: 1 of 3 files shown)


apis/stardust/api-spec/spec.json
```diff
@@ -1829,6 +1829,24 @@
...
+        }
+      ],
+      "returns": {
+        "description": "Converted color or same color as before.",
+        "type": "string"
+      },
+      "examples": [
+        "convertNamedColor('red') => '#FF0000'"
+      ]
+    },
```
johanlahti, line 1849: @veinfors Do you know why it's picking this up in `yarn spec`? How can I prevent this, by adding a JSDoc comment?

johanlahti: `@private` added to hide the func.

apis/nucleus/src/components/listbox/assets/styling.js
```diff
@@ -1,15 +1,54 @@
...
+  const background = convertNamedColor(backgroundColor);
 
-function getContrastingColor(backgroundColor, desiredTextColor = undefined, dark = DARK, light = LIGHT) {
   const CONTRAST_THRESHOLD = 3.0;
+  let isContrastingEnough;
+  try {
+    isContrastingEnough = getContrastRatio(desired, background) > CONTRAST_THRESHOLD;
+  } catch (err) {
+    // The function throws on unsupported or misspelled colors.
+    // In these cases, simply return true.
```
tasseKATT, line 41: Do you need to do something here?

Caele (approved):

apis/nucleus/src/components/listbox/assets/styling.js
johanlahti: 🤔 Yeah, I'm thinking whether to return `true`, `false` or `undefined` in this case. `true` implies the same color as was sent in, so I should probably default to that.

PR #1395, opened 2023-10-30
feat: toggleDataView

niekvanstaveren:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/stardust/types/index.d.ts
```diff
@@ -411,6 +412,12 @@ declare namespace stardust {
          */
         convertTo(newType: string, forceUpdate?: boolean, forcePatch?: boolean): Promise<object>;
 
+        /**
+         * Toggles the chart data view
+         * @param showDataView If included, turns the toggle into a one way-only operation. If true it will only toggle to the view data table.
```
niekvanstaveren, line 417: Could use a better description

apis/nucleus/src/viz.js
```diff
@@ -149,6 +150,46 @@ export default function viz({ model, halo, initialError, onDestroy = async () =>
...
+        newModel = await halo.app.createSessionObject(newProperties);
+        viewDataObjectId = newModel.id;
+      } else if (viewDataObjectId && showDataView !== true) {
+        newModel = model;
+        await halo.app.destroySessionObject(viewDataObjectId);
+        viewDataObjectId = undefined;
+      }
+      if (newModel) {
+        await unmountCell();
+        [unmountCell, cellRef] = glueCell({
```
niekvanstaveren, line 181: This actually creates a cell every time. Would there be any downside to this?

apis/conversion/src/index.js
```diff
@@ -23,7 +23,10 @@ const getDefaultExportPropertiesFn = (path) => {
   return undefined; // TODO: add listbox and other
 };
 
-const getExportPropertiesFnc = (qae) => {
+const getExportPropertiesFnc = (qae, convertToTable) => {
+  if (convertToTable && typeof qae.exportTableProperties === 'function') {
+    return qae.exportTableProperties;
```
niekvanstaveren, line 28: This needs to be exported by the individual charts. For histogram and boxplot this could look like this:
https://github.com/qlik-oss/dist-flow/pull/319

apis/stardust/api-spec/spec.json
```diff
@@ -720,6 +720,11 @@
           "optional": true,
           "defaultValue": false,
           "type": "boolean"
+        },
+        "dataViewType": {
```
cbt1, line 724: Maybe want to explicitly state that the type should be an `sn-table` but might be registered under a different name? I guess they techincally could set any object as chart here but since we pass in the specific `sn-table` properties when creating the view data object I assume that is not really supported?

apis/nucleus/src/__tests__/viz.test.js
```diff
@@ -232,6 +240,26 @@ describe('viz', () => {
     });
   });
 
+  describe('toggleDataView', () => {
+    test('should toggle data view and back', async () => {
```
cbt1, line 244: This would pass if `api.toggleDataView()` did not actually convert it to type `sn-table`. So consider adding a check for that as well.

apis/stardust/api-spec/spec.json
niekvanstaveren: This can be overwritten when you set this in your initial config. So if you do not register something as `sn-table`, but just as table you would need to add that here as well

cbt1: I'm just saying that we want to clearly document what the expections are. The object should be an `sn-table` but might be registered as something else, ex "table". Also assuming that the consumer of the `toggleDataView` API must have registered the `sn-table` (or by some other name) for it to work. Which is also something that should be documented IMO

niekvanstaveren: Got it. I added some more explanation now, hopefully that makes since more clear

apis/nucleus/src/viz.js
Caele: Compared to conversion you create a new model, so it won't automagically update the current Cell. 
But what you could do is replace what model the Cell uses. Either you'd have to restructure how the React stuff is setup OR possibly add a setModel function to the importativeHandle (cellRef) and replace it that way

niekvanstaveren: So went with the second option of adding a setModel function. Not sure how we would restructure the React setup. The state will need to be handled inside the React parts if we want to keep it as state. Otherwise I think we still have to setup the portal every single time we toggle

apis/stardust/types/index.d.ts
haxxmaxx: I think the wording `true it will only toggle to the view data table` is a bit misleading. You wont toggle to data view if it is already in data view. So maybe something like this
```suggestion
         * @param showDataView If included, forces the chart into the specified state. True will show data view, and false will show the original chart.
```

haxxmaxx: Also, we don't really mention in the rest of the description that if you call the function again, it will toggle back to the original chart

apis/nucleus/src/viz.js
```diff
@@ -149,6 +150,41 @@ export default function viz({ model, halo, initialError, onDestroy = async () =>
...
+     *
+     * The chart will be toggled to the type defined in the nebula context (dataViewType).
+     *
+     * The default dataViewType for nebula is sn-table. The specified chart type needs to be registered as well in to make it possible to render the data view.
+     *
+     * @experimental
+     * @since 4.7.0
+     * @param {boolean=} showDataView - If included, turns the toggle into a one way-only operation. If true it will only toggle to the view data table.
+     */
+    async toggleDataView(showDataView) {
```
haxxmaxx, line 164: clever logic in here 👍

apis/stardust/types/index.d.ts
niekvanstaveren: Yes that is a lot better, thanks!

apis/conversion/src/index.js
T-Wizard, line 28: consider change names to ex:
convertToTable -> convertToViewData
exportTableProperties -> exportViewDataProperties
as it is always for viewData but could be someting that is not a table if dataViewType is changed to something else

apis/nucleus/src/viz.js
```diff
@@ -149,6 +150,41 @@ export default function viz({ model, halo, initialError, onDestroy = async () =>
...
+        const oldProperties = await model.getEffectiveProperties();
+        const propertyTree = await conversionConvertTo({
+          halo,
+          model,
+          cellRef,
+          newType: halo.config.context.dataViewType,
+          properties: oldProperties,
+          convertToTable: true,
+        });
+        const newProperties = { ...propertyTree.qProperty, totals: { show: false }, usePagination: true };
```
T-Wizard, line 176: feels questionable to add sn-table specific properties here.
(but I am unsure where to put it instead, maybe in a importViewDataProperties? in sn-table? or dataViewExtraProperties setting)

apis/conversion/src/index.js
niekvanstaveren: The first one I think we can easily change. The second one is consistent with the current setup. It is already called exportTableProperties and it is only used for view data. I agree with your reasoning, but I think in that case we should also update it in the client properly. https://github.com/qlik-trial/sense-client/blob/main/web/assets/objects/utils/properties-generator/properties-generator.js#L160

@Caele What do you think?

apis/nucleus/src/viz.js
niekvanstaveren: That would probably help to keep this consistent between sense client and nebula and a good place to have it as source. @Caele do you agree on this?

cbt1: Probably makes more sense to have in sn-table. As currently nebula is dependent on some sn-table props that may or may not change in the future. So this logic in nebula may end up not being compatible with sn-table anymore because sn-table changed its properties.

cbt1: Ex: in version x.x.x of sn-table, we also want to use `someNewProp: "value"` for the view data case. If the user have a version where nebula does not have `someNewProp` set but does have sn-table x.x.x, viewData might not behave as expected

haxxmaxx: So we could have a `viewDataMode` prop or somthing that would currently enable pagination and disable totals?

niekvanstaveren: No I think more like @T-Wizard suggested where we add a function to sn-table (e.g. importViewDataProperties) which provides these properties. Similar to the other export and import properties functionality

haxxmaxx: I see, that would be nice

Caele: yea, should avoid placing them here.
I could see having them just in importProperties, as defaults if there are no table props on the object already. To avoid adding anothing API specific to view data.

apis/conversion/src/index.js
Caele: hmm, don't really care about this one. Either is fine.

apis/stardust/types/index.d.ts
```diff
@@ -411,6 +412,16 @@ declare namespace stardust {
          */
         convertTo(newType: string, forceUpdate?: boolean, forcePatch?: boolean): Promise<object>;
 
+        /**
+         * Toggles the chart to a data view of the chart.
+         * 
+         * The chart will be toggled to the type defined in the nebula context (dataViewType).
+         * 
+         * The default dataViewType for nebula is sn-table. The specified chart type needs to be registered as well in to make it possible to render the data view.
```
haxxmaxx, line 420: I assume you mean this?
```suggestion
         * The default dataViewType for nebula is sn-table. The specified chart type needs to be registered as well, in order to make it possible to render the data view.

```

apis/nucleus/src/__tests__/viz.test.js
```diff
@@ -232,6 +243,27 @@ describe('viz', () => {
     });
   });
 
+  describe('toggleDataView', () => {
```
haxxmaxx, line 246: Potentially one more when you specify the state (true/false) and it actually changes? Or is that overkill?

apis/conversion/src/__tests__/index.test.js
```diff
@@ -163,16 +170,55 @@ describe('objectConversion', () => {
...
         exportFormat: { qProperty: { prop3: 'value 3' } },
         initialProperties: { qInfo: { qType: 'new type' }, visualization: 'new type' },
         dataDefinition: {},
         hypercubePath: '',
+        viewDataMode: false,
       });
       expect(newPropertyTree).toEqual({ qProperty: { prop4: 'value 4' } });
     });
+
+    test('should call export properties with viewdatamode', async () => {
```
haxxmaxx, line 190: Now this is really important ;)
```suggestion
    test('should call export properties with viewDataMode', async () => {
```

haxxmaxx (commented):
very minor, except the wording in the api spec

veinfors (approved):

apis/nucleus/src/index.js
```diff
@@ -141,6 +141,12 @@ const DEFAULT_CONTEXT = /** @lends Context */ {
   keyboardNavigation: false,
   /** @type {boolean=} */
   disableCellPadding: false,
+  /**
+   * Type used for toggling to the data view (toggleDataView)
+   * This type need to be registered as well
+   * @type {string=}
+   * */
+  dataViewType: 'sn-table',
```
Caele, line 149: One thing we could do to help this out is to check for types when doing the toggling. There are basically two options for registering your table chart, either "sn-table" or "table". The "table" one should be quite common as it allows for embedding the old table as the new.
Could be when running toggle explicitly check for the registred types and then use one of the two if either is registred and this "dataViewType" isn't set? (so default it to an empty string, meaing somewhat "auto")

niekvanstaveren: Could look something like this https://github.com/qlik-oss/nebula.js/pull/1418
I put that logic in types on halo since it had easier access to the type list. Otherwise you would need to check them one by one inside the toggleDataView logic

Caele: yea, thats pretty nice. Could actually keep the default value of dataViewType and just use the others as a fallback instead.
I'd just argue for not adding a getDataViewType function to types and instead just either use the getSupportedVersion instead and keep the check in toggleDataView or add a function to get the object map as opposed to an array from getList.

niekvanstaveren: Updated it now. Keeping `sn-table` as default and checking if the dataViewType is supported. If not there is the fallback of using `table` as the type.  The nice thing with this is that we only need one fallback. If you change the default dataViewType I think it can also be expected that you have registered that type

Caele (approved):

PR #1399, opened 2023-11-07
refactor: do not contrast colors for checkboxes

johanlahti:
## Changes:

- do not contrast colors for checkboxes since they do not use a background color for selection states
- use different colors for search text and icons than for fields
- add rendering test for component styled checkboxes

## Motivation

The most important fix is that when contrasting the color in checkbox mode we can end up with white towards a white background.

### New rendering test (verifying this fix):

The rendering test snapshot below shows that selected text color is **not** inverted (it is green, after this fix). Without checkboxes, the same text color should be inverted.
image

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


johanlahti:
PTAL ⬆️

veinfors (approved):

PR #1407, opened 2023-11-14
feat(listbox): expose and inject background image

johanlahti:
## Motivation

This PR enables properties for Listbox background image in addition to color.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


DanielS-Qlik (approved):

apis/nucleus/src/utils/style/styling-props.js
```diff
@@ -51,8 +51,9 @@ function getBackgroundPosition(bgComp) {
 
 function getBackgroundSize(bgComp) {
   let bkgImageSize = imageSizingToCssProperty.originalSize;
-  if (bgComp?.bgImage?.sizing) {
-    bkgImageSize = imageSizingToCssProperty[bgComp.bgImage.sizing];
+  const size = bgComp?.bgImage?.sizing || bgComp?.bgImage?.size;
```
veinfors, line 54: It's a little annoying that the api differs here and that we have to support two cases like this.
Don't know if it's "too late" to change or we have good reason to add `size` as well.

apis/nucleus/src/utils/style/styling-props.js
```diff
@@ -67,9 +68,9 @@ export function resolveBgImage(bgComp, app) {
   if (bgImageDef) {
     let url = '';
     if (bgImageDef.mode === 'media' || bgComp.useImage === 'media') {
-      url = bgImageDef?.mediaUrl?.qStaticContentUrl?.qUrl
-        ? decodeURIComponent(bgImageDef.mediaUrl.qStaticContentUrl.qUrl)
-        : undefined;
+      const urlObj = bgImageDef?.mediaUrl || bgImageDef?.url;
```
veinfors, line 71: Same as the comment above but for `url` here. Any specific reason to why we add yet another prop for the same purpose.

apis/nucleus/src/utils/style/styling-props.js
johanlahti: Yeah, I acquired this from another place (Action button) which used these keys.

johanlahti: It's not too late to change. I can change to what Nebula used before.

PR #1408, opened 2023-11-16
fix: use session based model and request storage

Caele:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/nucleus/src/hooks/useAppSelections.js
```diff
@@ -80,18 +81,23 @@ function createAppSelections({ app }) {
...
     // assume the app is mocked if session is undefined
     return [];
   }
-  const [appSelectionsStore] = useAppSelectionsStore();
+  let store = selectionStore;
+  if (!store) {
+    store = useContext(InstanceContext).selectionStore;
+  }
+
+  const [appSelectionsStore] = store.useAppSelectionsStore();
```
T-Wizard, line 94: you should not have conditional hooks
(could still work if `!store` and `store.useAppSelectionsStore` never change change)

Caele: I can likely work around it yea, think it was the first hook I changed. Then I kept it as it simplified the tests.

apis/nucleus/src/components/listbox/components/ListBoxRowColumn/__tests__/list-box-row-column.test.jsx
```diff
@@ -48,7 +48,7 @@ describe('<ListBoxRowColumn />', () => {
         ],
       },
     ];
-    global.document = {};
+    // global.document = {}; // what?
```
johanlahti, line 51: Remainder from a removed unit test. Remove:
```suggestion
```

PR #1410, opened 2023-11-21
docs: remove internal fn from api spec

veinfors:
Removes the utility function `getFullBoxShadow` from the api spec (was accidentally included in the last nebula version).


apis/nucleus/src/utils/style/shadow-utils.js
```diff
@@ -19,6 +19,7 @@ export const getShadows = (shadowString) => {
 
 /**
  * Combine box shadow with box shadow color.
+ * @ignore
```
T-Wizard, line 22: I would go with
```suggestion
 * @private
```
from what is know that will also exclude it from the spec

T-Wizard (approved):

PR #1426, opened 2023-12-05
fix: bgColor with useColorExpression property

cbt1:
## Motivation

In Sense you can set a background per chart using the Styling panel and General tab, there you can set the background to "color by expression". When setting a color by expression via the Styling panel, `useColorExpression` is set as a property in indicate that the color expression should be used.

Not really sure what is the single source of truth here but I'm assuming charts have been using that panel to set the background, which means there are charts in the wild using `useColorExpression`.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/utils/style/__tests__/styling-props.test.js
+++ b/apis/nucleus/src/utils/style/__tests__/styling-props.test.js
@@ -80,6 +80,16 @@ describe('Styling property resolver', () => {
       expect(color).toBe('resolvedColor');
     });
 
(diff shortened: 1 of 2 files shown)


johanlahti:
I was cleaning up some of these double definitions. Yeah, It's very hard to know what has been used else-where. It's unfortunate that we unnecessarily have two different definitions for the same thing but that's hard to prevent I guess…
@veinfors @Caele

johanlahti (approved):

PR #1465, opened 2024-01-24
fix: always store session model (also when not fetching it)

johanlahti:
### Problem:

It was noticed that layout changes did not trigger the useEffect in `ListBoxInline.jsx`. After some investigation help from @T-Wizard it was found that this was caused by the session model not being stored in Nebula's model store. This in turn lead to the fetch layout call never being performed and the useEffect callback not being called.

### Expected:

The useEffect callback should always be called when layout changes (in this case, tested with `qLocked` state).

### Solution:

With this PR, the session model is *always* stored. The code was also refactored to reduce duplication of code logic.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/hooks/__tests__/useExistingModel.test.jsx
+++ b/apis/nucleus/src/components/listbox/hooks/__tests__/useExistingModel.test.jsx
@@ -21,7 +21,7 @@ describe('useExistingModel', () => {
   let ref;
   let once;
 
-  beforeEach(() => {
+  beforeEach(async () => {
     jest.useFakeTimers();
 
     setMock = jest.fn();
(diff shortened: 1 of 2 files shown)


apis/nucleus/src/components/listbox/hooks/__tests__/useExistingModel.test.jsx
```diff
@@ -49,20 +49,28 @@ describe('useExistingModel', () => {
...
   afterEach(() => {
     jest.useRealTimers();
-    jest.restoreAllMocks();
     jest.resetAllMocks();
     renderer.unmount();
   });
 
+  afterAll(() => {
+    jest.restoreAllMocks();
+  });
```
T-Wizard, line 58: it feels wrong to have this in afterAll as the setup of mocks are in beforeEach (not beforeAll)

PR #1466, opened 2024-01-25
feat(ListBox): break out header and add unlock button

johanlahti:
## Motivation

1. Due to increased complexity, the Listbox header has been broken out into its own `ListBoxHeader.jsx` file and the old lock button has been placed inside a new `ActionsGroup` in the `ActionsToolbar` component so that keyboard navigation and accessibility will work automagically and resemble the look of the existing Listbox. The confusing behaviour of lock and search icons showing conditionally of each other is now removed (previously, lock did not show if the search icon existed but did show if it was locked).

2. An unlock button has been added, visible when setting the option `showLock` to `true` (new option added to spec)

image

This means, the lock button (and the unlock cover button) will not be shown unless `showLock` is `true`.

3. Minor bug fixes found during test race:
- Reset zoom on iPhone (covering additional cases not seen before)
- Error thrown on keyboard nav focus row
- Right-align numerical search results
- Invert keyboard navigation in actions toolbar for RTL
- Ensure action buttons are refreshed (enabled) after unlocking

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


johanlahti:
#devbuild

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -207,14 +169,15 @@ function ListBoxInline({ options, layout }) {
       }
     };
     if (isPopover) {
+      // When isPopover toolbar == false will be ignored.
```
DanielS-Qlik, line 172:
```suggestion
      // When isPopover, toolbar == false will be ignored.
```

DanielS-Qlik (approved):

ashrles (approved):

PR #1476, opened 2024-02-06
feat: add option codeSplit to allow for code-splitting

eliseeborn:
## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/commands/build/lib/build.js
+++ b/commands/build/lib/build.js
@@ -14,6 +14,7 @@ const umd = async (argv) => {
     mode: argv.mode || 'production',
     format: 'umd',
     argv,
+    inlineDynamicImports: true,
   });
   const bundle = await rollup.rollup(c.input);
   await bundle.write(c.output);
--- a/commands/build/lib/build.js
+++ b/commands/build/lib/build.js
@@ -14,7 +14,6 @@ const umd = async (argv) => {
     mode: argv.mode || 'production',
     format: 'umd',
     argv,
-    inlineDynamicImports: true,
   });
   const bundle = await rollup.rollup(c.input);
   await bundle.write(c.output);
--- a/commands/build/lib/config.js
+++ b/commands/build/lib/config.js
@@ -61,7 +61,8 @@ const config = ({
   mode = 'production',
   format = 'umd',
   cwd = process.cwd(),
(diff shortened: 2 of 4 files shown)


commands/build/lib/config.js
```diff
@@ -113,6 +114,32 @@ const config = ({
...
+    }
+    return {
+      banner,
+      format,
+      file: path.resolve(dir, outputFile),
+      name: outputName,
+      sourcemap,
+      globals: {
+        '@nebula.js/stardust': 'stardust',
+      },
```
eliseeborn, line 139: if the option codeSplit is not sent in, output config remains the same

commands/build/lib/init-config.js
```diff
@@ -54,9 +54,8 @@ const options = {
     type: 'boolean',
     default: false,
   },
-  inlineDynamicImports: {
-    description:
-      'In Rollup, inlineDynamicImports option is a configuration option that allows code-splitting while still creating a single bundle.',
+  codeSplit: {
+    description: 'A description.',
```
veinfors, line 58: Add a proper description here

commands/build/lib/build.js
```diff
@@ -14,6 +14,7 @@ const umd = async (argv) => {
     mode: argv.mode || 'production',
     format: 'umd',
     argv,
+    inlineDynamicImports: true,
```
veinfors, line 17: This seem to alter the default build behaviour in nebula build, do we want to do that/is that necessary?

commands/build/lib/config.js
veinfors: Could make this a little cleaner (less code duplication) by having one output object with the default output props, and then, if `argv.codeSplit` you set `inlineDynamicImports` and `dir` accordingly before returning the output object from the function.

commands/build/lib/build.js
eliseeborn: it doesn't, since it's only actually put in the config if codeSplit is sent in as an option, but I'll change it since it's unclear

veinfors: Okey, I see 👍

Caele (approved):

Issue #1448, opened 2024-01-04
Nebula Hub local server - connection to on premise Qlik Sense Server


jeff-1984:
## 🐛 Bug report
I am trying nebula.js, following this page : https://qlik.dev/extend/extend-quickstarts/first-extension/
I have access to a Qlik Sense Enterprise server installed locally on a windows server.

The Nebula Hub page says : 
`
Qlik Sense Enterprise on Windows
WebSocket URL format: wss://<sense-host.com>/<virtual-proxy-prefix>
Example: wss://mycompany.com/bi

Note that for the Qlik Sense Proxy to allow sessions from this webpage, myserver:8000 needs to be whitelisted in QMC in your Qlik Sense Enterprise on Windows deployment. In addition, you need to enable Has secure attribute and set SameSite attribute to None.
Make sure you are logged in to Qlik Sense in another browser tab.
`

- I created a virtual proxy, whitelisted the URL, set the flags, and I am logged on the hub for that virtual proxy (with admin access)
- The Nebula hub displays "Something went wrong, check the devtools console"
- If I inspect the websocket when doing this, I see that the hub is sending "GetConfiguration" but the server answers "OnAuthenticationInformation" and send the authentication redirect URL

It seems the authentication cookie is not sent, compared to what is send on the websocket when I open a Qlik Sense app through the HUB.

Can you help, any advice ?

Thanks

Caele:
I did some testing and initially ran into the secure/sameSite issue but after updating the default virtual proxy it worked as intended.

One issue I noticed was the lack of a proper error, as the intended one should state:
`In your virtual proxy advanced settings in the QMC, make sure to whitelist ${window.location.host}, ensure "Has secure attribute" is enabled and that "SameSite attribute" is set to "None".`,
But it gets overwritten by `Failed to get configuration`. - we'll look into that

Before I updated the virtual proxy I could observe the same pattern as you, the websocked connects but the last message is "OnAuthenticationInformation" follwed by the socked closing. By debugging Nebula I could see that the code actually gets that message here: https://github.com/qlik-oss/nebula.js/blob/master/commands/serve/web/hooks/useConnection.js#L66

To be sure, can you instead of creating a new virtual proxy update the default one?

I got it working with a new virtual proxy as well, by using this:
- Host allow list: add your server here, so example: server.domain.com
  - Did not need to add my localhost url to that list
- Engine URL: wss://server.domain.com/{proxyprefix}

[...]

PR #1482, opened 2024-02-13
fix: allow prefix in serve

Caele:
## Motivation

Fix for https://github.com/qlik-oss/nebula.js/issues/1448

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/commands/serve/web/__tests__/connect.test.js
+++ b/commands/serve/web/__tests__/connect.test.js
@@ -431,6 +431,7 @@ describe('connect.js', () => {
           host: expect.any(String),
           port: undefined, // because of providing a link
           appId: undefined, // since there is not appid in link
+          prefix: undefined,
         }),
         engineUrl: expect.any(String),
         appUrl: undefined, // because no app has been provided
@@ -446,6 +447,7 @@ describe('connect.js', () => {
           host: expect.any(String),
           port: expect.any(String),
           appId: undefined, // since there is not appid in link
+          prefix: undefined,
         }),
         engineUrl: expect.any(String),
         appUrl: undefined, // because no app has been provided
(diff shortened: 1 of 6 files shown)


commands/serve/web/connect.js
```diff
@@ -24,7 +24,7 @@ const getParams = () => {
 // QCS:       wss://<tenant-url>.<region>.qlikcloud.com/app/<app-GUID>
 // QSEoK:     wss://<host>/app/<app-GUID>
 // QSEoW:     wss://<host>/<virtual-proxy-prefix>/app/<app-GUID>
-const parseEngineURL = (url, urlRegex = /(wss?):\/\/([^/:?&]+)(?::(\d+))?/, appRegex = /\/app\/([^?&#:]+)/) => {
+const parseEngineURL = (url, urlRegex = /(wss?):\/\/([^/:?&]+)(?::(\d+))?\/?(.*)/, appRegex = /\/app\/([^?&#:]+)/) => {
```
T-Wizard, line 27: when testing the new urlRegex with example url above.
it looks to me that prefix will also include app. 
Could that not cause problem later in the code?

Caele: hmm, yea true, best to strip app first

Caele: i think the intended flow works, but the whole thing is not exactly what I thought it was. I'll redo this a bit.

commands/serve/web/__tests__/connect.test.js
```diff
@@ -460,7 +462,38 @@ describe('connect.js', () => {
...
+          host: expect.any(String),
+          port: undefined, // because of providing a link
+          appId: 'SOME_APP_ID/', // since there is an appid in link
+          prefix: 'prefix',
+        }),
+        engineUrl: expect.any(String),
+        appUrl: url,
+      });
+
+      url = `wss://${authConfig.host}/prefix/app/SOME_APP_ID/`;
```
T-Wizard, line 488: testing with
url = `wss://${authConfig.host}/app/SOME_APP_ID/`;
give the questionable prefix `app`

Caele: you can likely create a virtual proxy named app, so if you input that it should be fine. Depends on the mode. I don't think the /app part is even used here, as that is a different query param.

Caele: yea I was wrong again, it is indeed used.

a-m-dev (approved):
LGTM

Caele:
Something else gets really odd, need to rethink this

T-Wizard (approved):

PR #1489, opened 2024-02-19
fix(listbox): disable selections when they should

johanlahti:
## Motivation

- Ensure `selectDisabled()` works everywhere, incl. search and keyboard selection (using Space)
- Refactor so that `selectDisabled()` is always used from the `selectionState` object

### After fix: Data manager –> Calculated field preview:
https://github.com/qlik-oss/nebula.js/assets/5780544/ff8d914d-302e-4211-8117-18a4136e4e10

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -72,6 +72,7 @@ export default function ListBoxSearch({
   const { translator } = useContext(InstanceContext);
   const [value, setValue] = useState('');
   const [wildcardOn, setWildcardOn] = useState(false);
+  const selectIsDisabled = !!selectionState.selectDisabled?.();
```
T-Wizard, line 75: recommend moving this into each function using it.
to remove the risk of using an old value

T-Wizard:
I am a bit worried that we affect some other case that expect disable selection to also disable search

johanlahti:
That's good to consider. Yet, I also think we should consider what is the right way, and `disableSelect` should disable *selections* and not search.
If the developer decides to show a search field, then the Listbox will (and should) permit searching. But if `selectDisabled` is `true`, then those searches will not result in selections when pressing `Enter` after the search result has come.

johanlahti:
I wonder if we have any cases where we rely on `selectDisabled` to prevent searching 🤔 (besides someone relying on this in their mashups).
In the data manager e.g. we allow searching although selections are disabled.

T-Wizard:
> I wonder if we have any cases where we rely on `selectDisabled` to prevent searching 🤔 (besides someone relying on this in their mashups). In the data manager e.g. we allow searching although selections are disabled.

maybe check filterpane snapshot?

T-Wizard:
If we disconnect can select & can search.
we should consider also change Locked to only disable select and not search

johanlahti:
I believe this has already been implemented with this PR:
1. When selections are disabled you can still search – in toggle mode you can press cmd+F to open up the search field although the search icon is disabled in edit mode because of `constraints.active === true`
2. Likewise, in locked mode, the locked icon will replace the search icon (for search mode "toggle") – but you can show the search by pressing cmd+F

So, this PR implements that *functionally*. But we may also consider implementing this *visually*, by not disabling (1) or hiding (2) the search icon.

johanlahti:
Here is how it works in live printing (selections disabled):
https://github.com/qlik-oss/nebula.js/assets/5780544/e54d8934-23c9-4126-92b3-d8ec9d6ab360

johanlahti:
@T-Wizard Any more thoughts on this – or concerns that we can and should do something about in this PR?

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -142,7 +144,7 @@ export default function ListBoxSearch({
   const performSearch = async () => {
     let response;
     const success = await model.searchListObjectFor(TREE_PATH, value);
-    if (success && value.length && hasHits()) {
+    if (!selectionState.selectDisabled() && success && value.length && hasHits()) {
```
T-Wizard, line 147: would not a
```
if (selectionState.selectDisabled()) {
  return true/false?; 
}
```
at the beginning of performSearch be more clear.

PR #1492, opened 2024-02-21
fix(cli-serve): OAuth instance cache issue

a-m-dev:
## Motivation
Since we hold only one instance of OAuth using SDK on the (dev) server side of cli serve, the caching mechanism of how we deal/return OAuth instance was failing to interpret the correct client id and host that user intended to connect to. 

This PR try to address that issue and fixes it by checking the (prev) cached required credentials with what ever newer one that user wants to connect. and only overrides it when user would want to connect to a new tenant (basically new host + new client id). 

Also In addition to ^ fix, this PR does a bit of refactor on the router side of OAuth routes, the affected routes are:
- `/oauth`
- `/login/callback`
- `/deauthorize`
- `/isAuthorized`
- `/getSocketUrl/:appId`

which all now have a new (domain related) prefix of `/auth` and makes it more specific about what they want to achieve. Also Since most of the affected urls has been **used internally**, there are not changes to be made on end user side, except one (see below).

### Note: 
By above changes, OAuth (client id) users need to add `/auth/` to their "Redirect URLs" of their token while they are creating a client id. For example if you had a redirect url that been tied to your client id like `http://localhost:8000/login/callback`, now it should be like `http://localhost:8000/auth/login/callback`. 

### Other changes: 
During this PR, there was one other issue related to a scenario when user tried to connect with wrong client id, I noticed that we don't handle that scenario gracefully and took care of it as well in this PR. 

### Pls note:
this is the same as the old one here https://github.com/qlik-oss/nebula.js/pull/1485, That pr was failing in ci when it was trying to push the results of tests as mentioned in comments so the conclusion to it was to create a new pr directly pushing the branch into this repo instead from forked repo.  

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

# Update 01:
Now we are covering the old `/login/callback` page in the web ui with a nice error page guiding through a step by steps things to do about fixing the redirect link, here is a quick look of it:

Screenshot 2024-02-26 at 09 51 29

 ---
Also it's worth mentioning that when a user still uses old redirect url in tenant for his/hers client id (`http(s)://somehost(:port?)/login/callback`) -> when their client id fails -> they will not land in a error page or url that is in control of us, instead they will fall into the a error link from their target tenant, check the url in bellow image: 

Screenshot 2024-02-26 at 09 54 47

we can't do anything about it within the scope of nebula serve!


Caele:
Two things:
- We should keep the old http://localhost:8000/login/callback route and display an error, detailing what to update
- Update the help text to reflect the new route.

Caele:
yea on that error page, I missunderstood it, even if you have the old redirect registred you won't end up on that page. 
For that reason could we rebase this towards the main branch instead? as it would be a more breaking change?
I can handle it if you want?

I also test ran this a bit, at first it didn't authenticate me at all, just displayed the dev error. But when I tried again a while later it all worked.

Caele (approved):

a-m-dev:
I actually went through like 5, 6 steps of `git rebase --continue` but stuck in conflicts about deps like mui stuff, specially conflict states like this, where I could not detect which one should i choose really: 

Screenshot 2024-02-29 at 11 03 28

Probably It's best that you do that favour (I would be so thankful about) since i have no idea about what should i accept regarding some commits in context of "main" branch.

about that dev error you mentioned, I'm guessing maybe it could be the tenant still redirected to the old "redirect url" (regardless of you already added the new one for your client id) but i'm not 100% sure of it. 

btw, since you already approved this, is it ok to merge this one or is there some thing else about main branch that we should consider and wait for?

PR #1494, opened 2024-02-22
fix(listbox): color radio buttons with custom styling

johanlahti:
### Problem:

Radio buttons do not pick up the selected color.

image

### Solution:

Apply `styles.selections.selected` on checked radio buttons.

Also, apply `styles.content.color` for unselected boxes to allow for customisation of unselected radio buttons through the theme (allowing for customisation through the selection state color, such as excluded or possible, might just add confusion and look bad in some cases).

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


T-Wizard (approved):
Do we also want a a new rendering test convering this case ?

johanlahti:
Yeah, I might look into adding one.

PR #1504, opened 2024-03-04
feat(ListBoxHeader): add icon for cyclic dimension and new drilldown icon

linhnguyen-qlik:
## Motivation
-  New icon for drilldown dimension.
- Introduce cyclic dimension which requires a button to step through cyclic dimension

Screenshot 2024-03-05 at 11 03 52

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


apis/nucleus/src/components/listbox/components/ListBoxHeader/DimensionIcon.jsx
```diff
@@ -0,0 +1,29 @@
...
+  const Icon = icon;
+  const title = translator.get(tooltip);
+  const isButton = onClick;
+
+  return isButton ? (
+    <StyledButton variant="outlined" onClick={onClick} tabIndex={-1} title={title} size="large" disableRipple>
+      <Icon style={iconStyle} />
+    </StyledButton>
+  ) : (
+    <Icon title={title} size="large" style={{ ...iconStyle, padding: `${ICON_PADDING}px` }} />
```
linhnguyen-qlik, line 26: Looks like the tooltip for drilldown doesn't work. It doens't work before this PR either.

apis/nucleus/src/components/listbox/components/ListBoxHeader/ListBoxHeader.jsx
```diff
@@ -203,13 +234,7 @@ export default function ListBoxHeader({
...
         <Grid item container alignItems="center" width={iconsWidth} className="header-action-container">
           {lockedIconComp || (showSearchIcon && searchIconComp)}
-          {isDrillDown && (
-            <DrillDownIcon
-              tabIndex={-1}
-              title={translator.get('Listbox.DrillDown')}
-              style={{ ...iconStyle, padding: `${ICON_PADDING}px` }}
-            />
-          )}
+          {dimensionIconComp()}
```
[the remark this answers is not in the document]
linhnguyen-qlik: Thank, I have updated the code as you suggested.

apis/nucleus/src/components/listbox/components/ListBoxHeader/utils.js
```diff
@@ -0,0 +1,36 @@
...
+            })
+            .catch(() => null);
+        },
+      };
+    default:
+      return undefined;
+  }
+};
+
+export default { createDimensionIconData };
```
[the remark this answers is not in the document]
linhnguyen-qlik: I changed it to `icon-utils.js ` since I think it is already in the `ListBoxHeader` folder but let me know if I should go with `header-icons` or move the function inside `DimensionIcon.jsx.`

apis/nucleus/src/components/listbox/components/ListBoxHeader/DimensionIcon.jsx
johanlahti: Ok, good to know. We can fix that in another PR.

apis/nucleus/src/components/listbox/components/ListBoxHeader/icon-utils.js
```diff
@@ -0,0 +1,36 @@
...
+            })
+            .catch(() => null);
+        },
+      };
+    default:
+      return undefined;
+  }
+};
+
+export default { createDimensionIconData };
```
johanlahti, line 36: Is it intentional to export it with default?

Typically, I would otherwise do:
```
export const createDimensionIconData(…) => {
```

and then either:
```
import { createDimensionIconData } from './icon-utils';
```
or
```
import * as iconUtils from './icon-utils';
```

johanlahti (approved):
Looks good!
Check my question regarding `export default` and change unless it's intentional.

johanlahti:
Not crucial, but you could also add simple unit test for `icon-utils`.

apis/nucleus/src/components/listbox/components/ListBoxHeader/icon-utils.js
linhnguyen-qlik: It is intentional since it is only one function I can only du `export default function...` but I don't like that because "cosummer" could write anything like `import anyName from "./icon-utils;` and then use `anyName()`. So it might not show up when you search for the function by its original name. But if you think I should change it then I will change since I usually not working with this repo and don't know the common rules in it.

linhnguyen-qlik:
> Not crucial, but you could also add simple unit test for `icon-utils`.

I have unit-test for it in `icon-utils.test.js`

johanlahti:
Ahh, yeah sorry. I looked at that yesterday already but already forgot about it… 
No worries with the `export default` – just wanted to check.

PR #1509, opened 2024-03-08
fix: make sure stepCycle exists before calling 

linhnguyen-qlik:
## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


apis/nucleus/src/components/listbox/components/ListBoxHeader/icon-utils.js
```diff
@@ -17,16 +18,24 @@ const createDimensionIconData = (dimInfo, app) => {
...
+              app
+                .getDimension(dimInfo.qLibraryId)
+                .then((dimensionModel) => {
+                  if (!dimensionModel.stepCycle) {
+                    // eslint-disable-next-line no-console
+                    console.log(
+                      "engine version or engine api spec version doesn't have support for function stepCycle"
+                    );
+                  }
+                  dimensionModel.stepCycle(1);
```
Caele, line 34: you shouldn't run the stepCycle function if it doesn't exist, just log the message.

linhnguyen-qlik: Yes you are right, I missed the return.

apis/nucleus/src/components/listbox/components/ListBoxHeader/icon-utils.js
```diff
@@ -17,16 +18,25 @@ const createDimensionIconData = (dimInfo, app) => {
...
-        },
+        onClick: app
+          ? () => {
+              app
+                .getDimension(dimInfo.qLibraryId)
+                .then((dimensionModel) => {
+                  if (!dimensionModel.stepCycle) {
+                    // eslint-disable-next-line no-console
+                    console.log(
+                      "engine version or engine api spec version doesn't have support for function stepCycle"
```
T-Wizard, line 31: maybe remove engine version?
from what I know it is only run when engine api spec version doesn't have support for function stepCycle.
if (only) engine version don't have support for stepCycle it will not get here (only the `.catch(() => null)` later)

linhnguyen-qlik: I see then I will remove engine version

linhnguyen-qlik: fixed

Caele (approved):

PR #1511, opened 2024-03-11
fix(Listbox): Limit search length to 64000 characters

johanlahti:
## Motivation

There is a limitation in how many characters we should allow users to search for. This limitation existed in the old Listbox in sense-client but has so far not been added to the new Listbox.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
+++ b/apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
@@ -6,7 +6,7 @@ import InstanceContext from '../../../contexts/InstanceContext';
 import useDataStore from '../hooks/useDataStore';
 import { CELL_PADDING_LEFT } from '../constants';
 
-const MAX_SEARCH_LENGTH = 5000; // limitation in Engine
+const MAX_SEARCH_LENGTH = 64000;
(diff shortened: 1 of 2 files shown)


apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -6,9 +6,12 @@ import InstanceContext from '../../../contexts/InstanceContext';
 import useDataStore from '../hooks/useDataStore';
 import { CELL_PADDING_LEFT } from '../constants';
 
+const MAX_SEARCH_LENGTH = 5000; // limitation in Engine
 const TREE_PATH = '/qListObjectDef';
 const WILDCARD = '**';
 
+const sanitizeValue = (val) => val?.substring(0, MAX_SEARCH_LENGTH); // limit length
```
veinfors, line 13: Maybe call it `limitValue` or `limitSearchValue` instead to indicate better what it's actually doing, since it's not actually "sanitizing" the string, right?

veinfors (approved):

afridell:
Let's wait with merging this PR. Some discussions are ongoing if there is a way to raise the limit of 5000 characters.

PR #1531, opened 2024-04-09
fix: cycle cyclic dimension doesn't work when filter pane is in selec…

linhnguyen-qlik:
…tion mode

## Motivation
Currently, the selection is always activated immediately when  the popover in filter pane is opened. This prevents the list from re-rendering to update with new data when user steps through cyclic dimension. 

The PR changes the above behaviour. For cyclic dimension, the cyclic button will be disabled once user clicks on on the row in the list = selection is activated. 

For popover filterpane, the selection is not activated until user clicks on one of the row in the list. 
For normal filterpane the cyclic button will not be disabled if the filterpane is in selection mode. When user clicks on cyclic button it will call `selection.confirm()` before going to next field
## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


linhnguyen-qlik:
How do I update the snapshots? My laptop generates snapshot but for Mac so it might not be the same as the one coming from Linux.

Caele:
Does the "cycling" always come from within the listbox? And if so, can't it force an update of its own data instead of checking the selectionstate?

Caele:
> How do I update the snapshots? My laptop generates snapshot but for Mac so it might not be the same as the one coming from Linux.

I usually go to the build itself and save those generated images.

linhnguyen-qlik:
I have no idea how to fix `ApiGovPolicy/Listbox JAVASCRIPT` , `ApiGovPolicy/Nebula Js JAVASCRIPT` and `ci/circleci: api-governance`. I don't think my changes cause API change.

linhnguyen-qlik:
> Does the "cycling" always come from within the listbox? And if so, can't it force an update of its own data instead of checking the selectionstate?

Just for the record we agreed on how the listbox should behave when having a cyclic dimension as implemented in this PR.

Caele (approved):
Please do a sanity check of the mashup api as well, just so that is covered. (so not just in client usage)

PR #1542, opened 2024-05-02
fix(QB-25634): press Enter key does not close listbox

Donya-qlik:
**Issue:**
Pressing Enter key after making some selections via straight table search box, does not close the search listbox.

**Solution:**
Being sure that `EnterKey` confirms the selection and "return" the event to avoid stopPropagation()

In this video I clicked Enter to confirm and close the listbox.

https://github.com/qlik-oss/nebula.js/assets/20701546/05d08d82-fe52-4ad0-ab6e-9af5bfd35c14

### DIFF
--- a/apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
+++ b/apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
@@ -114,14 +114,10 @@ export default function getRowsKeyboardNavigation({
         break;
       case KEYS.ENTER:
         confirm();
-        break;
+        return;
       case KEYS.ESCAPE:
-        if (isModal()) {
-          cancel();
-        } else {
-          return; // propagate to other Esc handler
-        }
-        break;
+        cancel();
+        return;
       case KEYS.HOME:
         focusListItems.setFirst(true);
         if (ctrlKey) {
--- a/apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
+++ b/apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
@@ -116,8 +116,12 @@ export default function getRowsKeyboardNavigation({
         confirm();
         return;
       case KEYS.ESCAPE:
+        if (isModal()) {
         cancel();
-        return;
+        } else {
+          return; // propagate to other Esc handler
+        }
+        break;
       case KEYS.HOME:
         focusListItems.setFirst(true);
         if (ctrlKey) {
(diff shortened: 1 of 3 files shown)


apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
```diff
@@ -114,14 +114,10 @@ export default function getRowsKeyboardNavigation({
         break;
       case KEYS.ENTER:
         confirm();
-        break;
+        return;
```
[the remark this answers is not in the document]
linhnguyen-qlik: Ah, NVM I see that you want to propagate the event

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
```diff
@@ -114,14 +114,10 @@ export default function getRowsKeyboardNavigation({
...
       case KEYS.ESCAPE:
-        if (isModal()) {
-          cancel();
-        } else {
-          return; // propagate to other Esc handler
-        }
-        break;
+        cancel();
+        return;
```
T-Wizard, line 120: I think this will not work for keyboard navigation of inline listboxes in filterpane.
then you probably not want to propagate escape when selection is aborted

Donya-qlik: This change is not actually part of the bug-report but I found that press Esc when you have some selections crashes the listbox, it is an obvious error. I think Esc should close the listbox under any circumstances(with/without selections), it is how it works on the old table. I just tried to keep the compatibility, unless it is not how it is designed to work with the nebula listbox?

Donya-qlik: ok, then it makes sense to not to change Esc like this. 
I'll remove Esc related changes since it was not part of the bug report and will take care of it in another PR.

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
```diff
@@ -114,7 +114,7 @@ export default function getRowsKeyboardNavigation({
         break;
       case KEYS.ENTER:
         confirm();
-        break;
+        return;
```
T-Wizard, line 117: would
```
if (isModal()) {
  confirm();
} else {
  return; // propagate to other handler
}
```
work?
I think it would be safer and more correct for the case when the listbox is not a popover.

I think the change it OK for the filterpane case anyway as enter will probably not do anything it should not.
but I would be good to test to verify it.

Donya-qlik: @T-Wizard Yes it works with a bit of modification. 
isModel() is almost always undefined in ListBoxPopover so the check statement throw a type error (isModel is not a function), we get the same error when using Esc after making selections(That I had mentioned before). EnterKey should confirm the selections in all listboxes so I should only ensure that if the listbox is not popover, it will stop propagation and keeps the list open.

T-Wizard (approved):

PR #1544, opened 2024-05-11
fix: Esc key down doesn't close searchbox

Donya-qlik:
**Issue:**
In both popover and filterpane, Escape key does not hide/close search box.

**Solution:**
There is no isModal() function in the case of popoverlistbox, a condition should be set to ensure that the listbox closes when isModal() is undefined. Moreover, in filterpane, pressing escape key should hide the search box and cancel the selections, also put the focus back on the first row

I only used Escape key to abort selections and the searchBox in both filterpane and popoverListbox.

https://github.com/qlik-oss/nebula.js/assets/20701546/808443ff-243e-4454-9d1d-bcdcc234dd13

https://github.com/qlik-oss/nebula.js/assets/20701546/87cb6caf-36b5-41db-b6f6-f483a9fba280


apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
```diff
@@ -119,9 +119,8 @@ export default function getRowsKeyboardNavigation({
         }
         break;
       case KEYS.ESCAPE:
-        if (isModal()) {
-          cancel();
-        } else {
+        cancel();
+        if (typeof isModal === 'undefined') {
```
T-Wizard, line 123: I would expect this to break escape in filterpane where it expect it to propagate only if `isModal() returns false`
have you tested the change there?

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
```diff
@@ -119,9 +119,8 @@ export default function getRowsKeyboardNavigation({
         }
         break;
       case KEYS.ESCAPE:
-        if (isModal()) {
-          cancel();
-        } else {
+        cancel();
```
a-m-dev, line 122: The old logic probably is not canceling selection in some scenarios (the else case), only in **modal** mode. So I'm guessing there exists some (special) scenarios/environments that selections cancelation might not happen on escape key press.

With new logic you are trying to **always** cancel selections first when escape key has been pressed -> which is maybe not the desired behaviour? 

but to assure if it's ok, probably Tobias or any other of our colleges who had worked on keyboard navigation logic may have more stronger/better opinion than me on this!

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -175,6 +175,9 @@ export default function ListBoxSearch({
       case 'Escape': {
         focusRow(container);
         cancel();
+        if (selections.isModal()) {
+          return undefined;
+        }
```
T-Wizard, line 180: have you tested it in filterpane? (I think the change could cause proablem there)

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
Donya-qlik: Yes I had tested it with filterpane but I think I found the case you concerned and I hadn't tried. When there is a text in the filterpane searchbox isModal() returns false so I turned that condition back.

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
Donya-qlik: Yes I tested it with filterpane, selection.isModal() here returns true only when the listbox is a popover list.

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -175,6 +179,11 @@ export default function ListBoxSearch({
       case 'Escape': {
         focusRow(container);
         cancel();
+        abortListObjectSearch();
```
Donya-qlik, line 182: The reason why I'm aborting list search here is because of a recent commit [here](https://github.com/qlik-oss/nebula.js/pull/1531/files#diff-79b554891611ff72399687b538975e8371f2825a6c0968181db2ce309fb58ca3) where a ListboxSearch condition for filterpane has been removed. As far as I know, the condition cannot not be there anymore so instead, Esc key can provide the condition to empty the searchBox, hide it and let the focus back to the main listbox, otherwise the searchbox stays open and pressing Esc or any other key will not help it to go back to its initial state.

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
T-Wizard: that sounds odd. I would have expected it to also be true when selecting in a listbox in a filterpane even when it is not a popover. (I have not tested it myself)

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
T-Wizard: hmm. maybe consider if the linked change should even apply to listbox popovers used by table.
I don't see any option to cycle a cyclic dimension in it anyway.

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
Donya-qlik: @T-Wizard I haven't seen any issue yet. I uploaded a video in description part, maybe it helps to see how it works between filterpane and popoverlistbox?

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
Donya-qlik: @T-Wizard sorry couldn't follow you, what do you mean by "if the link change ..." which link?
I'm not familiar with cyclic dimension conditions in this case. I just asked @linhnguyen-qlik and she explained briefly that they don't want the user go to selection mode until clicks an item in the list so the second option came up that is what you see in this code block.

T-Wizard: > if the link change ..." which link?

the change you linked

> [here](https://github.com/qlik-oss/nebula.js/pull/1531/files#diff-79b554891611ff72399687b538975e8371f2825a6c0968181db2ce309fb58ca3)

also when i tested on https://qcs.us.qlik-stage.com/ I don't see any cycle button in the listbox popover for tables.

linhnguyen-qlik: The cyclic dimension only shows up in filter-pane at the moment. For the table we haven't implemented for cyclic so I don't think we need to change how it works for the list box in table. For me as long as the listbox popover works as described above for popover in filter-pane I am happy.

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
T-Wizard: video don't play for me.

when testing the branch I can get selections.isModal() as true for inline listbox in filterpane
* click value select (not confirm)
* open search
* press escape
-> focus it lost (do work without this change)

I also get "Uncaught TypeError: hide is not a function" from filtarpane popover if closed by escape directly

T-Wizard:
I think it would be a better solution to make the selection deactivated event trigger close popover and make sure that is triggered.

Donya-qlik:
> I think it would be a better solution to make the selection deactivated event trigger close popover and make sure that is triggered.

@T-Wizard I ended up to pass a boolean property from `ListboxPopover` to `ListBoxSearch`, which will be  true when the popover is "open" so the event can be propagated and close the popover listbox only when the Escape is pressed on the table search box and not on a filterpane.

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -75,27 +76,30 @@ export default function ListBoxSearch({
...
   };
 
+  const abortSearch = () => {
+    // When select is disabled we always want to allow abort (but not permitting selections).
+    const preventAbort = selectionState.selectDisabled() ? false : !selections.isModal();
+    if (preventAbort) {
+      return;
+    }
+    abortListObjectSearch();
+  };
```
T-Wizard, line 101: I don't think this should be split into two functions

If we want to totally disconnect seach and selection then preventAbort should be removed in all cases
but if that was not the intention from the change in https://github.com/qlik-oss/nebula.js/pull/1531
then I think we should make start typing in the search input also enter selection mode and then I think the preventAbort should be OK to keep for all cases

Donya-qlik: Understand. I'm also a bit confused that what I should expect from filterpane in this case! I think not many really paid attention how the new changes on table searchbox might affect the filterpane, neither did I. It's good you brought it up then we can fix it.
I'll ask and choose one of your suggestions base on that! I'll let you know 🙏

linhnguyen-qlik: I know too little on how Listbox works but I have some similar/related issues so I tested a little and it seems like this [one](https://github.com/qlik-oss/nebula.js/blob/7560af739b36f3c28ebf6afffd7adc198414b783/apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx#L84) trigger the change of `visible` and the search is aborted.

What i did is to change the code to this  `const cancel = () => selections.cancel();` not sure if cancelling even though  the selection haven't began will cause side effect.

Donya-qlik: @T-Wizard I asked UX (Patrik L.) about this in filterpane. He was agree with the test results I showed in the video. We want the searchbox to be aborted regardless there is any selection or not. Pressing escape key should cancel the unconfirmed selections and hide searchbox in filterpane and close the popover listbox in table.
@linhnguyen-qlik I probably couldn't recognize what you referred to...what is wrong with aborting search in your case? don't you expect it to be aborted ?

linhnguyen-qlik: > @linhnguyen-qlik I probably couldn't recognize what you referred to...what is wrong with aborting search in your case? don't you expect it to be aborted ?

No there is nothing wrong with abortSearch it is just that with the code I mentioned above you could have a fix with just one line of code.

Donya-qlik: Maybe we are talking about different fixes ? 
The purpose of this PR is to be sure that Esc key cancels the selections if there is any, returns the focus back to the initial listbox and be sure the searchbox is gone. It's slightly different with table searchbox which I cancel the selections and close the popover listbox immediately because there we don't need it to stay open, this is the reason why I propagate the event at some points to close whole the listbox. Do these steps break anything from cyclic perspective ?

Donya-qlik: I tried Linh's solution and it sounds working with the known scenarios. From now and then we cancel all selections regardless they are active or not and the searchbox will be closed when Esc key is pressed. I think we don't need preventAbort anymore since we don't want to stop closing searchbox anyway. 

@T-Wizard & @linhnguyen-qlik I don't know about the side effects but we know that the changes have passed at least the existing tests. If you are agree, maybe you want to approve this PR? :)

linhnguyen-qlik: Maybe you could revert/remove the code that are no longer relevant?

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -75,20 +76,14 @@ export default function ListBoxSearch({
...
 
-  const cancel = () => selections.isActive() && selections.cancel();
+  const cancel = () => selections.cancel();
 
   const abortSearch = async () => {
-    // When select is disabled we always want to allow abort (but not permitting selections).
-    const preventAbort = selectionState.selectDisabled() ? false : !selections.isModal();
-    if (preventAbort) {
-      return;
-    }
```
linhnguyen-qlik, line 91: For example revert this one?

T-Wizard (approved):

T-Wizard:
hmm. did some more testing.
some cases of closing a (filterpane) popover don't trigger the abortSearch (still better then without this PR)
(the alternative, trigger selection mode on starting to search is maybe? better for that?)

linhnguyen-qlik:
> hmm. did some more testing. some cases of closing a (filterpane) popover don't trigger the abortSearch (still better then without this PR) (the alternative, trigger selection mode on starting to search is maybe? better for that?)

Since the search is broken I am fixing it and with the fix when user start searching it will go in selection mode. For case when user not search at all we still need to call cancel(). I can probably do that in my PR for fix search if you think it is more appropriate.

luckehal:
Is this behavior fixed? Or is there a tenant where I can test if behavior is still there?

https://qlikdev.slack.com/archives/CV2QW4KPT/p1718183449432139

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -175,6 +170,9 @@ export default function ListBoxSearch({
       case 'Escape': {
         focusRow(container);
         cancel();
+        if (popoverOpen) {
+          return undefined;
+        }
```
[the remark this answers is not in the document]
Donya-qlik: I should be here otherwise pressing Esc when the seachbox is open and there is no selection will not close the popover.

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
Donya-qlik: Why do you think we need preventAbort ?

linhnguyen-qlik: Because when user haven't initiate the search then no need to call `abortListObjectSearch`.

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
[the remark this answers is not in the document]
linhnguyen-qlik: With the change in cancel funktion the popover will be closed eventhough when there are no selection. Have you try to remove this and test again?

linhnguyen-qlik:
> ```js
> undefined
> ```

Why do we need to return `undefined`?

Donya-qlik:
> Is this behavior fixed? Or is there a tenant where I can test if behavior is still there?
> 
>  chrome_xCsb3dNGpb.mp4 
> https://qlikdev.slack.com/archives/CV2QW4KPT/p1718183449432139

@luckehal We hope this PR can fix the known issues. You can also check it by watching the filterpane video I uploaded on top and see if you recognize the scenario you have issue with there. I assume you have issue with Esc key as we do here.

luckehal:
> > Is this behavior fixed? Or is there a tenant where I can test if behavior is still there?
> > chrome_xCsb3dNGpb.mp4
> > https://qlikdev.slack.com/archives/CV2QW4KPT/p1718183449432139
> 
> @luckehal We hope this PR can fix the known issues. You can also check it by watching the filterpane video I uploaded on top and see if you recognize the scenario you have issue with there. I assume you have issue with Esc key as we do here.

@pnr referenced this fix as potentially fixing the behaviour I saw, but I do not touch the ESC key...I shall likely create a new issue then?

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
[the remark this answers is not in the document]
Donya-qlik: By saying popover I mean ListboxPopover in table. If one uses escape when search listbox is open on a table, it is expected to close the whole listbox but it is not what we want in filterpane.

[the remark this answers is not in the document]
linhnguyen-qlik: You mean in filterpane popover escape should not close the popover?

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
T-Wizard: with the change in #1557 i think we can keep preventAbort (then it will prevent unnecessary engine call in some cases)
if we decide to not do that then the preventAbort should be removed

linhnguyen-qlik: As discussed offline we decided to remove preventAbort. No action needed.

Donya-qlik:
> hmm. did some more testing. some cases of closing a (filterpane) popover don't trigger the abortSearch (still better then without this PR) (the alternative, trigger selection mode on starting to search is maybe? better for that?)
> 
> Edit: can be fixed by changing
> 
> ```js
> useEffect(() => {
>     if (!visible) {
>       abortSearch(); // abort when toggling off search
>     }
> }, [visible]);
> ```
> 
> (in ListBoxSearch.jsx) to
> 
> ```js
> useEffect(() => {
>     if (visible) {
>       return () =>abortSearch(); // abort when toggling off search
>     }
>     return undefined;
> }, [visible]);
> ```

It only affects the cases that the search button is clicked, if the searchbox is not visible anymore  it ensures that no value is stored from the previous attempt so the searchbox is empty.
Could you explain briefly how you reproduce the problematic case? is it related to Esc key handling?

T-Wizard:
> > hmm. did some more testing. some cases of closing a (filterpane) popover don't trigger the abortSearch (still better then without this PR) (the alternative, trigger selection mode on starting to search is maybe? better for that?)
> > Edit: can be fixed by changing
> > ```js
> > useEffect(() => {
> >     if (!visible) {
> >       abortSearch(); // abort when toggling off search
> >     }
> > }, [visible]);
> > ```
> >   
> > (in ListBoxSearch.jsx) to
> > ```js
> > useEffect(() => {
> >     if (visible) {
> >       return () =>abortSearch(); // abort when toggling off search
> >     }
> >     return undefined;
> > }, [visible]);
> > ```
> 
> It only affects the cases that the search button is clicked, if the searchbox is not visible anymore it ensures that no value is stored from the previous attempt so the searchbox is empty. Could you explain briefly how you reproduce the problematic case? is it related to Esc key handling?

(to me) Without this change the search is not aborted when filterpane popover closed by clicking somewhere else.
(close by Esc key did work OK)

linhnguyen-qlik (approved):
Tested locally and it seems to work as expected.

linhnguyen-qlik:
@T-Wizard I put your code about visible and click away in to the PR for fixing broken search yesterday. So it is in.

T-Wizard (approved):

PR #1552, opened 2024-05-30
feat: use selection colors from theme for listbox

haxxmaxx:
## Motivation

Adding support for selection colors from theme, for the listbox. This means that chart overrides have the highest prio, then the theme, then the mui theme, the the hardcoded default values. new functionality is behind flag.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


haxxmaxx:
#devbuild

apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -101,7 +102,8 @@ function ListBoxInline({ options, layout }) {
   const { translator, keyboardNavigation, themeApi, constraints } = useContext(InstanceContext);
 
   const { checkboxes = checkboxesOption } = layout || {};
-  const styles = getStyles({ app, themeApi, theme, components, checkboxes });
+  const themeSelectionColorsEnabled = flags.isEnabled('PS_22149_THEME_SELECTION_COLORS');
```
haxxmaxx, line 105: potentially we can get the flag value earlier, and just pass that. Potentially put flags on some kine of context?

T-Wizard: looking at how we done it before.
then we have had the use of flags in https://github.com/qlik-oss/sn-list-objects
and a parameter in nebula.js

but I am not sure why we did it that way

T-Wizard (approved):

apis/nucleus/src/components/listbox/ListBoxInline.jsx
haxxmaxx: ok, I might have a look at that and revisit this. Would like to refactor a bit to align the code for selection colors in filterpane and in here

PR #1554, opened 2024-05-31
fix: show more button even when list box is not in selection

linhnguyen-qlik:
## Motivation
When we introduced the cyclic dimension we decided that the popover listbox (ListBoxInline.jsx) should not be in selection mode immediately when user open it. Therefore this [PR](https://github.com/qlik-oss/nebula.js/pull/1531) is created to fix the issue. 
However, users found it inconvenient that they have to click on a row to activate selection mode to see the more button (...). this PR fixes the issue by always showing the more button in popover list box even when it is not in selection mode.

https://github.com/qlik-oss/nebula.js/assets/111750285/3c338b66-738c-4080-bff4-11a9dd668613

## Requirements checklist

- [X] Api specification
  - [x] Ran `yarn spec`
    - [X] No changes **_OR_** API changes has been formally approved


T-Wizard:
this reason could also apply to "clear selection" button.
should that also be show in this case?

T-Wizard:
have you considered it doing a selection this way should make the listbox enter selection mode?
(I think there are probably arguments both for and against)

linhnguyen-qlik:
> this reason could also apply to "clear selection" button. should that also be show in this case?

Sounds reasonable, I will try to show the clear button too.

linhnguyen-qlik:
@T-Wizard I have changed so that more button shows up even when the filter pane is not in selection mode.
The search and keyboard navigation are still broken but I am creating two separate PRs for them.

T-Wizard:
it it intentional that this change make popups from current selelection no longer show up starting in selection mode?
(even if from engine websocket traffic they actually are in selection mode)

linhnguyen-qlik:
> it it intentional that this change make popups from current selelection no longer show up starting in selection mode? (even if from engine websocket traffic they actually are in selection mode)

No, what I did was only active the selection mode (if not yet in selection mode) when user clicks on one of the options from the menu.

T-Wizard (approved):

Caele (approved):

PR #1557, opened 2024-06-12
fix(listbox): broken search

linhnguyen-qlik:
After this [change](https://github.com/qlik-oss/nebula.js/pull/1531)  the search seems to stop working.
The solution is to start selection once user start typing in  the search input.
## Requirements checklist

- [ ] Api specification
  - [x] Ran `yarn spec`


T-Wizard (approved):
I am OK with this way to solve search
but as an alternative I think it can also be done without starting selection once user start typing in the search input. (if we think that would result in a better user experience)

linhnguyen-qlik:
Close since it is solved by removing `preventAbort` check

PR #1566, opened 2024-06-19
fix: add clear search button to listbox search

jjxie:
Add clear search button to listbox search and add keyboard support to the button.
Screenshot 2024-06-20 at 14 36 32

Fix RTL mode, search text is far to the search icon
Screenshot 2024-06-20 at 14 28 49

Fix dense RTL mode, `search in listbox` text is not placed to the right.
Screenshot 2024-06-20 at 14 31 49

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
```diff
@@ -210,6 +231,45 @@ export default function ListBoxSearch({
     return undefined;
   };
 
+  const focusOnInput = () => {
+    inputRef.current?.setAttribute('tabindex', 0);
+    inputRef.current?.focus();
+    inputRef.current?.setAttribute('tabindex', -1);
```
T-Wizard, line 237: why two setAttribute with tabindex?

I think
```suggestion
    inputRef.current?.setAttribute('tabindex', -1);
    inputRef.current?.focus();
```
would do the same thing

T-Wizard:
tabbing backwards (shift + tab) skips the "clear search" (if intentional why?)

apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx
jjxie: I had some issues to focus back on input, so set tabindex to 0. Seems don't need to change tabindex at all, and focus back works. I will remove change tabindex logic

T-Wizard (approved):

linhnguyen-qlik (approved):

PR #1569, opened 2024-06-24
feat: add keyboard support for cyclic button and allow keyboard navigation to more button

linhnguyen-qlik:
## Motivation
Since we show "More" - button even when not in selection mode we now need to have keyboard navigation to this button.  Same for cyclic button that doesn't have keyboard support.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
```diff
@@ -75,12 +75,7 @@ export default function getRowsKeyboardNavigation({
         }
 
         // Without shift key
-        if (inSelection) {
-          keyboard.focusSelection();
-        } else {
-          currentTarget.blur();
-          keyboard.blur();
```
linhnguyen-qlik, line 82: Could we remove this one without causing problem?

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
```diff
@@ -75,12 +75,7 @@ export default function getRowsKeyboardNavigation({
...
-        }
+        keyboard.focusSelection();
```
linhnguyen-qlik, line 78: There is an issue  with this function `keyboard.focusSelection();` which doesn't return anything so we don't know if there is anything that get focused after this function is called. It would be good if this function does return the focused element. Is it ok to do that? There was a API spec for this function saying not return anything.

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
haxxmaxx: Probably not. `keyboard.blur()` tells nebula that the chart (listbox) has been blurred, and thus updates it state. There might be issues when you try to focus the listbox again, since nebula think the listbox still has focus

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
haxxmaxx: I think that would be fine, if you really need it. How come you need to know which elements get focused?

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
linhnguyen-qlik: I somehow find it strange that it suddenly get blur instead of going to the objects in side the filter-pane. I think it makes more sense that it is not blur unless user pressing escape but I don't much about filter-pane that change might not be good.

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-container.js
```diff
@@ -52,20 +53,12 @@ export default function getListboxContainerKeyboardNavigation({
     }
 
     const container = event.currentTarget.closest('.listbox-container');
-    const inSelection = isModal();
```
linhnguyen-qlik, line 55: what modal means in this context?

linhnguyen-qlik:
@haxxmaxx I decided not to fix the keyboard handler for the search icon. I think it is better to do that in another PR to make it easier to review. This one have already fix more things than what the PR title saying.

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-container.js
T-Wizard: it means is in selection mode:
(or maybe that there is another listbox for the same object that is in selection mode)

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-container.js
```diff
@@ -52,20 +53,12 @@ export default function getListboxContainerKeyboardNavigation({
...
-            focusRow(container) || focusSearch(container);
-          } else {
-            focusSearch(container) || focusRow(container);
-          }
-        } else if (shiftKey) {
-          keyboard.blur(true);
+        if (shiftKey) {
+          focusRow(container) || focusSearch(container);
         } else {
-          break;
```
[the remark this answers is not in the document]
linhnguyen-qlik: I just tried to restore logic but not sure if this is the correct conversion.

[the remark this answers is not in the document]
T-Wizard: to keep the old behavior in global selection you probably need to keep the `if (inSelection) {`

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
```diff
@@ -75,11 +75,10 @@ export default function getRowsKeyboardNavigation({
...
 
         // Without shift key
-        if (inSelection) {
-          keyboard.focusSelection();
-        } else {
+        if (!keyboard.focusSelection()) {
           currentTarget.blur();
-          keyboard.blur();
+          keyboard.blur(true);
+          event.stopPropagation();
```
T-Wizard, line 81: is this needed?
(there is already a event.stopPropagation(); at the end of the function, after the switch)

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-rows.js
```diff
@@ -75,11 +75,10 @@ export default function getRowsKeyboardNavigation({
         }
 
         // Without shift key
-        if (inSelection) {
-          keyboard.focusSelection();
-        } else {
+        if (!keyboard.focusSelection()) {
           currentTarget.blur();
-          keyboard.blur();
+          keyboard.blur(true);
```
T-Wizard, line 80: I am not sure I think going back to focus the container is the right change here.
compared to keep the focuse inside the listbox. (going to cyclic button or search if exist)
but I will accept it as OK.

T-Wizard (approved):

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-methods.js
```diff
@@ -7,8 +9,45 @@ export const focusSearch = (container) => {
...
+};
+
+export const focusCyclicButton = (container) => {
+  const button = container?.querySelector('.listbox-cyclic-button:not(:disabled)');
+  button?.setAttribute('tabIndex', 0);
+  button?.focus();
+  return button;
+};
+
+export const blur = (event, keyboard) => {
```
T-Wizard, line 29: with the use of blur outisde kayboard-nav-container
I think it need a

```
if (!keyboard.enabled) {
   return;
}
```

or something similar
(not 100% sure yet)

as it is it break the nebula.js mashup case (when keyboard.enabled === false)

T-Wizard (changes requested):

T-Wizard (approved):

apis/nucleus/src/components/listbox/interactions/keyboard-navigation/keyboard-nav-methods.js
linhnguyen-qlik: Fixed

PR #1570, opened 2024-06-27
fix: activate selection mode on show search input for expanded listbox

linhnguyen-qlik:
## Motivation
Recover the old behaviour of expanded listbox. Opening search input in expanded listbox should also activate selection mode.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -226,7 +226,7 @@ function ListBoxInline({ options, layout }) {
   const handleShowSearch = () => {
     const newValue = !showSearch;
     setShowSearch(newValue);
-    if (!isPopover && !selections.isActive()) {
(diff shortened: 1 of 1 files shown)


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -226,6 +226,9 @@ function ListBoxInline({ options, layout }) {
   const handleShowSearch = () => {
     const newValue = !showSearch;
     setShowSearch(newValue);
+    if (!isPopover && !selections.isActive()) {
```
T-Wizard, line 229: maybe also check for if it show or hide search?
(probably work OK without as I think for close it will already be in selection and the exit selections is done later)
```suggestion
    if (newValue && !isPopover && !selections.isActive()) {
```

T-Wizard (approved):

PR #1576, opened 2024-07-15
fix: always show action toolbar in popover but not necessary being in selection mode

linhnguyen-qlik:
## Motivation
This is an improvement of this [PR](https://github.com/qlik-oss/nebula.js/pull/1531) . Since the change is introduced the customer has been complaining about the new behaviour (That the action toolbar doesn't show up upon opening a popover filter-pane).
This PR revert back the old behaviour while also fixing the issue in cyclic dimension. 
With this PR the action toolbar will show up upon opening the popover filter-pane but the selection mode is not activated unless user actually clicks on at least one row in the list. For expanded filter-pane it works like before.

Downside of showing the action toolbar when not necessary being in selection mode is that it is not visible for user if filter-pan is in selection mode. It might be confusing for developers who associate selection mode is active with the selection toolbar showing up. Solution to this might be disable cancel and confirm button when filter-pane is not in selection mode. In case this is important the change will be implemented.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


Donya-qlik (approved):
It sounds working as expected on both filterpane and popover listbox of a chart :) LGTM

PR #1577, opened 2024-07-15
fix(QB-27479): Tab key navigation on popover listbox

Donya-qlik:
With this fix, it is possible to navigate through the listbox by TAB key when the popover list box is open on the other chart than filterpane. Multiple tabbing moves the tab to the confirmation button and returns it to the searchbox.

https://github.com/user-attachments/assets/3f7b05f2-089f-49d6-b967-e32064d6028d

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxPopover.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxPopover.jsx
@@ -220,7 +220,7 @@ export default function ListBoxPopover({
             onSetListCount={(c) => setListCount(c)}
             onCtrlF={onCtrlF}
             styles={styles}
-            keyboard={{ isEnabled: true }}
+            keyboard={keyboard}
           />
         </Grid>
       </Grid>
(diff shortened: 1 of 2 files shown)


apis/nucleus/src/components/listbox/ListBoxPopover.jsx
```diff
@@ -201,7 +204,7 @@ export default function ListBoxPopover({
               listCount={listCount}
               selections={selections}
               selectionState={selectionState}
-              keyboard={{ enabled: false }}
+              keyboard={{ innerTabStops: true, enable: false }}
```
linhnguyen-qlik, line 207: Do you know why it has to have these pros `{ innerTabStops: true, enable: false }`?

Donya-qlik: Because of this: https://github.com/qlik-oss/nebula.js/blob/22c64bc0a5089a81bbf8b8d6dd06429132455095/apis/nucleus/src/components/listbox/components/ListBoxSearch.jsx#L299

linhnguyen-qlik (approved):
I don't think this is the right solution and as discussed in slack the shift + tab is not working and I think the right way to solve to is to provide the handles like in [here](https://github.com/qlik-oss/nebula.js/blob/22c64bc0a5089a81bbf8b8d6dd06429132455095/apis/nucleus/src/components/listbox/ListBoxInline.jsx#L127) to fully support the keyboard navigation.
However it works for the tabbing now and you said that you will create a new bug for shift + tab so I am ok with the solution for now.

PR #1578, opened 2024-07-17
refactor: listbox, ListBoxSearch and add unit test

linhnguyen-qlik:
## Motivation
- Implementation of Listbox and ListBoxSearch is complicated and easy for creating bug so this PR tries to simplify the implementation while preserves the same behaviours. 
- Add some new unit tests.
- Migrate some unit test using 'react-test-renderer' to using  '@testing-library/react' instead.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


T-Wizard (approved):
I think the changes looks OK (from just reading the code)

Caele (approved):
looks ok, please make sure to test different scenarios such as the popover and inline render in mashups. Hard to see potential issues from just reading the code.

linhnguyen-qlik:
@Caele I have tested with mashup and it works as expected.

PR #1586, opened 2024-08-09
feat: add sheet navigation

quanho:
This PR is to add sheet navigation capability to support using sn-nav-menu in a mashup using qlik-embed to embed a sheet. The main idea is to create a navigation api for each sheet rendering to avoid the case when there are more than one sheet is rendered.
The navigation api needs to know the sheet to update the sheet based on new sheet ID.
The navigation api is passed to each chart context.

https://github.com/user-attachments/assets/e29414c0-3cba-48e5-86f1-6bc9d261c5d9

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/Sheet.jsx
+++ b/apis/nucleus/src/components/Sheet.jsx
@@ -20,7 +20,10 @@ const SheetElement = {
 function getCellRenderer(cell, halo, initialSnOptions, initialSnPlugins, initialError, onMount, navigation) {
   const { x, y, width, height } = cell.bounds;
   return (
-    <div style={{ left: `${x}%`, top: `${y}%`, width: `${width}%`, height: `${height}%`, position: 'absolute' }}>
+    <div
+      style={{ left: `${x}%`, top: `${y}%`, width: `${width}%`, height: `${height}%`, position: 'absolute' }}
+      key={cell.model.id}
+    >
       <Cell
         ref={cell.cellRef}
         halo={halo}
--- a/apis/nucleus/src/components/Sheet.jsx
+++ b/apis/nucleus/src/components/Sheet.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, useContext, useMemo } from 'react';
+import React, { useEffect, useState, useContext, useMemo, forwardRef, useImperativeHandle } from 'react';
 import useLayout from '../hooks/useLayout';
 import getObject from '../object/get-object';
 import Cell from './Cell';
@@ -17,7 +17,7 @@ const SheetElement = {
   className: 'njs-sheet',
 };
 
-function getCellRenderer(cell, halo, initialSnOptions, initialSnPlugins, initialError, onMount) {
+function getCellRenderer(cell, halo, initialSnOptions, initialSnPlugins, initialError, onMount, navigation) {
   const { x, y, width, height } = cell.bounds;
   return (
     <div style={{ left: `${x}%`, top: `${y}%`, width: `${width}%`, height: `${height}%`, position: 'absolute' }}>
@@ -30,6 +30,7 @@ function getCellRenderer(cell, halo, initialSnOptions, initialSnPlugins, initial
         initialSnPlugins={initialSnPlugins}
         initialError={initialError}
         onMount={onMount}
+        navigation={navigation}
       />
     </div>
   );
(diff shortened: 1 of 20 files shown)


apis/nucleus/src/object/initiate-sheet.js
```diff
@@ -1,22 +1,23 @@
...
+import createNavigationApi from './navigation/navigation';
 
-export default async function initSheet(model, optional, halo, initialError, onDestroy = async () => {}) {
+export default async function initSheet(model, optional, halo, store, initialError, onDestroy = async () => {}) {
   const api = sheetAPI({
     model,
     halo,
     initialError,
     onDestroy,
   });
```
Caele, line 11: I think it looks better to pass `navigation` in here, and not into the `mount()` call. We don't want a new navigation API instance for each mount call?

quanho: Good idea! Fixed! Thanks!

apis/nucleus/src/components/sheetGlue.jsx
```diff
@@ -14,10 +23,12 @@ export default function glue({ halo, element, model, initialSnOptions, initialSn
...
       onMount={onMount}
+      navigation={navigation}
     />,
     element,
     model.id
   );
+  navigation.setSheetRef(sheetRef);
 
   const unmount = () => {
     root.remove(portal);
```
Caele, line 34: Below here is a `model.on('closed', unmount)`, won't that potentially mess with things?

Caele:
Make sure you test this with a multi-sheet mashup, so that they don't interfere with eachother. They should be able to display the same sheet for example, and independently switch as well.

apis/nucleus/src/object/initiate-sheet.js
quanho: To support for single nav-menu object case, the navigation creation is moved up to apis/nucleus/src/object/get-generic-object.js and from that pass navigation to either a sheet or a single nav-menu object.

quanho:
> Make sure you test this with a multi-sheet mashup, so that they don't interfere with eachother. They should be able to display the same sheet for example, and independently switch as well.

Tested rendering the same sheet twice in a mashup using qlik-embed
https://github.com/user-attachments/assets/9f35bd50-2c21-4cae-9162-e59d1ba3f375

apis/supernova/src/hooks.js
```diff
@@ -772,6 +772,31 @@ export function useDeviceType() {
...
+ * // Render a sheet
+ * const nebbie = embed(app);
+ * const sheetRender = await nebbie.render({
+ *   element: sheetElement
+ *   id: 'sheetId',
+ * });
+ *
+ * @example
+ * // Use a navigation object to control sheet navigation
+ * const plugins = usePlugins();
```
T-Wizard, line 789: why usePlugin? (remove)

apis/supernova/src/hooks.js
```diff
@@ -772,6 +772,31 @@ export function useDeviceType() {
...
+ * const navigationRender = await nebbie.render({
+ *   element: navigationElement
+ *   id: 'navigationId',
+ *   navigation: sheetRender.navigation
+ * });
```
T-Wizard, line 794: maybe wrong place for this example?
it if for how to use navigation with .render function not how to use useNavigation

apis/supernova/src/hooks.js
```diff
@@ -772,6 +772,31 @@ export function useDeviceType() {
...
+ */
+export function useNavigation() {
+  return useInternalContext('navigation');
+}
```
T-Wizard, line 798: currently useNavigation always return a navigation object even if it is not connected to anything.
In my (current) opinion it would be better to return null/undefined when there is no navigation context making it the object aware if navigation is possible

T-Wizard:
should the new exposed api be merked as "@experimental"?

apis/nucleus/src/object/navigation/navigation.js
```diff
@@ -0,0 +1,54 @@
...
+    setCurrentSheetId: (sheetId) => {
+      State.sheetId = sheetId;
+    },
+    /**
+     * Set the sheet ref
+     * @param {object} sheetRef sheet ref object
+     */
+    setSheetRef: (sheetRef) => {
+      State.sheetRef = sheetRef;
+    },
```
T-Wizard, line 41: I think setCurrentSheetId and setSheetRef are only intended for internal use.
maybe mark as `@private`

apis/nucleus/src/object/navigation/navigation.js
```diff
@@ -0,0 +1,54 @@
...
+      const { modelStore, rpcRequestModelStore } = store;
+      const key = `${sheetId}`;
+      let rpc = rpcRequestModelStore.get(key);
+      if (!rpc) {
+        rpc = halo.app.getObject(sheetId);
+        rpcRequestModelStore.set(key, rpc);
+      }
+      const model = await rpc;
+      modelStore.set(key, model);
+      State.sheetRef?.current?.setModel?.(model);
```
T-Wizard, line 26: maybe add code for if sheetId is not a valid sheet id (not an object, or an object that is not a sheet)
(here or maybe in Sheet.jsx)

apis/supernova/src/hooks.js
quanho: Thanks, stupid mistake from copy/paste. Fixed!

apis/supernova/src/hooks.js
quanho: Indeed it is a wrong place. Updated!

apis/supernova/src/hooks.js
quanho: Thanks! Fixed!

apis/nucleus/src/object/navigation/navigation.js
quanho: Good idea! Fixed!

apis/nucleus/src/object/navigation/navigation.js
quanho: Fixed!

apis/nucleus/src/object/create-session-object.js
```diff
@@ -39,14 +39,15 @@ import init from './initiate';
...
-  { type, version, fields, properties, options, plugins, element, extendProperties },
+  { type, version, fields, properties, options, plugins, element, extendProperties, navigation: inputNavigation },
   halo,
   store
 ) {
   let mergedProps = {};
   const children = [];
   const { modelStore, subscribe } = store;
   let error;
+  const navigation = inputNavigation;
```
veinfors, line 50: No need to bring it in as `inputNavigation` here and assign it to `navigation` before returning. Might as well just take in `navigation` and return the same, if I'm not missing anything...

veinfors: Or, actually, I don't see that we pass in navigation into this fn at all, so probably can remove it from here completely?

apis/nucleus/src/components/Sheet.jsx
```diff
@@ -17,10 +17,13 @@ const SheetElement = {
...
 };
 
-function getCellRenderer(cell, halo, initialSnOptions, initialSnPlugins, initialError, onMount) {
+function getCellRenderer(cell, halo, initialSnOptions, initialSnPlugins, initialError, onMount, navigation) {
   const { x, y, width, height } = cell.bounds;
   return (
-    <div style={{ left: `${x}%`, top: `${y}%`, width: `${width}%`, height: `${height}%`, position: 'absolute' }}>
+    <div
+      style={{ left: `${x}%`, top: `${y}%`, width: `${width}%`, height: `${height}%`, position: 'absolute' }}
+      key={cell.model.id}
```
Caele, line 25: Whats with the key here?

apis/nucleus/src/components/Sheet.jsx
```diff
@@ -47,115 +51,141 @@ function getBounds(pos, columns, rows) {
...
+  ) => {
+    const { root } = halo;
+    const [model, setModel] = useState(inputModel);
+    const [layout] = useLayout(model);
+    const { theme: themeName, modelStore } = useContext(InstanceContext);
+    const [cells, setCells] = useState([]);
+    const [bgColor, setBgColor] = useState(undefined);
+    const [bgImage, setBgImage] = useState(undefined);
+    const [deepHash, setDeepHash] = useState('');
+    navigation?.setCurrentSheetId?.(model.id);
```
Caele, line 67: As we want some kind of event off this we need to make sure it doesn't trigger for each render.

apis/nucleus/src/components/glue.jsx
```diff
@@ -27,6 +28,7 @@ export default function glue({
       initialError={initialError}
       onMount={onMount}
       emitter={emitter}
+      navigation={navigation}
     />,
     element,
     currentId
```
Caele, line 34: You still have the model.on('closed', unmount) below here.

apis/nucleus/src/components/Sheet.jsx
```diff
@@ -47,115 +51,141 @@ function getBounds(pos, columns, rows) {
...
+      const onModelClose = () => {
+        model.removeListener('closed', onModelClose);
+        unmount();
       };
-      fetchObjects();
-    }
-  }, [layout]);
+      model.on('closed', onModelClose);
+      return () => model.removeListener('closed', onModelClose);
+    }, [model]);
```
Caele, line 130: Why did you move this in from sheetGlue?

apis/nucleus/src/components/Sheet.jsx
quanho: This is a list of cells, without a key it will use index and this causes a problem when we switch from a sheet to another sheet. Cells in a new list are not rendered, instead cells in old list are still used.

apis/nucleus/src/index.js
```diff
@@ -84,6 +84,7 @@ const DEFAULT_SNAPSHOT_CONFIG = /** @lends SnapshotConfiguration */ {
  * @property {Array<TypeInfo>=} types Visualization types to register
  * @property {Array<ThemeInfo>=} themes Themes to register
  * @property {object=} anything
+ * @property {object=} flags
```
Caele, line 87: I don't think we want to publicly export flags like this, its an internal thing.

apis/nucleus/src/object/navigation/navigation.js
```diff
@@ -0,0 +1,67 @@
...
+    },
+    /**
+     * @private
+     * Set the current sheet id
+     * @param {string} sheetId Id of the current sheet
+     */
+    setCurrentSheetId: (sheetId) => {
+      State.sheetId = sheetId;
+    },
```
Caele, line 46: We want some kind of trigger/event from this right?

apis/nucleus/src/object/navigation/navigation.js
```diff
@@ -0,0 +1,67 @@
+export default function createNavigationApi(halo, store) {
+  const { galaxy } = halo.public;
+  if (galaxy.anything?.sense?.navigation) {
+    return galaxy.anything?.sense?.navigation;
+  }
```
T-Wizard, line 5: is this still needed?

apis/nucleus/src/index.js
quanho: This is needed to pass apis?.features from qlik-embed to nebula.

Caele: no, you can override that right? Add a ts-ignore thing.

apis/nucleus/src/components/Sheet.jsx
quanho: There is a check if sheetId is changed or not in setCurrentSheetId, so it will do nothing if the sheetId is not changed.

apis/nucleus/src/object/navigation/navigation.js
```diff
@@ -0,0 +1,70 @@
...
+      try {
+        model = await rpc;
+        if (model.genericType !== 'sheet') {
+          return;
+        }
+      } catch (e) {
+        return;
+      }
+      modelStore.set(key, model);
+      State.sheetRef?.current?.setModel?.(model);
```
Caele, line 35: I think that this should also set the current sheet ID.
Is there really any reason for ever having to call setCurrentSheetId "manually"?

apis/nucleus/src/object/navigation/navigation.js
```diff
@@ -0,0 +1,70 @@
...
+     * @private
+     * Set the current sheet id
+     * @param {string} sheetId Id of the current sheet
+     */
+    setCurrentSheetId: (sheetId) => {
+      if (State.sheetId === sheetId) {
+        return;
+      }
+      State.sheetId = sheetId;
+      navigationAPI.emit('currentSheetIdChange');
```
Caele, line 47: think we should use a simpler name, like "sheetChanged", there is not much reason to have an ID change if the sheet itself didn't change.

apis/nucleus/src/sheet.js
```diff
@@ -72,6 +72,11 @@ export default function sheet({ model, halo, initialError, onDestroy = async ()
      * @type {string}
      */
     model,
+    /**
+     * The navigation api to control sheet navigation.
+     * @type {Navigation}
+     */
```
Caele, line 78: add `@since vx.y.z` and `@experimental`, same for the hook.

apis/nucleus/src/components/Sheet.jsx
```diff
@@ -47,115 +51,141 @@ function getBounds(pos, columns, rows) {
...
```
Caele, line 131: One last thing, I would like to move this code out into sheet.js. As that is the "object" that owns this component and should be responsible for unmounting. However that requires a bit of restructuring as the new model is never passed through that sheet API, like changing the navigations setSheetRef to setSheetAPI and add a set model on the sheet API (could be useful for other purposes actually).
So not gonna block you with this, just consider it. I think the current setup works just fine, and is all internal anyways.

Caele (approved):
Think we are good! Had one last comment, but nothing we have to do now.

PR #1591, opened 2024-08-16
refactor: update navigation specs

quanho:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/object/navigation/navigation.js
+++ b/apis/nucleus/src/object/navigation/navigation.js
@@ -10,10 +10,35 @@ export default function createNavigationApi(halo, store, model) {
 
   /**
    * @class Navigation
+   * @description The navigation api instance.
+   * @experimental
+   * @since 5.4.0
+   * @example
+   * const navigation = useNavigation();
+   * //...
+   * useEffect(() => {
+   *   const onSheetChanged = () => {
+   *     setActiveSheetId(navigation?.getCurrentSheetId() || "");
+   *   };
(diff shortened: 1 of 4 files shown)


apis/nucleus/src/object/navigation/navigation.js
```diff
@@ -10,10 +10,37 @@ export default function createNavigationApi(halo, store, model) {
...
    * @class Navigation
+   * @description The navigation api instance.
+   * @experimental
+   * @since 5.4.0
+   * @example
+   * const navigation = useNavigation();
+   * //...
+   * useEffect(() => {
+   *   const onSheetChanged = () => {
+   *     setActiveSheetId(navigation?.getCurrentSheetId() || "");
```
Caele, line 21: Replace this line with a `...` just so we don't reference anything that isn't explained.

apis/nucleus/src/object/navigation/navigation.js
```diff
@@ -10,10 +10,37 @@ export default function createNavigationApi(halo, store, model) {
...
+   *   return () => {
+   *     if (navigation?.removeListener) {
+   *       navigation.removeListener("sheetChanged", onSheetChanged);
+   *     }
+   *   };
+   * }, [navigation]);
+   *
+   * const onSheetClick = (sheetId: string) => {
+   *   navigation?.goToSheet(sheetId);
+   *   setActiveSheetId(sheetId);
```
Caele, line 35: same here

PR #1594, opened 2024-08-22
fix: support alpha value in color returned

quanho:
Just copied from sense client to nebula

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


veinfors:
I think it's overkill to pull in that whole color class implementation just for this use case... we could strip it down a lot I think, only adding minimal functionality to support this case for now, then possibly extend it further if we see the need for it. I also don't think we need to add that `class.js` file either, I think we can accomplish what we need without that as well.

haxxmaxx:
Is it possible to use this? https://github.com/qlik-oss/nebula.js/blob/main/apis/nucleus/src/utils/color/color.js

quanho:
> Is it possible to use this? https://github.com/qlik-oss/nebula.js/blob/main/apis/nucleus/src/utils/color/color.js

Now I move utils to https://github.com/qlik-oss/nebula.js/tree/main/packages/ui so it can be used in both @nebula.js/theme and @nebula.js/nucleus

veinfors (approved):

PR #1611, opened 2024-09-16
refactor: use external color handling

haxxmaxx:
## Motivation

The Color class has been modernized and moved to qlik-chart-modules. This means we can remove Color class and some other color stuff that was implemented twice.

Adding qlik-chart-modules to the deps does increase the bundle size. It is noticeable but too acceptable... 

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`


apis/nucleus/src/components/listbox/assets/__tests__/styling.test.js
```diff
@@ -200,65 +194,64 @@ describe('styling', () => {
     });
   });
 
-  const hasEnoughContrast = (a, b) => getContrast(a, b) > CONTRAST_THRESHOLD;
+  // const hasEnoughContrast = (a, b) => getContrast(a, b) > CONTRAST_THRESHOLD;
```
haxxmaxx, line 197: TODO: potentially move some of these tests to chart-modules

apis/nucleus/src/components/listbox/assets/__tests__/styling.test.js
```diff
@@ -116,14 +110,14 @@ describe('styling', () => {
       expect(styles.search.color).toEqual('object.listBox,content,color');
     });
     it('search - should get desired color if contrasting enough', () => {
-      themeApi.getStyle = () => '#999';
+      themeApi.getStyle = () => '#888888';
```
haxxmaxx, line 113: the contrast for this was actually pretty bad (< 3:1) which isn't acceptable in any case. Probably due to the incorrect contrast calculations, and the arbitrary threshold, this was considered good enough

PR #1613, opened 2024-09-18
feat: disable portal option

T-Wizard:
## Motivation

to get the fix from #1568 for cases where needed
but not when it causes the problem that caused it to be reverted (in #1593)

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/ActionsToolbar.jsx
+++ b/apis/nucleus/src/components/ActionsToolbar.jsx
@@ -72,6 +72,7 @@ function ActionsToolbar({
     alignTo: null,
     popoverProps: {},
     popoverPaperStyle: {},
+    disablePortal: true,
   },
   popover = {
     show: false,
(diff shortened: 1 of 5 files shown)


T-Wizard:
questions:
should the option be in __DO_NOT_USE__ part of the options?
any suggestion for better name?

Caele:
> questions: should the option be in **DO_NOT_USE** part of the options? any suggestion for better name?

When will we use it?
If it is some edge case then sure put it in DO_NOT_USE, but else its better to have in public. As for the name, I think we can keep disablePortal as that is exactly what we are doing. Trying to come up with a different name can just be more confusing.

T-Wizard:
> > questions: should the option be in **DO_NOT_USE** part of the options? any suggestion for better name?
> 
> When will we use it? If it is some edge case then sure put it in DO_NOT_USE, but else its better to have in public. As for the name, I think we can keep disablePortal as that is exactly what we are doing. Trying to come up with a different name can just be more confusing.

We will use it in filterpane (qlik-oss/sn-list-objects#462) - to fix issure that more menu render under other things
but can not be used in table/pivot-table listbox popup as there it will break stuff

Caele (approved):

veinfors (approved):

PR #1628, opened 2024-10-14
feat: add host config to config and galaxy

Caele:
## Motivation

Adds the possibility to pass a hostConfig into Nebula, making it available on the Galaxy env parameter for use in charts.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/commands/build/lib/systemjs.js
+++ b/commands/build/lib/systemjs.js
@@ -1,6 +1,13 @@
 const systemjsBehaviours = {
   getExternal: ({ config: cfg }) => {
-    const defaultExternal = ['@nebula.js/stardust', 'picasso.js', 'picasso-plugin-q', 'react', 'react-dom', 'qmfe-api'];
+    const defaultExternal = [
+      '@nebula.js/stardust',
+      'picasso.js',
+      'picasso-plugin-q',
+      'react',
+      'react-dom',
+      '@qlik-trial/qmfe-api',
+    ];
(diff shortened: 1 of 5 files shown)


commands/build/lib/systemjs.js
```diff
@@ -1,6 +1,6 @@
 const systemjsBehaviours = {
   getExternal: ({ config: cfg }) => {
-    const defaultExternal = ['@nebula.js/stardust', 'picasso.js', 'picasso-plugin-q', 'react', 'react-dom'];
+    const defaultExternal = ['@nebula.js/stardust', 'picasso.js', 'picasso-plugin-q', 'react', 'react-dom', 'qmfe-api'];
```
gabbaxx, line 3: @qlik-trial/qmfe-api ?

StefanEnberg (approved):
LGTM

Anuj0047 (approved):
thanks!

Caele:
> OK. maybe also add useHostConfig hook?

maybe, but it won't actually change across a session, so having it as a hook doesn't really give anything more.

T-Wizard:
> > OK. maybe also add useHostConfig hook?
> 
> maybe, but it won't actually change across a session, so having it as a hook doesn't really give anything more.

yes it would be more for consistency
(flags, deviceType, translator do have a hooks)

(and any hook do not have to be in this PR. can be added later if we decide that we want it)

PR #1633, opened 2024-10-21
fix: make browser the default target

Caele:
## Motivation

Browser should be the default target of the nebula build, but it turned out to be set to false.

Also fixes regex imports for default externals for core builds

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/commands/build/lib/config.js
+++ b/commands/build/lib/config.js
@@ -61,7 +61,7 @@ const config = ({
   mode = 'production',
   format = 'umd',
   cwd = process.cwd(),
-  argv = { sourcemap: true, codeSplit: false, browser: true },
+  argv = { sourcemap: true, codeSplit: false },
   core,
   behaviours: {
     getExternal = getExternalDefault,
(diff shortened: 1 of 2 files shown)


commands/build/lib/config.js
```diff
@@ -53,6 +53,20 @@ const getExternalDefault = ({ pkg }) => {
...
+    'picasso-plugin-q',
+    'react',
+    'react-dom',
+    /^(?!@qlik-trial\/qmfe-data-client-parcels)(@qlik-trial\/qmfe-)/,
+    /^@qlik\/api\//,
+  ];
+  const peers = Object.keys(pkg.peerDependencies || {});
+  return [...defaultExternal, ...peers];
+};
+
```
Caele, line 69: This is needed (the regex) to match for example "@qlik/api/reloads", and I don't think adding the others is an issue.

T-Wizard: you have verified that adding "@qlik/api" as peerDependencie in core package.json did not exclude it?

T-Wizard (approved):

commands/build/lib/config.js
Caele: yea I did, but after our discussion I will get rid of this anyways.

PR #1636, opened 2024-11-02
feat: add onError and onRender callbacks

Caele:
## Motivation

Adds on error and on render callbacks to the nebbie.render call.
Exposes internal render errors as RenderError + originalError

Original motivation is to make it easier for QlikEmbed to be debugged as currently any error during the loading process of the chart is just swallowed. Only tested initially with those types of errors, but could be expanded to let charts expose their own error handling as well.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/viz.js
+++ b/apis/nucleus/src/viz.js
@@ -34,7 +34,7 @@ export default function viz({
   });
 
   const createOnInitialRender = (override) => () => {
-    override && override(); // from options.onInitialRender
+    override?.(); // from options.onInitialRender
     onRenderResolve(); // internal promise in viz to wait for render
     onRender(); // from RenderConfig
   };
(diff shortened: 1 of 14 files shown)


apis/nucleus/src/viz.js
```diff
@@ -10,24 +10,33 @@ import saveSoftProperties from './utils/save-soft-properties';
...
 
   const rendered = new Promise((resolve) => {
-    onRender = resolve;
+    onRenderResolve = resolve;
   });
 
   const createOnInitialRender = (override) => () => {
-    override && override();
-    onRender();
+    override && override(); // from options.onInitialRender
```
johanlahti, line 37:
```suggestion
    override?.(); // from options.onInitialRender
```

johanlahti (approved):

quanho (approved):

niekvanstaveren (approved):

PR #1637, opened 2024-11-04
feat: add useRef

Caele:
## Motivation

Adds a useRef hook that works similarly to the preact/react counterpart.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/stardust/types/index.d.ts
+++ b/apis/stardust/types/index.d.ts
@@ -49,8 +49,7 @@ export function useMemo<T>(factory: ()=>T, deps: any[]): T;
  * renders and in it self does not trigger a render.
  * @param initialValue The initial value.
  */
-export function useRef<R>(initialValue: R): {
-};
+export function useRef<R>(initialValue: R): stardust.Ref<R>;
 
 /**
  * Runs a callback function when a dependent changes.
@@ -689,6 +688,13 @@ declare namespace stardust {
      * For example to remove any listeners added in the callback itself.
      */
     type EffectCallback = ()=>void | (()=>void);
+
+    /**
+     * Reference object returned from useRef
+     */
+    interface Ref<R> {
+        current: R;
+    }
 
     interface Rect {
         top: number;
--- a/apis/stardust/types/index.d.ts
+++ b/apis/stardust/types/index.d.ts
@@ -42,6 +42,16 @@ export function useEffect(effect: stardust.EffectCallback, deps?: any[]): void;
 export function useMemo<T>(factory: ()=>T, deps: any[]): T;
 
 /**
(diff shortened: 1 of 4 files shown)


VasilisKalaitzis (approved):
Awesome!

apis/supernova/src/hooks.js
```diff
@@ -398,6 +398,31 @@ export function useMemo(fn, deps) {
...
+ * const timesRendered = useRef(0);
+ *
+ * useEffect(() => {
+ *  render(layout);
+ *  // increments the render counter, a useState would trigger another render
+ *  timesRendered.current += 1;
+ * },[layout]);
+ *
+ */
+export function useRef(initialValue) {
```
T-Wizard, line 422: do add jsdoc to give it a proper generic type

apis/supernova/src/__tests__/hooks.test.js
```diff
@@ -300,6 +301,59 @@ describe('hooks', () => {
...
+
+      run(c);
+      expect(countValue.current).toBe(7);
+      countValue.current = 11;
+      expect(num).toBe(1);
+      await frame();
+      expect(num).toBe(1);
+      expect(countValue.current).toBe(11);
+    });
+  });
```
johanlahti, line 355: Is it possible to write a test which verifies that an updated does not trigger a re-render? (maybe not necessary…)

johanlahti (approved):

apis/supernova/src/__tests__/hooks.test.js
Caele: That last test should cover that, even if it is a bit hard/strange to test it properly. You'd basically need to test a use case where the developer used it incorrectly.

quanho (approved):

apis/stardust/types/index.d.ts
```diff
@@ -41,6 +41,17 @@ export function useEffect(effect: stardust.EffectCallback, deps?: any[]): void;
...
+/**
+ * Creates a reference to a value not needed for rendering
+ * 
+ * While Nebula does not have a virtual DOM, it is still useful
+ * to have a reference to an object that is retained across
+ * renders and in it self does not trigger a render.
+ * @param initialValue The initial value.
+ */
+export function useRef<R>(initialValue: R): {
+};
```
T-Wizard, line 53: return type is still not correct
(maybe could be fixed by adding and using a generic Ref type, similar to SetStateFn used by useState?)

Caele: tricky to get right, used an interface now

T-Wizard (approved):

PR #1670, opened 2025-01-22
fix: watch using rollup v4

### LINKED ISSUE #1471: nebula serve does not refresh visualisation/extension when changes are made to /src/index.js
## 🐛 Bug report

nebula serve does not refresh visualisation/extension when changes are made to /src/index.js

### Steps to Reproduce

1. Write some code into nebula generated component. For example element.innerHTML = '<div>Hello</div>';
2. In the CLI, nebula build
3. nebula serve
4. In the browser, enter web integration ID and select SaaS tenant.
5. Select app
6. View extension/visualisation.
7. Make some changes to /src/index.js such as element.innerHTML = '<div>Has this changed?</div>';

### Expected behavior

I would expect the visualisation to auto refresh and show 'Has this changed'.

### Actual behavior

Nothing happens. I need to repeat all of the steps above to see the changes

### Versions

- nebula.js: (4.9.0)
- Platform: Windows
- Browser: Chrome

### Additional context

Am I not doing something? The docs imply this should auto reload.

cbt1:
## Motivation

Fixes #1471 , the issue where the `watch` flag in the `build` command does not pick up file changes. This also includes the `serve` command since it's using the same underlying code.

The root cause of this issue seem to be that the `rollup-plugin-sourcemaps` plugin has not updated to support Rollup v4, which has some changes to it's Plugin API (https://rollupjs.org/migration/#changes-to-the-plugin-api).

From the tests I've done Babel source maps works fine as a replacement.

There still seem to be some issue related to `@rollup/plugin-commonjs ` where it can crash the build/watch process. But I'm not able to reproduce that consistently across the different nebula.js projects I've tested with. But this PR https://github.com/rollup/plugins/pull/1639 seem to be the culprint of that crash at least. So using v26.0.3 should still work in those projects.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/commands/build/lib/config.js
+++ b/commands/build/lib/config.js
@@ -3,7 +3,6 @@ const path = require('path');
 const babel = require('@rollup/plugin-babel');
 const postcss = require('rollup-plugin-postcss');
 const replace = require('@rollup/plugin-replace');
-const sourcemaps = require('rollup-plugin-sourcemaps');
 const json = require('@rollup/plugin-json');
 const { nodeResolve } = require('@rollup/plugin-node-resolve');
 const commonjs = require('@rollup/plugin-commonjs');
@@ -170,7 +169,7 @@ const config = ({
         babel({
           babelHelpers: 'bundled',
           babelrc: false,
-          inputSourceMap: false, // without this you get wrong source maps, but I don't know why
+          inputSourceMap: sourcemap,
           extensions,
           presets: [
             [
@@ -185,7 +184,6 @@ const config = ({
           ],
           plugins: [[jsxPlugin]],
         }),
-        ...[sourcemap ? sourcemaps() : undefined],
         postcss({}),
         ...[typescript ? typescriptPlugin() : false],
         ...[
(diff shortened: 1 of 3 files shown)


T-Wizard:
> From the tests I've done Babel source maps works fine as a replacement.

have you tested that source maps from dependencies are loaded? and not just source map for the chart itself

T-Wizard (approved):

cbt1:
> > From the tests I've done Babel source maps works fine as a replacement.
> 
> have you tested that source maps from dependencies are loaded? and not just source map for the chart itself

Just did some testing and it appears that it fails to load the sourcemaps for dependencies. For now I think we have to live with that but ideally it should be solved as well.

T-Wizard:
> > > From the tests I've done Babel source maps works fine as a replacement.
> > 
> > 
> > have you tested that source maps from dependencies are loaded? and not just source map for the chart itself
> 
> Just did some testing and it appears that it fails to load the sourcemaps for dependencies. For now I think we have to live with that but ideally it should be solved as well.

we could try using rollup-plugin-sourcemaps2
https://www.npmjs.com/package/rollup-plugin-sourcemaps2
https://github.com/2wce/rollup-plugin-sourcemaps2
that should be a version rollup-plugin-sourcemaps updated to works with rollup 4.x.x or later.

PR #1672, opened 2025-01-27
feat: add direction to listboxPopover properties

Donya-qlik:
## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

In this PR:
Regarding RTL improvement in sn-table, menu-list-search components should be reversed in RTL mode. To do so, direction as a new property has been added to ListboxPopover and the inner children such as Listbox.
The related PR in nebula-table can be found here: https://github.com/qlik-trial/nebula-tables/pull/897 
Feature Flag: CLIENT_IM_7150_TABLE_RTL. I have passed direction there to the Listboxpopover and without giving reference to that PR, you can't see the changes, so I attached a video to show how menu components are supposed to behave in RTL mode.

https://github.com/user-attachments/assets/6cf82793-9b1c-4378-9fa9-3bbe1c04e578

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxPopover.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxPopover.jsx
@@ -50,6 +50,7 @@ export default function ListBoxPopover({
       qSortByLoadOrder: 1,
     },
   ],
+  direction = 'ltr',
 }) {
   const isMasterDim = Boolean(fieldName?.qLibraryId);
   const open = show && Boolean(alignTo.current);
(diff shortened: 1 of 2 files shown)


Donya-qlik:
@linhnguyen-qlik @T-Wizard I've faced an issue in ci/circle: test-create and couldn't figure out what the problem comes from... seems it cannot create the mashup. Do you have any idea how to fix it ?

linhnguyen-qlik:
Should this menu also reflects the direction setting
Screenshot 2025-01-28 at 12 57 16

Donya-qlik:
> Should this menu also reflects the direction setting !

Oh, true! seems I missed it! will fix it.

linhnguyen-qlik (approved):
LGTM

PR #1687, opened 2025-03-11
fix: type error for properties in QAEDefinition

T-Wizard:
## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

(yarn spec have problems. I needed to run it 5-6 times to get minimal diff, before that it also include moving type around for no reasons)

### DIFF
--- a/apis/stardust/types/index.d.ts
+++ b/apis/stardust/types/index.d.ts
@@ -795,8 +795,15 @@ declare namespace stardust {
         hypercubePath: string;
     })=>stardust.ExportFormat;
 
+    type onPropertyChange = (properties: qix.GenericObjectProperties)=>void;
+
+    interface QAEProperties {
+        initial?: qix.GenericObjectProperties;
+        onChange?: stardust.onPropertyChange;
+    }
+
     interface QAEDefinition {
-        properties?: qix.GenericObjectProperties;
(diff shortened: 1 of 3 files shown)


apis/stardust/types/index.d.ts
```diff
@@ -795,8 +795,15 @@ declare namespace stardust {
         hypercubePath: string;
     })=>stardust.ExportFormat;
 
+    type onPropertyChange = (properties: qix.GenericObjectProperties)=>void;
+
+    interface QAEProperties {
+        initial?: qix.GenericObjectProperties;
+        onChange?: stardust.onPropertyChange;
+    }
```
Caele, line 803: Is this correct? Seems to diff between charts:
https://github.com/qlik-trial/sn-kpi/blob/master/src/utils/qae.js#L6
https://github.com/qlik-trial/nebula-tables/blob/main/packages/sn-table/src/qae/index.ts#L8
Nebula seems to want the "initial" part, but a bit unsure.

T-Wizard: looked again. and it looks to support both options
https://github.com/qlik-oss/nebula.js/blob/286be6bebff2aede3e29ffcdb5589f72a58e67b2/apis/supernova/src/qae.js#L125-L130

Caele (approved):

PR #1690, opened 2025-03-18
feat: cell spacing based on theme

T-Wizard:
when feature flag (VNA-13_CELLPADDING_FROM_THEME) is on set cell padding depending on if the sense theme is card or not. and add footer divider for card themes

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/Cell.jsx
+++ b/apis/nucleus/src/components/Cell.jsx
@@ -18,6 +18,7 @@ import useObjectSelections from '../hooks/useObjectSelections';
 import eventmixin from '../selections/event-mixin';
 import useStyling from '../hooks/useStyling';
 import RenderError from '../utils/render-error';
+import getPadding from '../utils/cell-padding';
 
 /**
  * @interface
@@ -538,6 +539,15 @@ const Cell = forwardRef(
       );
     }
 
+    const { cellPadding, bodyPadding } = getPadding({
+      disableCellPadding,
+      halo,
+      layout,
+      isError: state.error,
+      theme,
+      titleStyles,
+    });
+
     return (
       <Paper
         style={{
@@ -571,7 +581,7 @@ const Cell = forwardRef(
             position: 'relative',
             width: '100%',
             height: '100%',
-            ...(!disableCellPadding ? { padding: theme.spacing(1) } : {}),
+            ...(cellPadding ? { padding: cellPadding } : {}),
             ...(state.longRunningQuery ? { opacity: '0.3' } : {}),
           }}
         >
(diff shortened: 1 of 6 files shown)


apis/nucleus/src/components/Sheet.jsx
```diff
@@ -19,11 +19,16 @@ const SheetElement = {
 
 function getCellRenderer(cell, halo, initialSnOptions, initialSnPlugins, initialError, onMount, navigation, onError) {
   const { x, y, width, height } = cell.bounds;
+
+  const style = { left: `${x}%`, top: `${y}%`, width: `${width}%`, height: `${height}%`, position: 'absolute' };
+  const flags = halo.public.galaxy?.flags;
+  if (flags?.isEnabled('VNA-13_CELLPADDING_FROM_THEME')) {
+    style.boxSizing = 'border-box';
+    style.padding = '4px';
+  }
```
T-Wizard, line 28: could maybe be its own PR and/or flag.
but you probably want to turn it on at the same time

apis/nucleus/src/components/Cell.jsx
```diff
@@ -571,7 +581,7 @@ const Cell = forwardRef(
             position: 'relative',
             width: '100%',
             height: '100%',
-            ...(!disableCellPadding ? { padding: theme.spacing(1) } : {}),
+            ...(cellPadding ? { padding: cellPadding } : {}),
```
Caele, line 584: Can we move this up to the <Paper>?  That one has a className defined in the API, making it easier to change with a targeted selector.

apis/nucleus/src/components/Cell.jsx
```diff
@@ -594,6 +604,7 @@ const Cell = forwardRef(
             xs
             style={{
               height: '100%',
+              ...(bodyPadding ? { padding: bodyPadding } : {}),
```
Caele, line 607: We could add a defined className here as well, making sure it can be customized.
ex:
```
/**
 * @interface
 * @extends HTMLElement
 */
const CellBody = {
  /** @type {'njs-cell-body'} */
  className: 'njs-cell-body',
};
```

T-Wizard: done

apis/nucleus/src/components/Cell.jsx
T-Wizard: I would suggest not moving it,
As this is the old padding (that will go away when we turn of the flag)

apis/nucleus/src/components/Cell.jsx
```diff
@@ -28,6 +29,15 @@ const CellElement = {
...
 
+/**
+ * @interface
+ * @extends HTMLElement
+ */
+const CellBody = {
+  /** @type {'njs-cell-body'} */
+  className: 'njs-cell-body',
+};
+
```
Caele, line 40: Run the spec update as well

Caele (approved):

PR #1705, opened 2025-03-28
fix: sheet destroy

Caele:
## Motivation

Fix lingering event listener in sheet

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
+++ b/apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
@@ -15,21 +15,34 @@ export default function useExistingModel({ app, qId, options = {} }) {
   }
 
   useEffect(() => {
-    const handler = { cleanup: () => {} };
+    let isCleaned = false;
+    let cleanupFn = () => {};
     async function fetchObject(modelId) {
       const m = modelStore.get(modelId) || (await app.getObject(modelId));
       return m;
     }
 
+    async function awaiter() {
+      await new Promise((resolve) => {
+        setTimeout(() => {
+          resolve();
+        }, 10000);
+      });
+    }
+
     async function fetchModel() {
+      await awaiter();
       const m = await Promise.resolve(sessionModel || fetchObject(qId));
+      if (isCleaned) {
+        return;
+      }
       if (!modelStore.get(m.id)) {
         modelStore.set(m.id, m);
         const onClosed = () => {
           modelStore.clear(m.id);
         };
         m.once('closed', onClosed);
-        handler.cleanup = () => {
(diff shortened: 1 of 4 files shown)


apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
```diff
@@ -15,28 +15,30 @@ export default function useExistingModel({ app, qId, options = {} }) {
...
+          m.removeListener('closed', onClosed);
+        };
+      }
       setModel(m);
-    });
+    }
+    fetchModel();
+    return () => {
+      handler.cleanup();
+    };
```
T-Wizard, line 41: (I think) there is a possible timing problem here.
for a short live object `handler.cleanup();` could run before `handler.cleanup = ...` while waiting for the model

apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
```diff
@@ -15,28 +15,30 @@ export default function useExistingModel({ app, qId, options = {} }) {
...
-      }
-      // Nebula gets the model from this store in some of its
-      // core functions, so it must always be stored here.
-      modelStore.set(m.id, m);
-      m.once('closed', () => {
-        modelStore.clear(m.id);
-      });
-    }
-
+    const handler = { cleanup: () => {} };
```
T-Wizard, line 18: (not really important, but)
why an object an not just a function value
ex:
```suggestion
    let cleanupFn = () => {};
```
or
```suggestion
    let cleanupFn = undefined;
```

apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
Caele: hmm yea, can mark is as "cleaned"

apis/nucleus/src/components/listbox/hooks/useExistingModel.jsx
```diff
@@ -15,28 +15,44 @@ export default function useExistingModel({ app, qId, options = {} }) {
...
-    const p = Promise.resolve(sessionModel || fetchObject(qId));
-    p.then((m) => {
-      storeModel(m);
+    async function awaiter() {
+      await new Promise((resolve) => {
+        setTimeout(() => {
+          resolve();
+        }, 10000);
+      });
+    }
```
T-Wizard, line 31: remove? (I think it was only added to test things)

Caele: ah yes yes

PR #1708, opened 2025-03-31
fix: detach selection toolbar from title

jjxie:
## Motivation

Previously selection toolbar is inline with title when

- Has chart title or chart subtitle
- and chart width is more than 221 px

Which causing chart area jumping when enter and exit selection mode. Original implementation PR https://github.com/qlik-oss/nebula.js/pull/401.

The fix is detaching selection toolbar from title, always show as a popover. Also make it as same behaviour with classic chart in sense client. (The selection popover top position is slightly different than classic chart selection popover, I think is acceptable).
Tested header of listbox works as before, didn't affected by this change.

Screenshot 2025-03-31 at 10 05 44

**Before fix nebula chart:**

https://github.com/user-attachments/assets/cb0226f2-6f3d-456a-aed4-c721acc92b71

**After fix nebula chart:**

https://github.com/user-attachments/assets/1c309643-8345-4d23-8fd3-677524899b59

**Classic chart**

https://github.com/user-attachments/assets/ff6253d5-a1a6-43c6-8202-cd2501b5a1f9

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/Header.jsx
+++ b/apis/nucleus/src/components/Header.jsx
@@ -4,7 +4,6 @@ import { styled } from '@mui/material/styles';
 
 import { Grid, Tooltip, Typography } from '@mui/material';
 import ActionsToolbar from './ActionsToolbar';
-import useRect from '../hooks/useRect';
 
 const PREFIX = 'Header';
 
@@ -22,12 +21,6 @@ const StyledGrid = styled(Grid)(({ theme }) => ({
     paddingBottom: theme.spacing(1),
   },
 }));
-
-const ITEM_WIDTH = 32;
-const ITEM_SPACING = 4;
-const DIVIDER = 1;
-const NUMBER_OF_ITEMS = 6;
-const MIN_WIDTH = (ITEM_WIDTH + ITEM_SPACING) * NUMBER_OF_ITEMS + DIVIDER + ITEM_SPACING;
 
 /**
  * @interface
@@ -55,9 +48,6 @@ function Header({ layout, sn, anchorEl, hovering, focusHandler, titleStyles = {}
   const showInSelectionActions = layout.qSelectionInfo && layout.qSelectionInfo.qInSelections;
   const [actions, setActions] = useState([]);
 
-  const [containerRef, containerRect] = useRect();
-  const [shouldShowPopoverToolbar, setShouldShowPopoverToolbar] = useState(false);
-
   useEffect(() => {
     if (!sn || !sn.component || !sn.component.isHooked) {
       return;
@@ -67,20 +57,10 @@ function Header({ layout, sn, anchorEl, hovering, focusHandler, titleStyles = {}
     });
   }, [sn]);
 
-  useEffect(() => {
-    if (!containerRect) return;
-    const { width } = containerRect;
-    setShouldShowPopoverToolbar(width < MIN_WIDTH);
-  }, [containerRect]);
-
-  const showTitles = showTitle || showSubtitle;
-  const cls = [classes.containerStyle, ...(showTitles ? [classes.containerTitleStyle] : [])];
-  const showPopoverToolbar = (hovering || showInSelectionActions) && (shouldShowPopoverToolbar || !showTitles);
-  const showToolbar = showTitles && !showPopoverToolbar && !shouldShowPopoverToolbar;
+  const showPopoverToolbar = hovering || showInSelectionActions;
 
   const Toolbar = (
     <ActionsToolbar
-      show={showToolbar}
       selections={{
         show: showInSelectionActions,
         api: sn.component.selections,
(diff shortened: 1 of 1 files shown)


Caele:
So this would fully remove the inline toolbar right? And that is the same that we always do inside client? ok!
@veinfors did last look at this in client, so best he checks as well.

apis/nucleus/src/components/Header.jsx
```diff
@@ -94,7 +74,7 @@ function Header({ layout, sn, anchorEl, hovering, focusHandler, titleStyles = {}
   );
 
   return (
-    <StyledGrid ref={containerRef} item container wrap="nowrap" className={cls.join(' ')}>
+    <StyledGrid item container wrap="nowrap" className={classes.containerStyle}>
```
veinfors, line 77: This means we do not use `containerTitleStyle` any longer, so that should be removed.
However, it alters the styling slightly for charts - more room is given to the "chart area" (8px more in height). I think it's better and the right thing to do. The question is how much impact it would mean for customers though 🤔

veinfors:
I like this change because it improves the experience by removing that "jumpiness" as you describe and streamlines the behaviour across chart sizes. It also reduces the complexity of the code and gets rid of some useEffect which is good.

The only thing I'm concerned about is that we might affect the visual rendering of existing charts with this change by giving more height to the "chart area" for charts with titles...

apis/nucleus/src/components/Header.jsx
jjxie: `containerTitleStyle` is still used in `StyledGrid` which is the wrapper of title and subtitle area. https://github.com/qlik-oss/nebula.js/blob/main/apis/nucleus/src/components/Header.jsx#L21

jjxie:
> I like this change because it improves the experience by removing that "jumpiness" as you describe and streamlines the behaviour across chart sizes. It also reduces the complexity of the code and gets rid of some useEffect which is good.
> 
> The only thing I'm concerned about is that we might affect the visual rendering of existing charts with this change by giving more height to the "chart area" for charts with titles...

Yea, it will affect chart size when doing selections 

> I like this change because it improves the experience by removing that "jumpiness" as you describe and streamlines the behaviour across chart sizes. It also reduces the complexity of the code and gets rid of some useEffect which is good.
> 
> The only thing I'm concerned about is that we might affect the visual rendering of existing charts with this change by giving more height to the "chart area" for charts with titles...

Yea, I made a change to keep current title padding. So shouldn't change existing chart area now.
I noticed sense client title padding to bottom is 10px... Nebula chart will still look different with classic chart in new sheet. Not sure what we gonna do with it.

veinfors:
> > I like this change because it improves the experience by removing that "jumpiness" as you describe and streamlines the behaviour across chart sizes. It also reduces the complexity of the code and gets rid of some useEffect which is good.
> > The only thing I'm concerned about is that we might affect the visual rendering of existing charts with this change by giving more height to the "chart area" for charts with titles...
> 
> Yea, I made a change to keep current title padding. So shouldn't change existing chart area now. I noticed sense client title padding to bottom is 10px... Nebula chart will still look different with classic chart in new sheet. Not sure what we gonna do with it.

Your mean there will be a diff of 2px then? That sounds like something we want to adjust then, but not in this pr. Best to have a separate discussion/investigation about what is the best approach to that problem.

veinfors (approved):

jjxie:
> > > I like this change because it improves the experience by removing that "jumpiness" as you describe and streamlines the behaviour across chart sizes. It also reduces the complexity of the code and gets rid of some useEffect which is good.
> > > The only thing I'm concerned about is that we might affect the visual rendering of existing charts with this change by giving more height to the "chart area" for charts with titles...
> > 
> > 
> > Yea, I made a change to keep current title padding. So shouldn't change existing chart area now. I noticed sense client title padding to bottom is 10px... Nebula chart will still look different with classic chart in new sheet. Not sure what we gonna do with it.
> 
> Your mean there will be a diff of 2px then? That sounds like something we want to adjust then, but not in this pr. Best to have a separate discussion/investigation about what is the best approach to that problem.

Sounds good. I will compare the title and chart area, will report an issue if see difference

apis/nucleus/src/components/Header.jsx
```diff
@@ -67,20 +57,12 @@ function Header({ layout, sn, anchorEl, hovering, focusHandler, titleStyles = {}
...
-
   const showTitles = showTitle || showSubtitle;
   const cls = [classes.containerStyle, ...(showTitles ? [classes.containerTitleStyle] : [])];
-  const showPopoverToolbar = (hovering || showInSelectionActions) && (shouldShowPopoverToolbar || !showTitles);
-  const showToolbar = showTitles && !showPopoverToolbar && !shouldShowPopoverToolbar;
+  const showPopoverToolbar = hovering || showInSelectionActions;
 
   const Toolbar = (
     <ActionsToolbar
-      show={showToolbar}
```
T-Wizard, line 83: maybe a `show={false}` (default is true)
in case there are any object that tries to show actions when not in selection mode

jjxie: Yea, ok. Not sure what scenario will be like that, but I will add it

T-Wizard (approved):

PR #1710, opened 2025-04-02
fix: add filters info in footnote

jjxie:
## Motivation

Added support of showing filters information in footnote. 
`generateSetExpression.js` is same check for classic chart

**None card theme**
Screenshot 2025-04-29 at 15 58 30

**Card theme**
Screenshot 2025-04-29 at 15 58 51

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


T-Wizard:
wrong footer size when theme is a card theme and `VNA-13_CELLPADDING_FROM_THEME` is on

apis/nucleus/src/utils/generateSetExpression.js
```diff
@@ -0,0 +1,234 @@
+const FilterType = {
```
Caele, line 1: This file should have tests, I see that there are some for the original one, so bring them over.

jjxie:
> wrong footer size when theme is a card theme and `VNA-13_CELLPADDING_FROM_THEME` is on

I added padding and border top style for filter footnote with card theme. 

Removed padding top when new sheet flag is enabled (previously 8px). Padding top for non card theme, like sense classic is 6px, for card theme is 0px. I am not sure if it is always the case, or can change padding top in themes. I noticed font size and style is not applied for some themes, both for normal footnote text or filter information. I am not sure if it is related to your changes of applying theme. I will check, but will not make change in this PR.

apis/nucleus/src/components/listbox/utils/frequencyMaxUtil.js
```diff
@@ -1,12 +1,4 @@
-const escapeField = (field) => {
-  if (!field) {
```
jjxie, line 2: This line the new file check `if (!field || field === ']')`. I guess it is fine to use new file logic. @T-Wizard

T-Wizard: OK.
if I remember correctly `field === ']'` was from engine making it impossible to use a field with the name `]`
but now it looks like it works both with and without escape

apis/nucleus/src/utils/cell-padding.js
```diff
@@ -41,7 +41,8 @@ const getPadding = ({ layout, isError, senseTheme, titleStyles }) => {
   if (isCardTheme) {
     const showTitle = layout?.showTitles && !!layout?.title;
     const showSubtitle = layout?.showTitles && !!layout?.subtitle;
-    const showFootnote = layout?.showTitles && !!layout?.footnote;
+    const hasFilters = layout?.filters?.length > 0 && layout?.qHyperCube?.qMeasureInfo?.length > 0;
+    const showFootnote = layout?.showTitles && (!!layout?.footnote || hasFilters);
```
T-Wizard, line 45: also need to do `const filtersFootnoteString = generateFiltersString(layout?.filters ?? [], translator);` and check if filtersFootnoteString is empty string or not

apis/nucleus/src/components/Footer.jsx
```diff
@@ -30,15 +29,21 @@ const CellFooter = {
...
+  const newSheetEnabled = flags?.isEnabled('VIZ-7_NEW_SHEET');
 
   return layout && layout.showTitles && (layout.footnote || showFilters) ? (
     <StyledGrid container>
-      <Grid item className={classes.itemStyle} data-testid={CellFooter.className}>
+      <Grid
+        item
+        className={classes.itemStyle}
+        data-testid={CellFooter.className}
+        style={{ paddingTop: (theme) => (newSheetEnabled ? 0 : theme.spacing(1)) }}
```
T-Wizard, line 45: from my testing.
For classic theme (not cards) it should always have spacing.
and for cards theme 0 is to small, maybe 1px?

apis/nucleus/src/components/FiltersFooter.jsx
```diff
@@ -0,0 +1,63 @@
...
+      >
+        <Grid item display="flex">
+          <FilterIcon style={{ fontSize: '12px', color: footerStyle.color, margin: 'auto' }} />
+          <ItalicText styles={{ ...styles, marginLeft: '2px' }}>{translator.get('Object.FiltersApplied')}</ItalicText>
+          <ItalicText styles={styles}> &nbsp;</ItalicText>
+        </Grid>
+        <Grid item display="flex">
+          {filtersFootnoteLabels.map((filter) => (
+            <Grid container wrap="nowrap" key={`${filter.field}-${filter.label}`}>
+              <BoldText styles={styles}>{`${filter.field}:`}</BoldText>
```
T-Wizard, line 51: when I compare with sense-client this should be bold and italic

T-Wizard (approved):

apis/nucleus/src/components/FiltersFooter.jsx
```diff
@@ -0,0 +1,55 @@
...
+          <ItalicText styles={{ ...styles, marginLeft: '2px' }}>{translator.get('Object.FiltersApplied')}</ItalicText>
+          <ItalicText styles={styles}> &nbsp;</ItalicText>
+        </Grid>
+        <Grid item display="flex">
+          {filtersFootnoteLabels.map((filter) => (
+            <Grid container wrap="nowrap" key={`${filter.field}-${filter.label}`}>
+              <ItalicText styles={{ ...styles, fontWeight: 'bold' }}>{`${filter.field}:`}</ItalicText>
+              <ItalicText styles={styles}> &nbsp;</ItalicText>
+              <ItalicText styles={styles}> {filter.label}</ItalicText>
+              <ItalicText styles={styles}> &nbsp;</ItalicText>
```
Caele, line 46: These 3 all have the same style, can't they be a single on instead?

apis/nucleus/src/components/Footer.jsx
```diff
@@ -28,15 +29,37 @@ const CellFooter = {
   className: 'njs-cell-footer',
 };
 
-function Footer({ layout, titleStyles = {} }) {
-  return layout && layout.showTitles && layout.footnote ? (
+function Footer({ layout, titleStyles = {}, translator, flags, isCardTheme }) {
+  const footerStyle = titleStyles.footer;
+  const hasFilters = layout?.filters?.length > 0 && layout?.qHyperCube?.qMeasureInfo?.length > 0;
+  const filtersFootnoteString = generateFiltersString(layout?.filters ?? [], translator);
```
Caele, line 35: You have these two functions: generaFiltersString and generateFiltersLabels, but one is called out here in Footer and one inside FiltersFooter. I think it would be cleaner if all that is done inside the FiltersFooter.
I see that you need it to check if the FiltersFooter should exist or not, but you can do that check inside the component instead, which would clean up the Footer quite a bit.

apis/nucleus/src/utils/generateSetExpression.js
Caele, line 1: Most of this file isn't used right? So we should not keep those things here. Better clean it out together with those tests.

apis/nucleus/src/components/FiltersFooter.jsx
jjxie: Done

apis/nucleus/src/utils/generateSetExpression.js
jjxie: Most of this file is used, but `generateSetExpression` function is not used. I removed this function and moved some tests were testing `generateSetExpression` before to test label strings.

apis/nucleus/src/components/Footer.jsx
Caele: its ok

Caele (approved):

PR #1711, opened 2025-04-04
feat: add viewDataToggled and toggleFocus

linhnguyen-qlik:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/nucleus/src/viz.js
```diff
@@ -203,6 +203,15 @@ export default function viz({
         cellRef.current.setModel(newModel);
       }
     },
+    get viewDataToggled() {
+      return viewDataObjectId !== undefined;
+    },
+    supportViewData() {
+      return cellRef.current.supportViewData();
+    },
```
Caele, line 211: Remove this, we need to check it elsewhere.

apis/supernova/src/creator.js
```diff
@@ -224,6 +229,9 @@ function createWithHooks(generator, opts, galaxy) {
     getImperativeHandle() {
       return generator.component.getImperativeHandle(this);
     },
+    supportViewData() {
+      return generator.definition.ext.support.viewData || false;
+    },
```
Caele, line 234: remove

apis/nucleus/src/components/Cell.jsx
```diff
@@ -470,6 +470,9 @@ const Cell = forwardRef(
         getQae() {
           return state.sn.generator.qae;
         },
+        getExtDefinition() {
+          return state.sn.generator.definition.ext;
```
linhnguyen-qlik, line 474: When the chart is in view data the definition I guess coming from the table and therefor viewData is false.

linhnguyen-qlik: Somehow when I have boxplot the `ext` is undefined meanwhile in sense it has data.

apis/nucleus/src/viz.js
```diff
@@ -10,6 +10,18 @@ import saveSoftProperties from './utils/save-soft-properties';
 
 const noopi = () => {};
 
+function support(prop, supportObject, layout) {
+  // we want to you ext from object instead. Why using $scope.ext what is the different?
```
veinfors, line 14:
```suggestion
```

apis/nucleus/src/components/Cell.jsx
```diff
@@ -471,6 +470,12 @@ const Cell = forwardRef(
         getQae() {
           return state.sn.generator.qae;
         },
+        getExtDefinition() {
```
veinfors, line 473: It says "getExtensionDefinition" in the "--- Viz api ---" comment below...

Caele: yea this is inside the cell, haven't added it to the Viz one yet, going to hold off on it a bit as it requires us to resolve the property panel definition. Its not strictly needed yet either.

apis/nucleus/src/viz.js
```diff
@@ -237,6 +283,34 @@ export default function viz({
...
+      if (type === 'viewData' && viewDataObjectId !== undefined) {
+        return true;
+      }
+      const renderState = cellRef.current.getRenderState();
+      // Could be cleaner, but need to check that its rendered properly
+      if (mountedReference && renderState.loaded && !renderState.error && successfulRender) {
+        const supportObject = cellRef.current.getExtDefinition()?.support;
+        if (supportObject) {
+          // Passing in layout here should only be relevant if the chart has rendered
+          return support(type, supportObject, model?.layout);
```
T-Wizard, line 303: Info:
for model.layout to work it needs enigma.js mixins (added by qlik-embed / sense-client)
but if used in context without the mixins model.layout will always be undefined.

Caele: I can move it into Cell, that makes it simpler

apis/nucleus/src/viz.js
```diff
@@ -237,6 +272,28 @@ export default function viz({
...
+    /**
+     *  valid types: viewData, cssScaling, snapshot, exportData, exploration
+     *  questionable types: supportRefresh, quickMobile, fullscreen
+     *  deprecated?: sharing
+     *
+     */
+    support(type) {
+      if (type === 'viewData' && viewDataObjectId !== undefined) {
+        return true;
+      }
```
T-Wizard, line 285: why only for type viewData?

there may be cases where we want more types to return support for the original type even when data view is toggle on. (maybe needed for download in hover menu, but not sure as I have not found any object type that support data view but not download)

Caele: hmm, depends on how its used. in general I'd think you want to get the support response of the original chart always. But it also doesn't make much sense to check most things on the flipped chart.
The way its structured in nebula with toggleDataView makes it a bit tricky to return the original support object, would have to store it. Right now I'm not sure what the expected outcome is, so best leave it as is, and check back once we start using it and before adding it to the public API.

T-Wizard: consider if we should make it an error to call support in data view mode to prevent anyone from depending on it and make it more open for later changes

linhnguyen-qlik: a work around is maybe just return `viewDataObjectId != undefined`

Caele: i think an error is good to begin with, makes it easy to root out potential bugs instead of just returning always false for example.

T-Wizard (approved):
OK.

I am a bit unsure about `setOnBlurHandler`
I would like something that is easier to understand and use, but don't have any suggestion on what that would be

apis/nucleus/src/viz.js
```diff
@@ -82,6 +85,31 @@ export default function viz({
...
+  ---additions---
+  support: (type) => boolean;
+  getExtensionDefinition: () => object;
+      propertyPanelDef: extension.definition.ext.definition,
+      initialProperties: extension.qae.properties.initial,
+      data: extension.qae.data,
+
+  toggleFocus: v?.toggleFocus,
+  setOnBlurHandler: v?.setOnBlurHandler,
+  onContextMenu: v?.onContextMenu,
```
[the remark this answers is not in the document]
Caele: yea, later

Caele:
> OK.
> 
> I am a bit unsure about `setOnBlurHandler` I would like something that is easier to understand and use, but don't have any suggestion on what that would be

I agree, its messy. The whole thing depends on how we handle keyboard nav, either internally in Nebula or externally. And as we don't run a single Nebula instance in embed we need to keep it external (also as we need the sheet to control the ordering of the charts when tabbing/navigating). There might be a combined way we can simplify things, Lihn is fighting it.
But I think we can go ahead with this one for now.

Caele (approved):
@linhnguyen-qlik think you can merge when you are satisfied, we'll need to continue tinker with parts of this anyways.

apis/nucleus/src/viz.js
linhnguyen-qlik: Throwing an error is actually a problem. Let say toggled to view data of a barchart and I jump to another object. The next time I am at the barchart again and press space this time the barchart is already in view data. Therefore this check throw an error  
`const shouldToggleDataView = !focusedObjectApi.viewDataToggled && focusedObjectApi.support("viewData");`
Instead of giving me `shouldToggleDataView` so I can go and toggleFocus instead. 
https://github.com/qlik-trial/nebula-containers/blob/1251cf75d12ebfc7e154bc36923ec530b1920ccf/packages/sn-layout-container/src/hooks/use-interactions.ts#L325

I still vote for returning  `viewDataObjectId != undefined` when it is already in viewData.

apis/nucleus/src/viz.js
```diff
@@ -237,6 +272,28 @@ export default function viz({
...
+    /**
+     *  valid types: viewData, cssScaling, snapshot, exportData, exploration
+     *  questionable types: supportRefresh, quickMobile, fullscreen
+     *  deprecated?: sharing
+     *
+     */
+    support(type) {
+      if (viewDataObjectId !== undefined) {
+        throw new Error('Cannot call support when in data view');
+      }
```
linhnguyen-qlik, line 285: This is not ok. It throws the error on any type when the chart is in view data mode. Will cause code for other things not working

Caele: thats the idea, you are not supposed to call this when you are in viewDataMode as it will return incorrect settings. 
What would you propose instead?

linhnguyen-qlik: But then we have to check if we are in viewData in conjunction with the usage of `support` function every time we want to use the support function. It is kind of strange to have two functions that always have to use together. I don't have an idea yet. I will think about this. 
```
let support = false;
if (!viz.viewDataToggled) {
    support = viz.support("type");
}
```

Caele: The problem is to know what we should return, either always the settings for the base chart or for the table? What would we expect?

linhnguyen-qlik: I think it should always return the base chart and not the table.

linhnguyen-qlik: Since you can't do any thing in context menu when you are in view data we are not gonna need to know anything related to "table extension" so I think it should alway return things for the original chart type just like it is know in sense-client.

Caele: fair, i'll change to that

linhnguyen-qlik:
@Caele  The strange thing is that boxplot doesn't have definition.ext but in production it could still toggleViewData. Just comment here so we don't forget to check this out. I am going to merge this for now.

PR #1715, opened 2025-04-06
chore: add hypercube generic functions - part02

Donya-qlik:
## Motivation

Moving generic hypercube functions from sense-client to nebula.js - PART02
In this PR the functions that are required to **add field and alternative fields**, are considered.

P.S: (Part01 can be found [here](https://github.com/qlik-oss/nebula.js/pull/1716))

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/supernova/src/handler/data-property-handler.js
```diff
@@ -69,6 +173,105 @@ class DataPropertyHandler {
...
+  }
+
+  static autoSortMeasure() {
+    throw notSupportedError;
+  }
+
+  createExpressionMeasure(expression, label, defaults) {
+    const measure = merge({}, this.measureProperties || {}, defaults || {});
+
+    const meas = { ...measure };
```
cbt1, line 192: `merge()` returns a new object. Should be no need to create clone of that.

apis/supernova/src/handler/data-property-handler.js
```diff
@@ -69,6 +173,105 @@ class DataPropertyHandler {
...
+    const measures = args.map(({ expression, label, defaults }) =>
+      this.createExpressionMeasure(expression, label, defaults)
+    );
+    return this.addMeasures(measures);
+  }
+
+  createLibraryMeasure(id, defaults) {
+    const measure = merge({}, this.measureProperties || {}, defaults || {});
+
+    const meas = { ...measure };
```
cbt1, line 219: merge() returns a new object. Should be no need to create clone of that.

apis/supernova/src/handler/data-property-handler.js
Donya-qlik: Ah, true! I'll remove it.

cbt1:
This is mostly just copy/pasted right? Have you done any major refactoring changes?

apis/supernova/src/handler/utils/field-helper/expand-field-derived-data/index.js
Caele, line 1: Could we change all these folders with index files into a more flat structure? To better match the rest of Nebula, it has more named files and no /folder/index.js setups.

cbt1 (commented):
There are a couple of async functions and usage of Promise that I don't understand the usage of. Is there an reason behind it?

EDIT: Why is there a need to use async/Promise at all in the handler?

Donya-qlik:
> There are a couple of async functions and usage of Promise that I don't understand the usage of. Is there an reason behind it?
> 
> EDIT: Why is there a need to use async/Promise at all in the handler?

Yes, I see. In sense-client they all return promises, but using AngularJs promises, so I kept them as they were, to avoid breaking anything. I tried removing a couple of them, but the function failed, so it seems they are needed to ensure the data is fetched properly (e.x: adding field and sorting the orders at the right time). but I'll check those you left comment on, maybe I can remove those without facing any problem.

apis/supernova/src/handler/utils/hypercube-helper/hypercube-utils.js
```diff
@@ -98,24 +88,6 @@ function insertMainDimension(self, dimension, dimensions, idx) {
...
-export function addMainDimension(self, dimension, index) {
-  const dimensions = self.getDimensions();
-  const idx = index ?? dimensions.length;
-
-  if (dimensions.length < self.maxDimensions()) {
-    return insertMainDimension(self, dimension, dimensions, idx);
-  }
-
-  return Promise.resolve();
-}
```
Donya-qlik: Moved it to a separated file `add-main-dimension.js` because of having a nested function

apis/supernova/src/handler/utils/hypercube-helper/hypercube-utils.js
```diff
@@ -63,16 +63,6 @@ export function setPropForLineChartWithForecast(self) {
...
-export function insertFieldAtIndex(self, field, alternative, currentFields, index = undefined) {
-  if (alternative || (self.maxDimensions() <= currentFields.length && currentFields.length < TOTAL_MAX.DIMENSIONS)) {
-    const fields = self.hcProperties.qLayoutExclude.qHyperCubeDef.qDimensions;
-    const idx = index ?? fields.length;
-    fields.splice(idx, 0, field);
-    return Promise.resolve(field);
-  }
-  return undefined;
-}
-
```
Donya-qlik: I had created this function as a helper fn, but noticed that is redundant.

apis/supernova/src/handler/utils/hypercube-helper/hypercube-utils.js
```diff
@@ -98,24 +88,6 @@ function insertMainDimension(self, dimension, dimensions, idx) {
...
-  return Promise.resolve();
-}
-
-export function insertDimensionAtIndex(self, dimension, alternative, currentDimensions, index = undefined) {
-  const dimensions = self.hcProperties.qLayoutExclude.qHyperCubeDef.qDimensions;
-  const idx = index ?? dimensions.length;
-  dimensions.splice(idx, 0, dimension);
-  return Promise.resolve(dimension);
-}
-
```
Donya-qlik: The same here, it is redundant, so I removed it.

apis/supernova/src/handler/utils/hypercube-helper/hypercube-utils.js
```diff
@@ -163,31 +135,20 @@ export function addAlternativeMeasure(self, measure, index = undefined) {
...
-  const measures = self.getMeasures();
-  const idx = index ?? measures.length;
-
-  if (measures.length < self.maxMeasures()) {
-    return insertMainMeasure(measure, measures, idx);
-  }
-
-  return Promise.resolve();
-}
-
```
Donya-qlik: Moved it to a separated file `add-main-measure.js`

Donya-qlik:
@cbt1 @Caele could you take another look to my PR if you have some time, thanks.

cbt1 (commented):
> Yes, I see. In sense-client they all return promises, but using AngularJs promises, so I kept them as they were, to avoid breaking anything. I tried removing a couple of them, but the function failed, so it seems they are needed to ensure the data is fetched properly (e.x: adding field and sorting the orders at the right time). but I'll check those you left comment on, maybe I can remove those without facing any problem.

I can understand the concern to reduce risk of breaking something. But if there is no async operation in this code. The only other risk I can think of now is some outside code that consumes the code where it calls a method and expect it to be async, maybe it will call for example like this `handler.someMethod().then(() => // Do something after method resolves)`

apis/supernova/src/handler/utils/hypercube-helper/hypercube-utils.js
```diff
@@ -0,0 +1,182 @@
...
+}
+
+export function insertMainDimension(self, dimension, dimensions, idx) {
+  dimensions.splice(idx, 0, dimension);
+
+  return self.autoSortDimension(dimension).then(() => {
+    arrayUtil.indexAdded(self.hcProperties.qInterColumnSortOrder, self.getDimensions().length + dimension.length - 1);
+
+    if (typeof self.dimensionDefinition.add === 'function') {
+      return Promise.when(self.dimensionDefinition.add.call(null, dimension, self.properties, self));
```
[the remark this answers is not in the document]
Donya-qlik: Oh Sorry, I missed to fix it. 
I splitted this function in my next PR, into two for different type of orders, sort-order and column-order, there it's correct and added some test cases for that but not here.

The fact is this function is not called when I normally add fields so I didn't get any incorrect value or error even when I tested adding-fields manually. 

If you don't mind, in case of adding test cases, I'll go for the refactored version of this and the testcases in the next PR since I have already covered it there.

apis/supernova/src/handler/hypercube-handler.js
```diff
@@ -70,10 +54,52 @@ class HyperCubeHandler extends DataPropertyHandler {
...
+
+        if (hcHelper.isDimensionAlternative(this, alternative)) {
+          const altDim = await hcHelper.addAlternativeDimension(this, dim);
+          addedDimensions.push(altDim);
+        } else if (existingDimensions.length < this.maxDimensions()) {
+          await hcHelper.addActiveDimension(this, dim, existingDimensions, addedDimensions, addedActive);
+          addedActive++;
+        }
+        return addedDimensions;
+      })
```
T-Wizard, line 91: I do expect addDimensions to add thing in same order as dimensions array.
this looks to add them in the order of what is done first.

cbt1: Good catch. But I do think it does add them in the correct order since the "awaited" call is not actually an async function and gets resolved directly.

Either way, the calls inside `addDimensions` does have to be async. But the method itself might still need to be, since it's possible called as an async function but some code that consumes it.

cbt1: Turns out `addActiveDimension` is making something that looks like an async call. So you're right, there is a risk of a race condition here.

I would then think `for await (const value of object) {}` is the way to go. Instead of using `dimensions.map`.

apis/supernova/src/handler/hypercube-handler.js
```diff
@@ -70,10 +54,55 @@ class HyperCubeHandler extends DataPropertyHandler {
...
+
+    return addMainDimension(this, dim, idx);
+  }
+
+  async addDimensions(dimensions, alternative = false) {
+    const existingDimensions = this.getDimensions();
+    const addedDimensions = [];
+    let addedActive = 0;
+
+    await dimensions.reduce(async (prevPromise, dimension) => {
```
cbt1, line 76: While using `reduce` in this way will work. I do think you should use `for await` loop instead as it's easier to read and built to do what you want.

apis/supernova/src/handler/hypercube-handler.js
Donya-qlik: @T-Wizard @cbt1 I ended up using `await array.reduce()` instead. Apparently for await requires regenerator-runtime, which is too heavyweight for it to allow as Eslint error says. The best replacement which follows the same concept is reduce, it waits for each previous round promise to be resolved and then it goes to the next one.

apis/supernova/src/handler/hypercube-handler.js
cbt1: I seem like a hack to me. If someone in the future where to look at this they might wonder why reduce is being used.

IMO it's fine to just ignore the lint error. The error is not valid in this case.

apis/supernova/src/handler/utils/field-helper/field-utils.js
```diff
@@ -0,0 +1,87 @@
...
+    } else {
+      // Default auto sorting
+      sortCriterias.qSortByNumeric = 1;
+      sortCriterias.qSortByAscii = 1;
+    }
+
+    if (!dim.qDef.qSortCriterias) {
+      const updatedDimension = { ...dim };
+      updatedDimension.qDef.qSortCriterias = [sortCriterias];
+      Object.assign(dim, updatedDimension);
```
cbt1, line 65: I don't understand the usage of `Object.assign` here? Why not just assigned directly on `dim.qDef.qSortCriterias` instead?

cbt1 (approved):
Just had some final comments. This LGTM now. But it is kind of hard to tell what has changed in this move. So for upcoming PR:s maybe change as little as possible and then in a follow-up PR make refactoring changes?

apis/supernova/src/handler/utils/field-helper/field-utils.js
Donya-qlik: Umm... true! couldn't remember why I did this... I'll change it.

apis/supernova/src/handler/hypercube-handler.js
```diff
@@ -11,46 +16,25 @@ class HyperCubeHandler extends DataPropertyHandler {
 
   setProperties(properties) {
     if (!properties) {
-      return;
+      return {};
```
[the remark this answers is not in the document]
cbt1: This works fine for me without any lint error

```ts
setProperties(properties) {
    if (!properties) {
      return;
    }

    super.setProperties(properties);

    this.hcProperties = this.path ? utils.getValue(properties, `${this.path}.qHyperCubeDef`) : properties.qHyperCubeDef;

    if (!this.hcProperties) {
      return;
    }

    hcHelper.setDefaultProperties(this);
    hcHelper.setPropForLineChartWithForecast(this);

    // Set auto-sort property (compatibility 0.85 -> 0.9),
    // can probably be removed in 1.0
    this.hcProperties.qDimensions = hcHelper.setFieldProperties(this.hcProperties.qDimensions);
    this.hcProperties.qMeasures = hcHelper.setFieldProperties(this.hcProperties.qMeasures);
  }
```

Donya-qlik:
> Just had some final comments. This LGTM now. But it is kind of hard to tell what has changed in this move. So for upcoming PR:s maybe change as little as possible and then in a follow-up PR make refactoring changes?

Thanks @cbt1 for reviewing my PR I totally understand. These are not used anywhere in this repo to give a reference how they used to work. This one was the largest PR, hopefully the next one will be easier to review, however there also I made a few changes, but it is smaller.

PR #1716, opened 2025-04-06
refactor: add hypercube generic functions to nebuls.js, part01

Donya-qlik:
## Motivation

Moving generic hypercube functions from sense-client to nebula.js - Part01.
At this PR only the functions that are required to display the fields in property-panel are considered. The rest of the functions will be moved to this repo in the coming PRs.

Required Flag: NEBULA_DATA_HANDLERS

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/stardust/api-spec/listbox-spec.json
```diff
@@ -189,4 +189,4 @@
       }
     }
   }
-}
```
Donya-qlik, line 192: I'll fix it

apis/nucleus/src/utils/handlers/data-property-handlers.js
```diff
@@ -0,0 +1,83 @@
+import { getFieldById } from './handlers-helpers';
+
+class DataPropertyHandler {
```
Donya-qlik, line 3: The class source is in sense-client [data-hypercube-handler.ts](https://github.com/qlik-trial/sense-client/blob/main/web/assets/objects/utils/data-properties/data-property-handler.ts)

apis/nucleus/src/utils/handlers/hypercube-handlers.js
```diff
@@ -0,0 +1,102 @@
+import utils from '@nebula.js/conversion/src/utils';
+import DataPropertyHandler from './data-property-handlers';
+import { getHyperCube, setFieldProperties } from './handlers-helpers';
+
+class HyperCubeHandler extends DataPropertyHandler {
```
Donya-qlik, line 5: The class source is in sense-client [hypercube-handler.ts](https://github.com/qlik-trial/sense-client/blob/main/web/assets/objects/utils/data-properties/hypercube-handler.js)

apis/supernova/src/generator.js
```diff
@@ -46,6 +47,8 @@ export default function generatorFn(UserSN, galaxy) {
 
   if (typeof UserSN === 'function') {
     sn = UserSN(galaxy);
+    sn.ext = sn.ext || {};
+    sn.ext.handlers = (opts) => new HyperCubeHandler(opts);
```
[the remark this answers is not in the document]
Donya-qlik: To be able later to fetch the data from it in sense-client.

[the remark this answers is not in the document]
Caele: So is this a new property on the ext object?

apis/supernova/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "src/index.js",
   "devDependencies": {
     "extend": "3.0.2",
-    "node-event-emitter": "0.0.1"
+    "node-event-emitter": "0.0.1",
+    "@nebula.js/nucleus": "^5.12.4"
```
Caele, line 9: this would make it a dependency loop, need to change how this is setup. Move the whole hypercube handler stuff into the supernova package instead?

apis/supernova/src/generator.js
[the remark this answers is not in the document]
Donya-qlik: yes

apis/supernova/package.json
Donya-qlik: ok, I'll try it.

apis/supernova/src/generator.js
```diff
@@ -46,6 +47,8 @@ export default function generatorFn(UserSN, galaxy) {
 
   if (typeof UserSN === 'function') {
     sn = UserSN(galaxy);
+    sn.ext = sn.ext || {};
+    sn.ext.dataHandler = (opts) => new HyperCubeHandler(opts);
```
Caele, line 51: This looks a bit odd, why would sn.ext be undefined?

Currently I think you find this further down already, on the generator:
This part:
```
  Object.keys(sn).forEach((key) => {
    if (!generator[key]) {
      generator.definition[key] = sn[key];
    }
  });
```

Should put ext under definition along with the dataHandler:
```
{
  qae: ...,
  component: ...,
  create:()...,
  definition: {
    ext: {
      ...,
      dataHanlder: ...
    },
  )
}

```

But maybe we should explicitly put it on the definition instead, to make it a bit more clear;

```
  const generator = /** @lends generator */ {
    /**
     * @type {QAE}
     */
    qae: qae(sn.qae),
    /**
     * @type {SnComponent}
     */
    component: sn.component || {},
    /**
     * @param {object} p
     * @param {EnigmaAppModel} p.app
     * @param {EnigmaObjectModel} p.model
     * @param {ObjectSelections} p.selections
     */
    create(params) {
      const ss = create(generator, params, galaxy);
      return ss;
    },
    definition: {
      handler: ... <-- add it here?
    },
  };

```

Donya-qlik: Done!

apis/supernova/src/generator.js
```diff
@@ -73,7 +74,9 @@ export default function generatorFn(UserSN, galaxy) {
       const ss = create(generator, params, galaxy);
       return ss;
     },
-    definition: {},
+    definition: {
+      dataHandler: (opts) => new HyperCubeHandler(opts),
+    },
```
Caele, line 79: How do we solve this later:
https://github.com/qlik-trial/nebula-tables/blob/main/packages/sn-pivot-table/src/ext/index.ts#L17
We will be moving them into each respective chart, so how do we make sure they get picked up correctly? Will that be done here somewhere as well? (later)

Caele: Also think we want to flag this, to avoid issues and keep it limited for the time being.

Donya-qlik: In case of the pivot table, there is actually a task to move the related handlers to nebula-table so I guess they will be handled there, haven't looked into it yet but I'll soon.

Yes, the flag is on the way, don't have access to create by myself but have asked for that.

Donya-qlik: In case of pivot table, there is actually a task to move the related handlers to nebula-table so I guess they will be handled there, haven't looked into it yet but I'll soon.

Donya-qlik: @Caele Btw, do you mean we put the feature flag in Nebula as an open-source repo? it's a question @enell came up with.

Caele (approved):
Can go ahead with this now. Not sure about the extra line in the spec file. but thats probably an auto fix thing.

PR #1717, opened 2025-04-08
revert: react-virtualized-auto-sizer and ignore version updates

Caele:
## Motivation

The new auto-sizer version causes issues with the popover listbox, haven't been able to sort out how to resolve it so reverting and locking that version for now.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/package.json
+++ b/apis/nucleus/package.json
@@ -21,7 +21,7 @@
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "react-test-renderer": "18.3.1",
-    "react-virtualized-auto-sizer": "1.0.26",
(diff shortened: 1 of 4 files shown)


T-Wizard (approved):

T-Wizard:
> ## Motivation
> The new auto-sizer version causes issues with the popover listbox, haven't been able to sort out how to resolve it so reverting and locking that version for now.

yes some kind of odd timing issue.
tried to debug why auto-sizer give another size, and if I add a breakpoint before the getBoundingClientRect call it works.
(but not sure why or how to fix it. If the element actually change size I would expect the ResizeObserver to trigger but it don't)

PR #1730, opened 2025-04-29
fix: allow for external focus management

Caele:
## Motivation

When rendering multiple Nebula instances then handling keyboard navigation between Cells needs to be done external to Nebula. The same applies if we need to move Cells around and control the navigation order.
We do need keyboardNavigation to be true, but we also need to disable Nebulas tabIndecies and keyboard handlers.

Additionally for viewData when setting an external blurHandler, we need to make sure it runs instead of the internal one that gets created as the chart model changes.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/Cell.jsx
+++ b/apis/nucleus/src/components/Cell.jsx
@@ -351,6 +351,7 @@ const Cell = forwardRef(
       translator,
       language,
       keyboardNavigation,
+      externalFocusManagement,
       disableCellPadding = false,
     } = useContext(InstanceContext);
     const [internalEmitter] = useState(emitter || createEmitter);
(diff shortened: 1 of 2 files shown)


apis/nucleus/src/components/Cell.jsx
```diff
@@ -351,6 +351,7 @@ const Cell = forwardRef(
       translator,
       language,
       keyboardNavigation,
+      externalFocusManagement,
```
T-Wizard, line 354: consider making it possible to set a generic / default? blurCallback(?) for the externalFocusManagement case

apis/nucleus/src/components/Cell.jsx
```diff
@@ -383,18 +384,21 @@ const Cell = forwardRef(
 
     useEffect(() => {
       eventmixin(focusHandler.current);
+      focusHandler.current.blurCallback = (resetFocus) => {
+        if (focusHandler.current.onBlurHandler) {
+          focusHandler.current.onBlurHandler();
```
T-Wizard, line 389: need to pass on resetFocus to onBlurHandler (and maybe what cell it is connected to)

apis/nucleus/src/components/Cell.jsx
Caele: yea, this might evolve to that, we'll see

T-Wizard (approved):

apis/nucleus/src/viz.js
```diff
@@ -253,25 +278,7 @@ export default function viz({
...
-      }
-      return false;
-    },
-    toggleFocus(focus) {
-      cellRef.current.toggleFocus(focus);
-    },
-    setOnBlurHandler(cb) {
-      cellRef.current.setOnBlurHandler(cb);
-    },
+    ...newExperimental,
```
Caele, line 281: I did this as well to guard against unnessecary usage before we are ready to set it as a public API.
@linhnguyen-qlik @T-Wizard

linhnguyen-qlik: And all the function inside newExperimental which you spread here will get the warning? Great I think.
*  valid types: viewData, cssScaling, snapshot, exportData, exploration
       *  questionable types: supportRefresh, quickMobile, fullscreen
       *  deprecated?: sharing

Caele: There is no warning, they are basically not present if you don't pass in the context option. Mainly want to do this to stop people from seeing the functions when debugging and then using them before we make them official.

linhnguyen-qlik (approved):

PR #1733, opened 2025-05-05
fix: work with client mixins

Caele:
## Motivation

Fixes a compatability issues with the Sense mixins.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


haxxmaxx (commented):
Does useRpc usually return the same instance? Could this lead to extra renders?

Caele:
> Does useRpc usually return the same instance? Could this lead to extra renders?

I think it always returns a new array, and a new object with the state inside

haxxmaxx (approved):

T-Wizard (approved):

PR #1734, opened 2025-05-05
fix: more cell spacing fixes

T-Wizard:
## Motivation

found that padding/spacing was still different then in sense-client after: https://github.com/qlik-oss/nebula.js/pull/1690
mostly for button object

change to compromise between more like sense-client and make it simpler

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/utils/cell-padding.js
+++ b/apis/nucleus/src/utils/cell-padding.js
@@ -1,23 +1,20 @@
 import { generateFiltersString } from './generateFiltersInfo';
 
-const NO_PADDING_IN_CARDS_WITHOUT_TITLE = ['action-button', 'sn-nav-menu'];
+const NO_BORDER_IN_CARDS = ['action-button', 'sn-nav-menu', 'filterpane', 'sn-shape'];
 const NO_PADDING_IN_CARDS = [
+  ...NO_BORDER_IN_CARDS,
   'pivot-table',
   'table',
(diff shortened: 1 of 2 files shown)


apis/nucleus/src/utils/cell-padding.js
```diff
@@ -1,69 +1,68 @@
...
   'sn-filter-pane',
   'sn-layout-container',
-  'sn-listbox',
-  'sn-shape',
   'sn-tabbed-container',
 ];
 
-const shouldUseCardPadding = ({ isCardTheme, layout, isError }) => {
-  const visualization = layout?.visualization;
+const getPaddTitle = (visualization) => NO_BORDER_IN_CARDS.indexOf(visualization) === -1;
```
T-Wizard, line 17: not sure if this variation should be kept, and if yes maybe it should also apply to footer?

Caele: This is the extra horizontal padding for some odd reason? I'd be happy to see if we can get rid of it.

T-Wizard: in most object we do want the padding to get some space to the border.
but not sure if we also want the padding for object that don't have border (by default)

Caele: don't know really

PR #1736, opened 2025-05-07
feat: add theme to galaxy and getPropertyPanelDefinition to viz

Caele:
## Motivation

Adds theme to the galaxy env which means we can get the same setup for property panel def in and out of client.
Adds getPropertyPanelDefinition to new viz APis.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/viz.js
+++ b/apis/nucleus/src/viz.js
@@ -104,6 +104,12 @@ export default function viz({
         }
         return false;
       },
+      getPropertyPanelDefinition() {
+        if (mountedReference && successfulRender) {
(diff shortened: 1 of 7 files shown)


apis/nucleus/src/viz.js
```diff
@@ -104,6 +104,12 @@ export default function viz({
         }
         return false;
       },
+      getPropertyPanelDefinition() {
+        if (mountedReference && successfulRender) {
+          return originalExtensionDef ? originalExtensionDef.ext : cellRef.current.getExtensionDefinition().definition;
```
T-Wizard, line 109: should it not be `originalExtensionDef.definition`?

Caele: yes, right!

PR #1745, opened 2025-05-13
feat: allow mocking of getListObject

veinfors:
Makes it possible to provide a mocked response for `getListObject` when using enigma-mocker through nebula serve.

Instead of always resolving an empty promise it now looks in the list of provided genericObject mocks and returns the corresponding one if provided.

### DIFF
--- a/apis/enigma-mocker/src/from-generic-objects.js
+++ b/apis/enigma-mocker/src/from-generic-objects.js
@@ -16,7 +16,7 @@ export default function fromGenericObjects(genericObjects, options = {}) {
     destroySessionObject: async () => {},
     getObject,
     getAppLayout,
-    getListObject: async () => {},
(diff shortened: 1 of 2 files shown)


apis/enigma-mocker/src/from-generic-objects.js
```diff
@@ -16,7 +16,7 @@ export default function fromGenericObjects(genericObjects, options = {}) {
     destroySessionObject: async () => {},
     getObject,
     getAppLayout,
-    getListObject: async () => {},
+    getListObject: async (props) => getObject(props.qInfo?.qId),
```
Caele, line 19: I wonder why I added this in the first place, as it is only part of the client mixins. Think it was something with embed, but unsure. This is fine though.

Caele (approved):

PR #1729, opened 2025-04-27
Dmh/vna 46 hs generic fn part03

Donya-qlik:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`
    - [ ] No changes **_OR_** API changes has been formally approved
- [ ] Unit/Component test coverage
- [ ] Correct PR title for the changes (fix, chore, feat)

When build and tests have passed:

- [ ] Add code reviewers, for example @qlik-oss/nebula-core

PR #1746, opened 2025-05-14
chore: add hypercube generic functions - part04

Donya-qlik:
## Motivation

Moving generic hypercube functions from sense-client to nebula.js - PART04
In this PR the functions that are required to move, replace and reinsert fields and alternative fields, are considered.

P.S: Other parts of this task can be found here: 
[Part01](https://github.com/qlik-oss/nebula.js/pull/1716) ✅ 
[Part02](https://github.com/qlik-oss/nebula.js/pull/1715) ✅ 
[Part03](https://github.com/qlik-oss/nebula.js/pull/1729) ✅

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/supernova/src/handler/hypercube-handler.js
+++ b/apis/supernova/src/handler/hypercube-handler.js
@@ -11,6 +11,10 @@ import removeMainDimension from './utils/hypercube-helper/remove-main-dimension'
 import removeAlternativeMeasure from './utils/hypercube-helper/remove-alternative-measure';
 import removeMainMeasure from './utils/hypercube-helper/remove-main-measure';
 import removeAlternativeDimension from './utils/hypercube-helper/remove-alternative-dimension';
+import reinsertMainDimension from './utils/hypercube-helper/reinsert-main-dimension';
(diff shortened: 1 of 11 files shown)


apis/supernova/src/handler/hypercube-handler.js
```diff
@@ -163,38 +202,38 @@ class HyperCubeHandler extends DataPropertyHandler {
...
+    const meas = measure;
     meas.qSortBy = {
       qSortByLoadOrder: 1,
       qSortByNumeric: -1,
     };
     return Promise.resolve(meas);
   }
 
-  addMeasures(measures, alternative = false) {
+  addMeasures(measures, alternative) {
```
a-m-dev, line 224: I'd suggest better to be safe in this case, lets keep the false default value

a-m-dev (commented):
Generally looks good, just some minor comments and question. 
But for final approval, I'd highly recommend that @Caele to also have a look at changes

a-m-dev (approved):

PR #1747, opened 2025-05-15
fix: handle outside navigation

jjxie:
## Motivation

Handle navigation api

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


Caele:
@quanho please take a look, I think this is correct as we previously wouldn't get this.
Only other place we read the outside navigation is here: https://github.com/qlik-oss/nebula.js/blob/main/apis/nucleus/src/object/get-generic-object.js#L20 -> in createNavigationAPi.

@jjxie 
What I'm a bit unsure of is if we could move it to the context instead, and have it added directly instead of this roundabout way through galaxy.anything. Compare it to how theme is passed to the useTheme hook when inside sense client:
start here:https://github.com/qlik-trial/sense-client/blob/main/web/assets/objects/extension/supernova-wrapper/sn-ext.js#L131 and  https://github.com/qlik-trial/sense-client/blob/main/web/assets/objects/extension/supernova-wrapper/sn-ext.js#L436 -  its on the context and directly on galaxy, meaning it gets passed in properly already.
Then theme is set in creator: https://github.com/qlik-oss/nebula.js/blob/main/apis/supernova/src/creator.js#L110 (a bit differently as we need the theme object to update when its "name" changes).

jjxie:
For sn-nav-menu, it gets `navigation` from context.  For sn-action-button it gets `navigation` and `platform` from sense currently. https://github.com/qlik-trial/sn-action-button/blob/main/src/index.js#L29. I did a change to get navigation and platform from `anything`. https://github.com/qlik-trial/sn-action-button/pull/262

I think it makes sense to unify sn-nav-menu and sn-action-button to get `navigation` in a same way. I am not sure why sn-nav-menu want to get it from context. 

Platform and logInfo were added through anything in qmfe-embed. https://github.com/qlik-trial/qmfe-embed/blob/main/src/private/flavours/nebula/nebbie/stardust-nebbie-stores.ts#L41. I am not sure what level should navigation to be passed to nebula. From sn-action-button code, platform and navigation are under sense (the deprecated way).

Caele:
> For sn-nav-menu, it gets `navigation` from context. For sn-action-button it gets `navigation` and `platform` from sense currently. https://github.com/qlik-trial/sn-action-button/blob/main/src/index.js#L29. I did a change to get navigation and platform from `anything`. [qlik-trial/sn-action-button#262](https://github.com/qlik-trial/sn-action-button/pull/262)
> 
> I think it makes sense to unify sn-nav-menu and sn-action-button to get `navigation` in a same way. I am not sure why sn-nav-menu want to get it from context.
> 
> Platform and logInfo were added through anything in qmfe-embed. https://github.com/qlik-trial/qmfe-embed/blob/main/src/private/flavours/nebula/nebbie/stardust-nebbie-stores.ts#L41. I am not sure what level should navigation to be passed to nebula. From sn-action-button code, platform and navigation are under sense (the deprecated way).

We get it from context (from useNavigation) because its now available outside of client as well, when embedding the nebula sheet for example.

apis/supernova/src/creator.js
```diff
@@ -109,6 +109,11 @@ function createWithHooks(generator, opts, galaxy) {
           // changed is set further down only if the name is different
           this.context.theme = r.context.theme;
         }
+
+        if (r.context?.navigation) {
+          this.context.navigation = r.context.navigation;
+        }
+
```
T-Wizard, line 116: Not sure but have some worry that this in some edge cases) will override local (object level) navigation with global navigation

T-Wizard (approved):

apis/stardust/types/index.d.ts
```diff
@@ -261,6 +261,19 @@ declare namespace stardust {
...
+        getUrlForSheet?: (sheetId: string) => string;
+        goToSheet?: (sheetId: string, chartId?: string) => void;
+        nextSheet?: () => boolean;
+        prevSheet?: () => boolean;
+        goToStory?: (storyId: string) => void;
+        getCurrentSheetId?: () => string | boolean;
+        getCurrentStoryId?: () => string | boolean;
+        refreshDynamicViews?: () => void;
+        getOdagLinks?: (app: qix.Doc) => unknown[];
+        openOdagPopup?: (app: qix.Doc, odagLinkId: string, element: HTMLButtonElement) => void;
```
Caele, line 274: Where did all this come from?

Caele: This file is auto generated from the JSdoc comments, so those should be updated first. Also this isn't the Navigation API from client, its most often just a subset.

apis/supernova/src/creator.js
Caele: I don't think this is needed on Render, it could be set when the object is created instead. Navigation is a global API that can't change during rendering.
So I think just moving this up to where for example this is done:
```
  const c = {
    context: {
      // static values that are not expected to
      // change during the component's life
      // --------------------
      model: opts.model,
      app: opts.app,
      global: qGlobal,
      selections: opts.selections,
      nebbie: opts.nebbie,
      element: undefined, // set on mount
      emitter: opts.emitter,
```

apis/stardust/types/index.d.ts
jjxie: I wrote the types myself according to https://github.com/qlik-trial/sense-client/blob/main/web/assets/objects/extension/supernova-wrapper/navigation.js

apis/supernova/src/creator.js
jjxie: I moved navigation to cell.js instead where has access of context and can set navigation

apis/stardust/types/index.d.ts
jjxie: I just put optional object for navigation in context.

apis/nucleus/src/index.js
```diff
@@ -157,6 +157,8 @@ const DEFAULT_CONTEXT = /** @lends Context */ {
    * @type {string=}
    * */
   dataViewType: 'sn-table',
+  /** @type {object=} */
```
Caele, line 160: This should be of type Navigation? (from navigation.js)

jjxie: It should be navigation from sense client [navigation.js](https://github.com/qlik-trial/sense-client/blob/main/web/assets/objects/extension/supernova-wrapper/navigation.js), not sheet navigation from nebula which has less functions. 
I am not sure if we should write all the functions from sense in here, which we will maintain when sense client navigation changes, and if we want to expose what functions we have? I saw for example hostConfig doesn't specify, maybe a different case?

Caele: I think it should always be typed as Navigation from Nebula, as that is what Nebula explicitly supports. Then it can be a different object but the required properties on that object is what Nebula defines.

jjxie: I added an objectnavigation type

apis/nucleus/src/object/navigation/typedefs.js
```diff
@@ -0,0 +1,79 @@
...
+ * @property {getUrlForSheet=} getUrlForSheet
+ * @property {goToSheet=} goToSheet
+ * @property {nextSheet=} nextSheet
+ * @property {prevSheet=} prevSheet
+ * @property {goToStory=} goToStory
+ * @property {getCurrentSheetId=} getCurrentSheetId
+ * @property {getCurrentStoryId=} getCurrentStoryId
+ * @property {refreshDynamicViews=} refreshDynamicViews
+ * @property {getOdagLinks=} getOdagLinks
+ * @property {openOdagPopup=} openOdagPopup
```
Caele, line 78: these are client internals, don't add them here. Just use the already existing Nebula type please. We can look at extending that later, but we can't just add these here.

jjxie: Fixed using existing type. I thought you meant build a different object according to sense client type.

Caele (approved):

PR #1755, opened 2025-05-27
feat: include onContextMenu in experimental api

T-Wizard:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


linhnguyen-qlik (commented):
Do we maybe need to handle the case args is undefined?

T-Wizard:
> Do we maybe need to handle the case args is undefined?

no, that is not possible

linhnguyen-qlik (approved):

PR #1756, opened 2025-05-27
feat: support screen reader in gridless sheet

linhnguyen-qlik:
## Motivation
Support screen reader in gridless sheet. This will probably support screen reader for embed of object too. Not nice to put translation for all objects here but it is faster doing this. 

## Requirements checklist

- [ ] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/Header.jsx
+++ b/apis/nucleus/src/components/Header.jsx
@@ -83,7 +83,13 @@ function Header({ layout, sn, anchorEl, hovering, focusHandler, titleStyles = {}
         <Grid container wrap="nowrap" direction="column">
           {showTitle && (
             <Tooltip title={layout.title}>
-              <Typography variant="h6" noWrap className={CellTitle.className} style={titleStyles.main}>
+              <Typography
+                id={`${layout.qInfo.qId}_title`}
+                variant="h6"
+                noWrap
+                className={CellTitle.className}
+                style={titleStyles.main}
+              >
                 {layout.title}
               </Typography>
             </Tooltip>
--- a/apis/nucleus/src/components/Header.jsx
+++ b/apis/nucleus/src/components/Header.jsx
@@ -4,6 +4,7 @@ import { styled } from '@mui/material/styles';
 
 import { Grid, Tooltip, Typography } from '@mui/material';
 import ActionsToolbar from './ActionsToolbar';
+import hiddenScreenReaderText from '../utils/style/screen-reader';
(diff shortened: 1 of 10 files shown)


apis/nucleus/src/components/Header.jsx
```diff
@@ -83,7 +83,13 @@ function Header({ layout, sn, anchorEl, hovering, focusHandler, titleStyles = {}
         <Grid container wrap="nowrap" direction="column">
           {showTitle && (
             <Tooltip title={layout.title}>
-              <Typography variant="h6" noWrap className={CellTitle.className} style={titleStyles.main}>
+              <Typography
+                id={`${layout.qInfo.qId}_title`}
```
Caele, line 87: This might not be unique within a html page, there is a `currentId` in Cell you can use. (same below for `${layout.qInfo.qId}_content` and the `${layout?.qInfo.qId}_type`)

apis/nucleus/src/components/Supernova.jsx
```diff
@@ -19,7 +19,7 @@ const VizElement = {
   className: 'njs-viz',
 };
 
-function Supernova({ sn, snOptions: options, snPlugins: plugins, layout, appLayout, halo }) {
+function Supernova({ sn, snOptions: options, snPlugins: plugins, layout, appLayout, halo, cellId }) {
```
linhnguyen-qlik, line 22: @Caele Could this function be used without going through cell.jsx?

Caele: yes, but only inside current client. It is only expose through here: https://github.com/qlik-oss/nebula.js/blob/main/apis/stardust/src/index.js#L48, the `__DO_NOT_USE__`.

Caele (approved):

apis/nucleus/src/components/Header.jsx
```diff
@@ -83,7 +83,13 @@ function Header({ layout, sn, anchorEl, hovering, focusHandler, titleStyles = {}
         <Grid container wrap="nowrap" direction="column">
           {showTitle && (
             <Tooltip title={layout.title}>
-              <Typography variant="h6" noWrap className={CellTitle.className} style={titleStyles.main}>
+              <Typography
```
jjxie, line 86: For classic chart, when there is no title will read no title. Probably better add something to make it work same.

jjxie (approved):

PR #1779, opened 2025-06-26
chore: update types and export HyperCubeHandler

Donya-qlik:
## Motivation

In this PR, I updated the type of index-file and exported HyperCubeHandler class for later use in nebula-tables [PR](https://github.com/qlik-trial/nebula-tables/pull/1355), where pivot hypercube handlers will be defined.

## Requirements checklist

- [x] Api specification
  - [x] Ran `yarn spec`

### DIFF
--- a/apis/supernova/src/handler/data-property-handler.js
+++ b/apis/supernova/src/handler/data-property-handler.js
@@ -17,7 +17,7 @@ import { notSupportedError } from './utils/hypercube-helper/hypercube-utils';
  * @entry
  * @export
  * @example
- * import DataPropertyHandler from './data-property-handler';
+ * import DataPropertyHandler from '@nebula.js/stardust';
  *
  * class PivotHyperCubeHandler extends DataPropertyHandler {
  *
--- a/apis/supernova/src/handler/hypercube-handler.js
+++ b/apis/supernova/src/handler/hypercube-handler.js
@@ -29,7 +29,7 @@ import reinsertMainMeasure from './utils/hypercube-helper/reinsert-main-measure'
  * @entry
  * @export
  * @example
- * import { HyperCubeHandler } from './data-property-handler';
+ * import { HyperCubeHandler } from '@nebula.js/stardust';
  *
  * class PivotHyperCubeHandler extends HyperCubeHandler {
  *


apis/supernova/src/handler/utils/field-helper/__tests__/get-sorted-field.test.js
```diff
@@ -1,4 +1,4 @@
-import getAutoSortFieldDimension from '../get-sorted-field';
```
Donya-qlik, line 1: just a name changing

apis/nucleus/handler/data-property-handler.js
```diff
@@ -1,11 +1,18 @@
...
 
+/**
+ * @class DataPropertyHandler
+ * @description A class to handle data properties for dimensions and measures in a data model.
+ * @export
+ */
 class DataPropertyHandler {
+  /**
+   * Creates an instance of DataPropertyHandler.
+   * @param {object} opts - Options for the handler.
```
T-Wizard, line 14: recommend to also document what possible option that can be in the opts parameter

apis/nucleus/handler/data-property-handler.js
```diff
@@ -256,10 +281,6 @@ class DataPropertyHandler {
     measure.qDef = measure.qDef ?? {};
     measure.qDef.qNumFormat = measure.qDef.qNumFormat ?? {};
 
-    if (isEnabled('MASTER_MEASURE_FORMAT')) {
-      useMasterNumberFormat(measure.qDef);
-    }
-
```
T-Wizard, line 262: why is this removed?

apis/nucleus/handler/hypercube-handler.js
```diff
@@ -41,30 +55,70 @@ class HyperCubeHandler extends DataPropertyHandler {
...
   getDimensionLayouts() {
     const hc = hcUtils.getHyperCube(this.layout, this.path);
     return hc ? hc.qDimensionInfo : [];
   }
 
+  /**
+   *
+   * @param {object} dimension
+   * @param {boolean} alternative
+   * @param {number} idx
```
T-Wizard, line 114: mark as optional (?)
```suggestion
   * @param {number} [idx]
```

apis/nucleus/handler/hypercube-handler.js
```diff
@@ -156,62 +241,119 @@ class HyperCubeHandler extends DataPropertyHandler {
...
+   * const measure = hyperCubeHandler.addMeasure(measure, alternative, idx);
+   */
   addMeasure(measure, alternative, idx) {
-    const measures = this.getAlternativeMeasure();
-    const meas = initializeId(measure);
+    const meas = initializeField(measure);
 
-    if (hcUtils.isMeasureAlternative(this, measures, alternative)) {
-      return hcUtils.addAlternativeMeasure(this, meas, idx);
+    if (hcUtils.isMeasureAlternative(this, meas, alternative)) {
```
T-Wizard, line 340: should not the second argument to isMeasureAlternative be the list of current measures?

apis/stardust/api-spec/spec.json
```diff
@@ -2225,7 +2225,7 @@
       },
       "entries": {},
       "examples": [
-        "import { useElement, useLayout } from '@nebula.js/stardust';\n\nexport default function() {\n  return {\n    qae: {\n      properties: {\n        dude: 'Heisenberg',\n      }\n    },\n    component() {\n      const el = useElement();\n      const layout = useLayout();\n      el.innerHTML = `What's my name? ${layout.dude}!!!`;\n    }\n  };\n}"
+        "import { useElement, useLayout } from '@nebula.js/stardust';\r\rexport default function() {\r  return {\r    qae: {\r      properties: {\r        dude: 'Heisenberg',\r      }\r    },\r    component() {\r      const el = useElement();\r      const layout = useLayout();\r      el.innerHTML = `What's my name? ${layout.dude}!!!`;\r    }\r  };\r}"
```
Caele, line 2228: Are you running a windows machine?

apis/supernova/src/handler/data-property-handler.js
```diff
@@ -1,11 +1,27 @@
 import extend from 'extend';
-// eslint-disable-next-line import/no-relative-packages
-import isEnabled from '../../../nucleus/src/flags/flags';
-import { findFieldById, useMasterNumberFormat } from './utils/field-helper/field-utils';
+import { findFieldById, initializeField } from './utils/field-helper/field-utils';
 import { INITIAL_SORT_CRITERIAS } from './utils/constants';
 import { notSupportedError } from './utils/hypercube-helper/hypercube-utils';
 
+/**
+ * @class DataPropertyHandler
```
Caele, line 7: Add:
`@entry` - as this _should be?_ exported by stardust.

apis/stardust/src/index.js
```diff
@@ -1,14 +1,14 @@
...
-import { generator, hook } from '@nebula.js/supernova';
+import { generator, hook, HyperCubeHandler } from '@nebula.js/supernova';
 import theme from '@nebula.js/theme';
 import locale from '@nebula.js/locale';
 import conversion from '@nebula.js/conversion';
 import EnigmaMocker from '@nebula.js/enigma-mocker';
 
 // mashup api
-export { embed, conversion, EnigmaMocker };
+export { embed, conversion, EnigmaMocker, HyperCubeHandler };
```
Caele, line 11: HypercubeHandler is not part of the mashup API right? its for charts/viz so move it down to the component api. 
Also add DataPropertyHandler as well?

apis/supernova/src/handler/data-property-handler.js
```diff
@@ -82,11 +98,17 @@ class DataPropertyHandler {
     throw notSupportedError;
   }
 
+  /**
+   * Creates a type of library dimension with a field definition.
+   * @param {string} id - Dimension id
+   * @param {object | undefined} [defaults] - Default properties for the dimension.
+   * @returns {NxDimension} The created dimension object.
```
Caele, line 105:
```suggestion
   * @returns {qix.NxDimension} The created dimension object.
```

apis/supernova/src/handler/hypercube-handler.js
```diff
@@ -156,62 +241,119 @@ class HyperCubeHandler extends DataPropertyHandler {
...
+      const hcMeasures = this.hcProperties.qLayoutExclude.qHyperCubeDef.qMeasures;
+      return hcUtils.addAlternativeMeasure(meas, hcMeasures, idx);
     }
 
     return addMainMeasure(this, meas, idx);
   }
 
+  /**
+   * @param {NxMeasure} measure
+   * @returns {NxMeasure} measure with auto-sort properties
```
Caele, line 350:
```suggestion
   * @param {qix.NxMeasure} measure
   * @returns {qix.NxMeasure} measure with auto-sort properties
```

apis/supernova/src/handler/hypercube-handler.js
```diff
@@ -222,6 +364,17 @@ class HyperCubeHandler extends DataPropertyHandler {
     return Promise.resolve(meas);
   }
 
+  /**
+   * @param {NxMeasure[]} measures
+   * @param {boolean} alternative
+   * @returns {NxMeasure[]} added measures
```
Caele, line 370:
```suggestion
   * @param {qix.NxMeasure[]} measures
   * @param {boolean} alternative
   * @returns {qix.NxMeasure[]} added measures
```

apis/supernova/src/handler/data-property-handler.js
```diff
@@ -189,7 +198,7 @@ class DataPropertyHandler {
 
   /**
    * Gets the maximum number of dimensions allowed.
-   * @param {number} [decrement=0] - The number to decrement from the maximum dimensions.
+   * @param {number} [decrement=0] - The number to decrement from the current dimensions.
```
T-Wizard, line 201:
```suggestion
   * @param {number} [decrement=0] - The number to decrement from the current number of measures.
```

apis/nucleus/handler/data-property-handler.js
T-Wizard: but why remove it?
why do we not want to default to use master measure format (like in sense-client)?

apis/nucleus/handler/hypercube-handler.js
```diff
@@ -156,62 +241,119 @@ class HyperCubeHandler extends DataPropertyHandler {
...
+   * If the measure is an alternative, it will be added to the alternative measures.
+   * If the total number of measures exceeds the limit, it will stop adding measures.
+   * @memberof HyperCubeHandler
+   * @example
+   * const measure = hyperCubeHandler.addMeasure(measure, alternative, idx);
+   */
   addMeasure(measure, alternative, idx) {
-    const measures = this.getAlternativeMeasure();
-    const meas = initializeId(measure);
+    const meas = initializeField(measure);
```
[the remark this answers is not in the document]
Donya-qlik: Regarding the original code it is checked and added for dimensions as well
https://github.com/qlik-trial/sense-client/blob/d568423b54640ee1f2942ac58dd846af7c3751f0/web/assets/objects/utils/data-properties/data-property-handler.ts#L337

apis/supernova/src/handler/data-property-handler.js
```diff
@@ -113,72 +270,172 @@ class DataPropertyHandler {
...
+   * @example
+   * handler.addFieldDimension('A', 'AA');
+   */
   addFieldDimension(field, label, defaults) {
     const dimension = this.createFieldDimension(field, label, defaults);
     return this.addDimension(dimension);
   }
 
+  /**
+   * @param {FieldDimensions[]} args - Array of field dimension args.
```
Donya-qlik, line 295: @Caele Do you think it makes sense to define a typedef `FieldDimensions` as I did, to set here instead of giving it the type of `object[]`

/**
   * @typeof {object} FieldDimensions
   * @property {string} field
   * @property {string=} label
   * @property {qix.NxDimension=} defaults
   */

apis/supernova/src/handler/data-property-handler.js
```diff
@@ -39,54 +123,117 @@ class DataPropertyHandler {
...
 
     return findFieldById(dimensions, id) ?? findFieldById(alternativeDimensions, id);
   }
 
+  /**
+   * Throws an error indicating the method must be implemented in subclasses.
+   * @throws {Error}
+   * @memberof DataPropertyHandler
+   */
   static getAlternativeDimensions() {
```
Donya-qlik, line 156: @Caele I'm not sure what to do with these functions... moving them to nebula.js thrown a typeError so I added `static` to it but don't know if it is how they are supposed to be!

Caele: hmm, `static` is fine to have on these, but you could also have static on all class functions. Where do you get this typeError?

apis/supernova/src/handler/data-property-handler.js
Caele: Best would be if we could align the whole file, for example I see that:
```
addFieldDimensions: (args:FieldDimensions[]):Promise<Promise<qix.NxDimension[]>
```
compared to:
```
addAlternativeFieldDimension: (field, label, defaults):Promise<qix.NxDimension=>
```
So we are not consistent. The problem with adding a type for one of them is that the other won't have a type.
And that we can't change the function signature, it has to stay the same.
So I think it best to use object here instead, and same for other similar cases. But I'm not sure.

apis/supernova/src/handler/data-property-handler.js
```diff
@@ -113,72 +270,172 @@ class DataPropertyHandler {
...
     }
-
     return this.dimensionDefinition.min || 0;
   }
 
+  /**
+   * Gets the maximum number of dimensions allowed.
+   * @param {number} [decrement=0] - The number to decrement from the current number of dimensions.
+   * @returns {number} The maximum number of dimensions allowed.
+   * @description Checks if the max property is a function and calls it with the current number of dimensions, or returns a default value.
```
T-Wizard, line 415: the argument to the max dimensions function is current number of **measures** (decremented by decrement parameter)

```suggestion
   * Gets the maximum number of dimensions allowed.
   * @param {number} [decrement=0] - The number to decrement from the current number of measures.
   * @returns {number} The maximum number of dimensions allowed.
   * @description Checks if the max property is a function and calls it with the current number of measures, or returns a default value.
```

apis/supernova/src/handler/data-property-handler.js
```diff
@@ -18,77 +46,205 @@ class DataPropertyHandler {
...
+    return findFieldById(dimensions, libraryDimension.id) ?? findFieldById(alternativeDimensions, libraryDimension.id);
   }
 
-  static getAlternativeDimensions() {
+  /**
+   * Throws an error indicating the method must be implemented in subclasses.
+   * @throws {Error}
+   * @memberof DataPropertyHandler
+   */
+  // eslint-disable-next-line class-methods-use-this
```
Donya-qlik, line 158: @Caele I ended up disabling typeError to prevent using `static`
typeError: `Expected 'this' to be used by class method 'addMeasure'`

apis/supernova/src/handler/hypercube-handler.js
```diff
@@ -13,26 +12,51 @@ import removeMainMeasure from './utils/hypercube-helper/remove-main-measure';
...
+ * const handler = new HyperCubeHandler({
+ *   app: qlikApp,
+ *   dimensionDefinition: { max: 2 },
+ *   measureDefinition: { max: 2 },
+ *   properties: {},
+ * });
+ *
+ * handler.setProperties({ qDimensions: [], qMeasures: [] });
+ * const dims = handler.getDimensions();
+ */
```
Caele, line 34: We should add an example here (and in DataPropertyHandler) how its used in practice, so in a chart.

apis/supernova/src/handler/hypercube-handler.js
```diff
@@ -13,26 +12,51 @@ import removeMainMeasure from './utils/hypercube-helper/remove-main-measure';
...
 class HyperCubeHandler extends DataPropertyHandler {
+  /**
+   * Creates an instance of HyperCubeHandler.
+   * @param {object} opts - Options for the handler.
+   */
```
Caele, line 39: Remove this part, it will only add duplicates into the spec documentation, making it harder to read. (Same for DataPropertyHandler)

apis/supernova/src/generator.js
```diff
@@ -76,7 +77,10 @@ export default function generatorFn(UserSN, galaxy) {
     },
     definition: galaxy.flags.isEnabled('NEBULA_DATA_HANDLERS')
       ? {
-          dataHandler: (opts) => new HyperCubeHandler(opts),
+          dataHandler: (opts) => ({
+            HyperCubeHandler: new HyperCubeHandler(opts),
+            DataPropertyHandler: new DataPropertyHandler(opts),
+          }),
```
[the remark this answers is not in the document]
Caele: Shouldn't client import those from Nebula instead?

PR #2083, opened 2026-06-12
fix: fix listbox title for master dimension

quanho:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/object/filterpane-handler.js
+++ b/apis/nucleus/src/object/filterpane-handler.js
@@ -36,9 +36,16 @@ export default function filterpaneHandler({ /* dc, def, properties, */ children,
                 qExpr: dimProps.qDim.qLabelExpression,
               },
             };
-          } else {
-            listboxProps.title = dimProps.qDim.title;
+          } else if (dimProps.qDim.qGrouping === 'C') {
+            listboxProps.title = {
+              qStringExpression: {
+                qExpr: `GroupDimensionLabel('${dimProps.qDim.title}')`,
+              },
+            };
           }
+          // For regular (non-cyclic) master dimensions without a label expression,
+          // leave title empty so qFallbackTitle is used instead. This ensures the
+          // filterpane title always reflects the current master dimension name.
         } else {
           listboxProps.title =
             dimension.qDef.title || dimension.qDef.qFieldLabels?.[0] || dimension.qDef.qFieldDefs?.[0];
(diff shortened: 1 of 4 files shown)


Caele (approved):

Caele (commented):
I don't exactly understand the change. Looks like we never use the qDim.title after it?
And thats what is normally set from client?

apis/nucleus/src/object/filterpane-handler.js
```diff
@@ -36,9 +36,16 @@ export default function filterpaneHandler({ /* dc, def, properties, */ children,
...
+          } else if (dimProps.qDim.qGrouping === 'C') {
+            listboxProps.title = {
+              qStringExpression: {
+                qExpr: `GroupDimensionLabel('${dimProps.qDim.title}')`,
+              },
+            };
           }
+          // For regular (non-cyclic) master dimensions without a label expression,
+          // leave title empty so qFallbackTitle is used instead. This ensures the
+          // filterpane title always reflects the current master dimension name.
```
T-Wizard, line 48: why have a special case for cyclic dimension, but not drilldown dimension?

T-Wizard:
@Caele not sure this was ready to be merged yet.
could be to breaking for existing filter pane charts
and it should probably be synced with the other related PRs in sense-client and "filter pane" repos

Caele:
> @Caele not sure this was ready to be merged yet. could be to breaking for existing filter pane charts and it should probably be synced with the other related PRs in sense-client and "filter pane" repos

ok, I found that it fixed a problem I noticed so check a bit and it worked. But i'll revert this then

apis/nucleus/src/object/filterpane-handler.js
quanho: Good question — it looks like it should handle drilldown too, but it turns out they behave differently in the engine.

For cyclic dimensions, GroupDimensionLabel('group name') is the correct expression to use — the engine resolves it to the currently active field's label as the user cycles through.

For drilldown dimensions, GroupDimensionLabel doesn't apply. The engine already exposes the current active field via qFallbackTitle in the layout, so leaving the stored title empty and reading qFallbackTitle at render time (which is what ListBoxInline does) is the correct approach.

This is consistent with how sense-client handles it — cyclic gets the expression, drilldown (and regular master dims) rely on qFallbackTitle.

quanho:
> @Caele not sure this was ready to be merged yet. could be to breaking for existing filter pane charts and it should probably be synced with the other related PRs in sense-client and "filter pane" repos

I checked with existing filter-pane and it works as it should.
The change only affects existing filter-panes with a master-dimension but this is what we want to change.
@T-Wizard : Do you mean we should put this under a feature flag?

PR #2086, opened 2026-06-16
feat: images in listbox

PurwaShrivastava:
Added a new way to represent data in Listbox as Image. Two options are available for that -
1.) By Dimension value (which is a URL)
2.) URL expression (the user provides) 

User can also select Image Size option from a dropdown -
Skärmavbild 2026-07-28 kl  13 24 06

Attached is video showing the two options -
1.) Dimension Value is URL
https://github.com/user-attachments/assets/380826e2-d24f-41e3-b510-bc050ee57ded
2.) URL Expression
https://github.com/user-attachments/assets/c9982350-873f-4b33-94a3-c4418d3dd143

- [ ] Api specification
  - [ ] Ran `yarn spec`

Initial PR to implement the new feature - Images in Filterpane


apis/nucleus/src/components/listbox/components/ListBoxRowColumn/helpers/get-row-from-pages.js
```diff
@@ -1,11 +1,12 @@
-export default function getCellFromPages({ pages, cellIndex }) {
-  let c;
-  const page = pages.filter((p) => p.qArea.qTop <= cellIndex && cellIndex < p.qArea.qTop + p.qArea.qHeight)[0];
+// Returns the full qMatrix row for a given list index. With list-object `qExpressions`, a row
+// holds more than the dimension cell: [dimensionCell, exprCell0, exprCell1, ...].
+export default function getRowFromPages({ pages, cellIndex }) {
```
veinfors, line 3: With this change, I guess it also makes sense to rename the file to `get-row-from-pages.js`

apis/nucleus/src/components/listbox/components/ListBoxRowColumn/components/Image.jsx
```diff
@@ -0,0 +1,145 @@
...
+  }
+};
+
+function Image({ representation, src, label }) {
+  const { imageSize, imagePosition } = representation;
+  const isFitHeight = imageSize === 'fitHeight';
+  const resolvedImagePosition = resolveImagePosition(imagePosition);
+  const maxImageHeight = '200px';
+
+  const imgNode = src ? (
```
veinfors: I think we should validate the image src for security before putting it into the DOM.
Something like this:

```js
const isSafeImageSrc = (src) => {
  if (typeof src !== 'string' || !src.trim()) {
    return false;
  }

  try {
    const url = new URL(src, window.location.origin);
    return url.protocol === 'https:' || url.protocol === 'http:';
  } catch {
    return false;
  }
};

function Image({ representation, src, label }) {
  const { imageSize, imagePosition } = representation;
  const isFitHeight = imageSize === 'fitHeight';
  const resolvedImagePosition = resolveImagePosition(imagePosition);
  const maxImageHeight = '200px';
  const safeSrc = isSafeImageSrc(src) ? src : null;

  const imgNode = safeSrc ? (
    <img
      src={safeSrc}
```

apis/nucleus/src/components/listbox/components/grid-list-components/grid-list-components.jsx
```diff
@@ -46,6 +46,17 @@ export default function getListBoxComponents({
   const { layoutOptions = {} } = layout || {};
   const { columnWidth, listHeight, itemHeight, rowCount, columnCount } = sizes || {};
 
+  // Representation settings and attribute-expression index map from dimension info, used for both List and Grid.
+  const dimensionInfo = layout?.qListObject?.qDimensionInfo;
+  const representation = dimensionInfo?.representation;
```
veinfors: Hmm.. it feels a little bit weird/wrong to have the representation property in here. I would expect it to be in the properties root level of the object. That is where we have other "similar" settings. 

The expressions though - imageUrl and imageLabel should be kept here, or possibly straight under dimensionInfo in my opinion.

veinfors (approved):

PR #2123, opened 2026-07-01
fix: make listbox cover entire parent element

Caele:
## Motivation

When mounting a listbox in an element it has 100% height but not width, so will not fit into for example a popover properly.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


Caele:
@quanho Do you think this is a safe change in regards to the filterpane?

quanho (approved):

PR #2134, opened 2026-07-06
feat(listbox): use qFallbackTitle when title is undefined

T-Wizard:
useful for using title from master dimension as it will be in qFallbackTitle

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`


apis/nucleus/src/components/listbox/ListBoxInline.jsx
```diff
@@ -183,7 +184,11 @@ function ListBoxInline({ options, layout }) {
   const isLocked = layout?.qListObject?.qDimensionInfo?.qLocked;
   const showSearchIcon = searchEnabled !== false && search === 'toggle' && !isLocked;
 
-  const canShowTitle = layout?.title?.length && layout?.showTitle !== false;
+  const dimInfo = layout?.qListObject?.qDimensionInfo;
+  const effectiveTitle = flags.isEnabled('QCB-35224_FILTERPANE_TITLE_IMPROVEMENT')
+    ? (layout?.title ?? dimInfo?.qFallbackTitle)
```
quanho, line 212: If layout?.title === '' then should we still fallback to dimInfo?.qFallbackTitle?

T-Wizard: to keep old filterpane/listboxes unchanged it only fallback to dimInfo?.qFallbackTitle
when layout?.title === undefined and not when layout?.title === ''

quanho (approved):

PR #2104, opened 2026-06-22
fix: fix listbox title for master dimension

Caele:
## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`
    - [ ] No changes **_OR_** API changes has been formally approved
- [ ] Unit/Component test coverage
- [ ] Correct PR title for the changes (fix, chore, feat)

When build and tests have passed:

- [ ] Add code reviewers, for example @qlik-oss/nebula-core

PR #2188, opened 2026-07-27
fix: fix listbox title for master dimension (re-land + drilldown support)

quanho:
Re-lands #2083 which was reverted in #2103.

**Changes**

- Re-applies the original fix for listbox title on master dimensions
- Adds unit tests for cyclic and drilldown grouped master dimension title handling

Why GroupDimensionLabel is only for cyclic, not drilldown
Cyclic dimensions (qGrouping === 'C') need the GroupDimensionLabel('group name') engine expression because the active field changes when the user cycles — the expression resolves dynamically to the current field's label.

Drilldown dimensions (qGrouping === 'H') don't need this. The engine already reflects the current active field in qFallbackTitle on the layout, which ListBoxInline reads at render time. This is also consistent with how sense-client handles it.

Backward compatibility
This does not break existing filterpanes. The three changed files each handle a distinct case:

- filterpane-handler.js: For regular (non-cyclic) master dims, the stored title is now left empty instead of capturing a static name at creation time. ListBoxInline compensates by reading the engine's live qFallbackTitle when qLibraryId is present — so the displayed title is always current, even if the dimension is renamed in the library.
- ListBoxInline.jsx: The effectiveTitle logic uses qFallbackTitle for master dims and layout.title for inline dims — both paths behave correctly for existing and new objects.
- ListBoxHeader.jsx: The new title prop is optional with a ?? layout.title fallback, so any caller that does not pass it is unaffected.

Before the change:

https://github.com/user-attachments/assets/8dad4d64-06e4-4f15-849b-56f0a72180d3

After the change:

https://github.com/user-attachments/assets/f21896f1-a3b4-4411-b8a7-2675b373873f

### DIFF
--- a/apis/nucleus/src/components/listbox/ListBoxInline.jsx
+++ b/apis/nucleus/src/components/listbox/ListBoxInline.jsx
@@ -184,7 +184,7 @@ function ListBoxInline({ options, layout }) {
   const showSearchIcon = searchEnabled !== false && search === 'toggle' && !isLocked;
 
   const dimInfo = layout?.qListObject?.qDimensionInfo;
-  const effectiveTitle = (dimInfo?.qLibraryId ? dimInfo?.qFallbackTitle : layout?.title);
+  const effectiveTitle = dimInfo?.qLibraryId ? dimInfo?.qFallbackTitle : layout?.title;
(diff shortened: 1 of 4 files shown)


quanho:
Duplicate of https://github.com/qlik-oss/nebula.js/pull/2104

PR #2226, opened 2026-08-07
feat: make requirements more responsive + toolbarmode

Caele:
## Motivation

The requirements show up strangely when the chart is too small, this fixes that by removing the "1 of 2 dimension" texts when its too small to fit,

Additionally adds a toolbarMode to the listbox API to allow for better control over how it is displayed, needed when the inline listbox is used inside a different popover.

## Requirements checklist

- [ ] Api specification
  - [ ] Ran `yarn spec`

### DIFF
--- a/apis/nucleus/src/components/Error.jsx
+++ b/apis/nucleus/src/components/Error.jsx
@@ -92,8 +92,7 @@ export default function Error({ title = 'Error', message = '', data = [] }) {
     baseContentRect.width + descriptionsMeasureRect.width <= containerRect.width;
 
   const hasMeasurements = !!containerRect && !!baseContentRect && !!descriptionsMeasureRect;
-  const showDescriptions =
-    hasDescriptions && hasMeasurements && (!needsSecondColumn || secondColumnFits);
+  const showDescriptions = hasDescriptions && hasMeasurements && (!needsSecondColumn || secondColumnFits);
 
   return (
     <Grid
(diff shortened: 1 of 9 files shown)


quanho:
This is how it works when I tested:

https://github.com/user-attachments/assets/ec40889d-76d2-4704-8ddc-19a788f5fc80

LGTM

quanho:
`toolbarMode` and `ToolbarMode` type were missing from apis/stardust/types/index.d.ts
Screenshot 2026-08-10 at 19 56 22

quanho:
**Test coverage gaps for the 6 `toolbarMode` scenarios:**

| # | Scenario | Covered? |
|---|---|---|
| 1 | `toolbar=false` | N/A — `ListBoxHeader` is never called, handled at `ListBoxInline` level |
| 2 | `'attach'`, short title | ⚠️ Default test uses `toolbarMode: 'auto'`, not `'attach'` explicitly |
| 3 | `'detach'`, short title | ✅ `'should render a detached toolbar when told to do so'` |
| 4 | `'attach'`, long title → stays inline | ❌ Missing — no test verifies that auto-detection is skipped and toolbar stays inline |
| 5 | `'auto'`, short title → attach | ⚠️ Covered by default test but no explicit assertion that `isDetached === false` |
| 6 | `'auto'`, long title → detach | ❌ `test.skip('should render a detached toolbar when space is limited')` — skipped |

Scenario 4 is the most important gap: it's the core guarantee that `toolbarMode='attach'` forces the toolbar inline even when the title is truncated. Without a test for it, a future change could silently break that behaviour.

PR #2234, opened 2026-08-28
fix(listbox): adjust row count to avoid scrollbar from covering last row

pernyl:
## Summary
- Fixes QCB-35617: in a filterpane, a grid-view listbox with more columns than fit the width covered the last row of items with its horizontal scrollbar instead of leaving room for it.
- Root cause: `column-mode.js` computed `rowCount` to exactly fill the available height with no space reserved for a horizontal scrollbar track, so when one rendered, it clipped the bottom row.
- Fix: layout is computed once with the full height; if that layout would actually overflow the container width (i.e. a scrollbar will render), it's recomputed with the height reduced by the scrollbar's size, freeing a row for the scrollbar. Listboxes that don't overflow take the same path as before and are unaffected.

## Changes
- `apis/nucleus/src/components/listbox/assets/list-sizes/column-mode.js` — conditional scrollbar-height reservation when horizontal overflow is detected.
- `apis/nucleus/src/components/listbox/assets/list-sizes/__tests__/column-mode.test.jsx` — new tests covering both the overflow case (row reserved) and non-overflow case (unchanged).
- `apis/nucleus/src/components/listbox/assets/list-sizes/__tests__/use-list-sizes.test.jsx` — adjusted one existing test's input height, since its setup exactly matched the bug pattern (zero scrollbar headroom); test intent (max-column clamp) is preserved.

## Test plan
- [x] Existing unit tests pass, no regressions
- [x] New unit tests added for both overflow and non-overflow cases
- [x] Manual visual check in a real tenant filterpane with a wide grid-view listbox (recommended before merge)

Before fix:
image

After fix:
image

Manually tested on Firefox, chrome and edge.


pernyl:
#devbuild

veinfors (approved):
Looks good to me!

